PR #6, opened 2017-10-18
Adding sections of pom requested by central repository

alobodzki:


pom.xml
```diff
@@ -22,6 +27,20 @@
...
 
+  <licenses>
+    <license>
+      <name>Apache License, Version 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+    </license>
+  </licenses>
+
+  <scm>
+     <connection>scm:git:git://github.com/netty/netty.git</connection>
```
mikkokar: Is this reference to a netty git repository intentional?

alobodzki: Totally not.

pom.xml
```diff
@@ -22,6 +27,20 @@
...
+  <licenses>
+    <license>
+      <name>Apache License, Version 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+    </license>
+  </licenses>
+
+  <scm>
+     <connection>scm:git:git@github.com:HotelsDotCom/styx.git</connection>
+     <developerConnection>scm:git:ssh://git@github.com:HotelsDotCom/styx.git</developerConnection>
```
massdosage: Why does this have "ssh" in it but the normal connection doesn't? I think this could just be the same "scm:git:git@github.com:HotelsDotCom/styx.git". This is what we have in WaggleDance which pushes artifacts to Sonatype.

PR #7, opened 2017-10-18
Adding guides to get started with Styx quickly

kvosper:
Caveats: waiting for releases to be public.


README.md
```diff
@@ -43,6 +43,11 @@ info here](https://conferences.oreilly.com/software-architecture/sa-eu/public/sc
 
 ## Useful Information
 
+### Quick Start
+
+* [Quick-start guide](./docs/quick-start.md) explains how to get Styx up and running quickly
```
mikkokar: Quick start guide has significant overlap with https://github.com/HotelsDotCom/styx/blob/master/docs/user-guide/basic-usage.md. Please consider merging the quick start guide into basic-usage.

docs/plugins-dev-intro.md
```diff
@@ -0,0 +1,40 @@
+# STYX
+## Introduction to developing plugins
+
+This is a quick guide to developing plugins for the Styx platform. It is written with the assumption that you 
+are already familiar with running Styx. If not, please consult the [Quick-start guide](quick-start.md).
+
+We have not yet released our modules online as maven dependencies, but this is coming soon!
+For now, please look in the `lib/` folder inside the ([Styx release](https://github.com/HotelsDotCom/styx/releases)).
```
mikkokar: An alternative, perhaps preferable way is to build the project locally, and using `mvn install` lifecycle phase to install the jars locally. Then the plugin project can just add a dependency in the pom file as they would normally do. 
This is BTW what `make release-no-tests` does.

README.md
kvosper: I think we should look into the overlap, but as a separate increment.

docs/plugins-dev-intro.md
kvosper: Agreed.

mikkokar:
Looks good. Thanks @kvosper.

PR #13, opened 2017-10-19
fix documentation

ggaeta1:


docs/user-guide/configure-tls.md
```diff
@@ -67,11 +67,11 @@ backend services configuration, as follows:
     sslProvider:         OPENSSL     # Also supports JDK
     addlCerts:
       - alias:           "my certificate"
-        path:            /path/to/mycert
+        path:            "/path/to/mycert"
       - alias:           "alt certificatfe"
```
massdosage: Shouldn't this say "alt certificate"? I know the typo was here before you but seeing as you're changing this anyway...

PR #15, opened 2017-10-23
Don't try to parse the inbound request paths

taer:
The test updatae in HttpRequestOperationTest is the validation of our bug.
/foo%2Cbar is the same as /foo,bar

Without this patch, styx converts the URL to the comma variant when it passes to the backend.
But the backend was expecting the %2C version and issues a redirect back to the %2C. This causes a redirect loop.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/Url.java
+++ b/components/api/src/main/java/com/hotels/styx/api/Url.java
@@ -157,15 +157,7 @@ public final class Url implements Comparable<Url> {
      */
     public URI toURI() {
         try {
-            String auth = authority.map(Authority::toString).orElse(null);
-
-            String query = query()
-                    .map(q -> q.parameters().stream()
-                            .map(UrlQuery.Parameter::toString)
-                            .collect(joining("&")))
-                    .orElse(null);
-
-            return new URI(scheme, auth, path, query, fragment);
+            return new URI(toString());
         } catch (URISyntaxException e) {
             throw propagate(e);
         }
@@ -236,36 +228,7 @@ public final class Url implements Comparable<Url> {
      * @return the encoded url string
      */
     public String encodedUri() {
-        StringBuilder url = new StringBuilder();
-        if (authority.isPresent()) {
-            if (!isNullOrEmpty(scheme)) {
-                url.append(scheme).append("://");
-            }
-            url.append(authority.get());
-        }
-        url.append(encodePathAndQuery());
-        if (!isNullOrEmpty(fragment)) {
-            url.append("#").append(fragment);
-        }
-
-        return url.toString();
-    }
-
-    private String encodePathAndQuery() {
-        return encodePath(path()) + query.map(query -> "?" + query.encodedQuery()).orElse("");
-    }
-
-    private static String encodePath(String path) {
-        Iterable<String> pathElements = Splitter.on(PATH_DELIMITER).omitEmptyStrings().split(path);
-        StringBuilder encodedPath = new StringBuilder();
-        for (String pathElement : pathElements) {
-            encodedPath.append(PATH_DELIMITER);
-            encodedPath.append(encodePathElement(pathElement));
-        }
-        if (path.endsWith(PATH_DELIMITER)) {
-            encodedPath.append(PATH_DELIMITER);
-        }
-        return encodedPath.toString();
+        return toString();
     }
 
     private static CharSequence encodePathElement(String pathElement) {
@@ -490,7 +453,7 @@ public final class Url implements Comparable<Url> {
             return new Builder()
                     .scheme(uri.getScheme())
                     .authority(uri.getAuthority())
-                    .path(uri.getPath())
+                    .path(uri.getRawPath())
                     .rawQuery(uri.getRawQuery())
                     .fragment(uri.getFragment());
         }
--- a/system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/UnwiseCharactersSpec.scala
+++ b/system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/UnwiseCharactersSpec.scala
@@ -66,7 +66,7 @@ class UnwiseCharactersSpec extends FunSpec with StyxProxySpec {
 
       decodedRequest(req)
 
-      recordingBackend.verify(receivedRewrittenUrl("/url/unwiseQQblah"))
+      recordingBackend.verify(receivedRewrittenUrl("/url/unwise%51%51blah"))
       assertThat(logger.log(), hasItem(loggingEvent(WARN, "Value contains unwise chars. you should fix this. raw=/url/unwiseQQblah, escaped=/url/unwise%51%51blah.*")))
     }
   }
--- a/components/api/src/test/java/com/hotels/styx/api/UrlTest.java
+++ b/components/api/src/test/java/com/hotels/styx/api/UrlTest.java
@@ -149,7 +149,7 @@ public class UrlTest {
     public void handlesPathSegmentEncodedInUTF8() {
         Url url = url("http://example.com/foo/\u2603")
                 .build();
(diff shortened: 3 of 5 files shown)


system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/UnwiseCharactersSpec.scala
```diff
@@ -66,7 +66,7 @@ class UnwiseCharactersSpec extends FunSpec with StyxProxySpec {
 
       decodedRequest(req)
 
-      recordingBackend.verify(receivedRewrittenUrl("/url/unwiseQQblah"))
+      recordingBackend.verify(receivedRewrittenUrl("/url/unwise%51%51blah"))
```
taer: The logic here is
* Unwise says Q is not legal, changes Q -> %51
* This is as if the client sent a %51
* Thus, styx should NOT change %51 to a Q, even though it's legal to do so

components/api/src/main/java/com/hotels/styx/api/Url.java
```diff
@@ -490,7 +490,7 @@ public static Builder url(String value) {
             return new Builder()
                     .scheme(uri.getScheme())
                     .authority(uri.getAuthority())
-                    .path(uri.getPath())
```
taer: This line is problematic. This decodes the path into a proper String, ie, `%2C -> ,`. But then, when you call toString on this URL, you get the decoded version instead of the raw one.

But, if this remains "raw", then the path remains already encoded. So when you call url.encodedUri()  on a path with encoded characters in it, you end up w/ encoded encoding characters.. ie, %D7 -> %25D7   This is because `%` encodes to `%25` 

Going to do what the JDK's URI does. Store both.

taer:
Updated PR. Had to store rawPath on Styx's URL much like the java URI does. This allows us to keep the original path as well as the decoded. The decoded is necessary to use in the encodeURI method. This way, %2C gets stored as %2C for the downstream call, and stored as `,` in URI. When you call encodeUrl, you use the decoded one to re-encode with.

A possible udate would be to use the rawPath in the encodeURI method, since it's presumably already properly encoded.

taer:
Rebased on master and repushed to get Travis to kick off again. Just reran tests locally and they passed.

```[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Styx ............................................... SUCCESS [ 14.634 s]
[INFO] Styx - Bill of Materials ........................... SUCCESS [  0.052 s]
[INFO] Styx - Support ..................................... SUCCESS [  0.078 s]
[INFO] Styx - Support - Non-Styx Test Support ............. SUCCESS [  4.665 s]
[INFO] Styx - Components .................................. SUCCESS [  0.242 s]
[INFO] Styx - Components - Core Api ....................... SUCCESS [  8.477 s]
[INFO] Styx - Components - Styx API Test Support .......... SUCCESS [  1.367 s]
[INFO] Styx - Components - Common ......................... SUCCESS [  5.045 s]
[INFO] Styx - Components - Client ......................... SUCCESS [ 33.332 s]
[INFO] Styx - Components - Server ......................... SUCCESS [  5.107 s]
[INFO] Styx - Components - Proxy .......................... SUCCESS [ 27.370 s]
[INFO] Styx - Support - Test API .......................... SUCCESS [  9.607 s]
[INFO] Styx - System Tests ................................ SUCCESS [  0.156 s]
[INFO] Styx - System Tests - Test Support ................. SUCCESS [  2.058 s]
[INFO] Styx - Origins Starter Application ................. SUCCESS [  0.263 s]
[INFO] Styx - System Tests - Test Plugin .................. SUCCESS [  0.285 s]
[INFO] Styx - System Tests - End To End Tests ............. SUCCESS [02:08 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 04:01 min
[INFO] Finished at: 2017-10-30T08:11:24-05:00
[INFO] Final Memory: 94M/1007M
[INFO] ------------------------------------------------------------------------
```

mikkokar:
Hi Rob,

Thinking of it, the URL handling in Styx should be based on the following principles:

1) Incoming requests may or may not be already percent encoded. Styx doesn't know.

2) Styx core (i.e. not considering the plugins) doesn't care if the URL contains percent-encoded characters or not. Styx core should avoid any unnecessary changes to the percent encoding. Whatever comes in, goes out.

3) Active plugins can change the percent-encoding if they so wish. Styx cannot or will not prevent them from doing so. However, the active plugins doesn't know if the request URL is precent encoded or not. So when the plugins need to change the URL, they should decode the URL themselves, or transform the percent-encoded URL as it is.

4) There are additional concerns for the path-prefix routing which we can do as a future increment (i.e. don't have to address as a part of this PR). The path-prefix routing needs to determine the equivalence of possibly percent-encoded path against the path prefix that is present in the origins configuration file. Also, as a future enhancement we can allow the path prefixes in the origins configuration file to be percent-encoded themselves. The URL RFC has a chapter about determining an equivalence between two URLs. We should follow similar principles in the styx router.

components/api/src/main/java/com/hotels/styx/api/Url.java
```diff
@@ -529,7 +536,8 @@ public Builder authority(String authority) {
          * @return this builder
          */
         public Builder path(String path) {
-            this.path = path;
+            this.rawPath = path;
+            this.path = QueryStringDecoder.decodeComponent(path);
```
mikkokar: We don't need to store the rawPath (or path). Only one of them is necessary. We should store the path component as-is regardless of encoding. As long as we take the raw path from *url()* method like so:

        public static Builder url(String value) {
            URI uri = URI.create(value);
            return new Builder()
                    .scheme(uri.getScheme())
                    .authority(uri.getAuthority())
                    .path(uri.getRawPath())               <--- here
                    .rawQuery(uri.getRawQuery())
                    .fragment(uri.getFragment());
        }

mikkokar: Please change this accordingly. I have write some justification in https://github.com/HotelsDotCom/styx/pull/15#issuecomment-342511138.

taer:
Updated PR with removal of encodedURI() logic. The method just delegates to toString() now.

mikkokar:
Hi Robert,

With the PR as it is, we would be unable to get a decoded representation of the URL components. I can demonstrate this with the following two (new) test cases. BTW please add them to the UrlTest.java class. Here:

    @Test
    public void exposesDecodedPath() throws Exception {
        Url url = Url.Builder.url("http://localhost/foo%20bar").build();
        assertThat(url.toURI().getPath(), is("/foo bar"));
    }

Result:

    java.lang.AssertionError: 
    Expected: is "/foo bar"
         but: was "/foo%20bar"
    Expected :/foo bar
    Actual   :/foo%20bar

So it failed because the path component in the URI itself was not decoded.


    @Test
    public void exposesRawPath() throws Exception {
        Url url = Url.Builder.url("http://localhost/foo%20bar").build();
        assertThat(url.toURI().getRawPath(), is("/foo%20bar"));
    }

Results:

    java.lang.AssertionError: 
    Expected: is "/foo%20bar"
         but: was "/foo%2520bar"
    Expected :/foo%20bar
    Actual   :/foo%2520bar

It failed because the path component is now encoded twice.

So although your changes are otherwise OK, we need to change the Url.toURI() code a bit ensure we can build a correct URI representation from the styx URL class. Like so:

    public URI toURI() {
        try {
            return new URI(toString());
        } catch (URISyntaxException e) {
            throw propagate(e);
        }
    } 

With these additional changes, I believe we should be good to go.

mikkokar:
Hi Robert. The changes look good. Please upgrade your forked repository, and rebase this branch onto latest master, and I will take it from there.

taer:
@mikkokar Done.

PR #16, opened 2017-10-23
Add getOptional to Http context

taer:


components/api/src/main/java/com/hotels/styx/api/HttpInterceptor.java
```diff
@@ -57,6 +59,11 @@ public void add(String key, Object value) {
                 return null;
             }
         };
+
+        default <T> Optional<T> getOptional(String key, Class<T> clazz){
```
kvosper: I think we should use a name like `getIfAvailable` rather than `getOptional`, as it is better to describe the behaviour of the method rather than simply the return type, which is already visible.

kvosper:
@mikkokar I agree with the deprecation. I have also added a comment to the file regarding the naming of the proposed new method.

components/api/src/main/java/com/hotels/styx/api/HttpInterceptor.java
taer: That's fine for me. I'll update w/ the change(and deprecation)

PR #19, opened 2017-10-25
Additional travis related config/Fixes for javadocs/removing styx par…

alobodzki:
…ent pom as suggested

Change to travis config will result in snapshot release on each commit.

### DIFF
--- a/system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/https/EmptyTlsProtocolsListSpec.scala
+++ b/system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/https/EmptyTlsProtocolsListSpec.scala
@@ -1,18 +1,18 @@
 /**
-  * Copyright (C) 2013-2017 Expedia Inc.
-  *
-  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


components/api/src/main/java/com/hotels/styx/api/Clocks.java
```diff
@@ -26,14 +26,15 @@ private Clocks() {
     }
```
kvosper: I don't think we should simply remove the comment text on these methods to the "@return" part. They should have different roles. The main comment text should help make clear when someone would use the method, the "@return" should be a short description of the return value.

components/api/src/main/java/com/hotels/styx/api/FormData.java
```diff
@@ -70,10 +70,10 @@
     /**
      * Helper method to access data of a single parameter.
      *
-     * @param key
+     * @param key parameter key
      * @return The parameter value wrapped in an {@link Optional} if the parameter exists, {@link Optional#empty} otherwise.
      */
-    public Optional<String> postParam(String key) {
+    Optional<String> postParam(String key) {
```
kvosper: `FormData` is part of the API. We cannot reduce visibility just because the internal code is not using it from outside the package.

PR #21, opened 2017-10-26
Log file changes when file backed registry fails at notifying listene…

kvosper:
Log file changes when file backed registry fails at notifying listeners of reload

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
+++ b/components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
@@ -15,20 +15,17 @@
  */
 package com.hotels.styx.infrastructure;
 
-import com.google.common.hash.HashCode;
-import com.google.common.hash.Hashing;
-import com.google.common.io.Files;
 import com.hotels.styx.api.Identifiable;
 import com.hotels.styx.api.Resource;
 import org.slf4j.Logger;
 
-import java.io.File;
 import java.io.IOException;
+import java.util.function.Function;
 
-import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Throwables.propagate;
 import static com.google.common.io.ByteStreams.toByteArray;
 import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
 import static org.slf4j.LoggerFactory.getLogger;
 
 /**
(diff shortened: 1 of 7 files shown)


components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
```diff
@@ -92,15 +96,29 @@ public synchronized void reload(ReloadListener listener) {
     }
 
     private boolean updateResources() {
+        previousFileContent = fileContent;
+        fileContent = readFileAsString();
+
         Iterable<T> resources = reader.read(readFile());
```
mikkokar: We are reading the file twice in succession (in lines 100 and 102). Please change this.

components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
```diff
@@ -118,6 +136,10 @@ private boolean hasFileContentChanged() {
         }
     }
 
+    private String readFileAsString() {
+        return new String(readFile());
+    }
+
     /**
      * Reader.
      *
```
mikkokar: The *Reader* interface should really be called *Parser*. It doesn't really read anything, just parses (or deserialises) an array of bytes into objects of *T*.

components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
```diff
@@ -59,6 +62,7 @@ private HashCode md5(Resource resource) {
     protected void startUp() {
         LOG.info("starting {}", getClass().getSimpleName());
 
+        fileContent = readFileAsString();
```
mikkokar: You can make your unit tests much simpler by extracting the file reading into a separate class and injecting it in.

components/proxy/src/test/java/com/hotels/styx/infrastructure/FileBackedRegistryTest.java
```diff
@@ -41,14 +75,54 @@ public void calculatesTheDifferenceBetweenCurrentAndNewResources() {
         assertThat(changes.toString(), is(expected.toString()));
     }
 
-    private BackendService backendService(String id, int port) {
+    @Test
+    public void logsFileChangeWhenChangeListenerThrowsAnExceptionDuringReload() throws IOException, TimeoutException {
+        log = new LoggingTestSupport(FileBackedRegistry.class);
+        tempFile = createTempFile("origins-", ".txt");
+
+        write(tempFile, "original_test_content".getBytes());
```
mikkokar: Here you have to construct temporary files for testing purposes because code to read origins file (*readFile()*, *readFileAsString()*) is embedded inside *FileBackedRegistry*.

Please extract file reading out into its own class. Then use a mock in its place to get rid of temporary files.

I know this PR is just for adding instrumentation, but lets use the opportunity to add a small improvement. There is a chance it will stick around for awhile.

PR #26, opened 2017-11-06
Map OutOfDirectMemoryError to 500 Internal Server Error status code.

mikkokar:
Fixes an issue where an OutOfDirectMemoryError in Styx Client is mistakenly treated as a bad HTTP response from origin. It used to result a 502 Bad Gateway, but now results a more appropriate 500 Internal Server Error.

Q: You might ask what is the point of attempting to respond when Styx has already ran out of direct memory? 

A: We have discovered after upgrading to Netty 4.1 we started observing occasional OutOfDirectMemoryErrors, which don't appear to be fatal. Styx appears to straight away recover to serve the next request without problems. (At the moment we suspect this is a direct memory sizing issue.)

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/StyxClientError.java
+++ b/components/client/src/main/java/com/hotels/styx/client/StyxClientError.java
@@ -1,0 +1,31 @@
+/**
+ * Copyright (C) 2013-2017 Expedia Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.styx.client;
+
+/**
+ * Signifies an unexpected error from StyxClient.
+ *
+ * A root cause for the error is stored in the `cause` field.
+ * This could be thrown for example if StyxClient runs out of
+ * direct memory.
+ *
+ */
+public class StyxClientError extends RuntimeException {
+    public StyxClientError(String message, Throwable cause) {
+        super(message, cause);
+    }
+}
+
--- a/components/client/src/main/java/com/hotels/styx/client/netty/NettyToStyxResponsePropagator.java
+++ b/components/client/src/main/java/com/hotels/styx/client/netty/NettyToStyxResponsePropagator.java
@@ -24,6 +24,7 @@ import com.hotels.styx.api.client.Origin;
 import com.hotels.styx.api.netty.exceptions.ResponseTimeoutException;
 import com.hotels.styx.api.netty.exceptions.TransportLostException;
 import com.hotels.styx.client.BadHttpResponseException;
+import com.hotels.styx.client.StyxClientError;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
 import io.netty.channel.ChannelHandlerContext;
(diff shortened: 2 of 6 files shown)


components/client/src/main/java/com/hotels/styx/client/netty/NettyToStyxResponsePropagator.java
```diff
@@ -104,7 +105,15 @@
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
         ensureContentProducerIsCreated(ctx);
-        contentProducer.ifPresent(producer -> producer.channelException(new BadHttpResponseException(origin, cause)));
+        contentProducer.ifPresent(producer -> producer.channelException(toStyxException(cause)));
+    }
+
+    private RuntimeException toStyxException(Throwable cause) {
+        if (cause instanceof OutOfMemoryError) {
+            return new StyxClientError("Styx Client: Not enough memory.", cause);
```
kvosper: Perhaps we can make the stack trace more readable by specifying which kind of memory we have exhausted, since this condition will cover all kinds. The easiest way is probably to append the message of the parent error.

components/client/src/main/java/com/hotels/styx/client/StyxClientError.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Signifies an unexpected error from StyxClient.
+ *
+ * A root cause for the error is stored in the `cause` field.
+ * This could be thrown for example if StyxClient runs out of
+ * direct memory.
+ *
+ */
+public class StyxClientError extends RuntimeException {
```
kvosper: In Java, an Error and an Exception are not the same thing. We should ensure that our throwable class names match their parent types.

components/client/src/test/unit/java/com/hotels/styx/client/netty/NettyToStyxResponsePropagatorTest.java
```diff
@@ -268,6 +272,27 @@ public void shouldReleaseAlreadyReadBufferInCaseOfError() throws Exception {
...
+    public void mapsOutOfDirectMemoryExceptionsToResourceExhaustedException() throws Exception {
+        EmbeddedChannel channel = new EmbeddedChannel(new NettyToStyxResponsePropagator(responseSubscriber, SOME_ORIGIN));
+        channel.writeInbound(new DefaultHttpResponse(HTTP_1_1, OK));
+
+        channel.pipeline().fireExceptionCaught(newOutOfDirectMemoryError("Simulated out of direct memory error in a test."));
+
+        assertThat(responseSubscriber.getOnErrorEvents().get(0), is(instanceOf(StyxClientError.class)));
+    }
+
+    private OutOfDirectMemoryError newOutOfDirectMemoryError(String message) throws InstantiationException, IllegalAccessException, java.lang.reflect.InvocationTargetException {
```
kvosper: Perhaps a brief comment to explain that this is because OutOfDirectMemoryError has package-private visibility.
Also, `InvocationTargetException` does not need to be fully-qualified.

components/client/src/test/unit/java/com/hotels/styx/client/netty/NettyToStyxResponsePropagatorTest.java
```diff
@@ -268,6 +272,27 @@ public void shouldReleaseAlreadyReadBufferInCaseOfError() throws Exception {
...
+        Constructor<OutOfDirectMemoryError> constructor = null;
+        try {
+            constructor = OutOfDirectMemoryError.class.getDeclaredConstructor(String.class);
+        } catch (NoSuchMethodException e) {
+            Throwables.propagate(e);
```
kvosper: The flow of the code would be more clear to unfamiliar observers if we wrote `throw ` before `Throwables.propagate(e);` (but the functionality would remain the same)

components/client/src/test/unit/java/com/hotels/styx/client/netty/NettyToStyxResponsePropagatorTest.java
```diff
@@ -268,6 +272,27 @@ public void shouldReleaseAlreadyReadBufferInCaseOfError() throws Exception {
...
+        }
+        constructor.setAccessible(true);
+        return constructor.newInstance("message");
```
kvosper: I think this should be the variable `message`, not the string `"message"`.

PR #27, opened 2017-11-14
Styx server tls protocol

mikkokar:
Adds a *protocols* config option to Styx HTTPS server connector to specify the accepted TLS protocol versions. This is to allow operators to enforce TLS 1.2 support.

End-to-end test enhancements:
- Allow styx server to be configured with YAML string alone.
- Automatically allocate styx server port when a port number 0 is used.

### DIFF
--- a/system-tests/e2e-suite/src/test/scala/com/hotels/styx/StyxServerSupport.scala
+++ b/system-tests/e2e-suite/src/test/scala/com/hotels/styx/StyxServerSupport.scala
@@ -122,20 +122,21 @@ trait StyxServerSupplements {
 
   def adminURL(path: String) = s"http://$adminHost$path"
 
-  def secureHttpPort = if(styxServer.proxyHttpsAddress() != null) {
-    styxServer.proxyHttpsAddress().getPort
-  } else {
-    -1
-  }
+  def secureHttpPort = portNumberOrElse(styxServer.proxyHttpsAddress())
 
-  def httpPort = styxServer.proxyHttpAddress().getPort
+  def httpPort = portNumberOrElse(styxServer.proxyHttpAddress())
 
-  def adminPort = styxServer.adminHttpAddress().getPort
+  def adminPort = portNumberOrElse(styxServer.adminHttpAddress())
 
   def metricsSnapshot = {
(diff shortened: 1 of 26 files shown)


system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/https/EmptyTlsProtocolsListSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+  override protected def afterAll() = {
+    recordingBackend.stop()
+    super.afterAll()
+  }
+
+  describe("Empty TLS protocols list") {
+    recordingBackend.stub(urlPathEqualTo("/secure"), aResponse.withStatus(200))
+
+    it("Empty TLS protocols list activates all supported protocols") {
+      val req = new HttpRequest.Builder(GET, styxServer.secureRouterURL("/secure"))
```
kvosper: Why not simply import `HttpRequest.Builder.get` instead of using this verbose alternative?

system-tests/e2e-suite/src/test/scala/com/hotels/styx/StyxServerSupport.scala
```diff
@@ -122,16 +122,21 @@ trait StyxServerSupplements {
...
 
-  def adminPort = styxServer.adminHttpAddress().getPort
+  def adminPort = portNumberOrElse(styxServer.adminHttpAddress())
 
   def metricsSnapshot = {
     val adminHostName = styxServer.adminHttpAddress().getHostName
     new CodaHaleMetricsFacade(StyxMetrics.downloadFrom(adminHostName, adminPort))
   }
+
+  private def portNumberOrElse(address: InetSocketAddress) = Option(address)
```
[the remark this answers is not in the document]
mikkokar: It means the port has not been configured.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/https/EmptyTlsProtocolsListSpec.scala
mikkokar: Addressed all comments. Please have another look.

PR #29, opened 2017-11-17
Implement Wiremock API on top of Styx/Netty HTTP servers.

mikkokar:
Implements a MockOriginServer class which implements a Wiremock mocking API on top of Styx/Netty HTTP servers.

Why is this necessary? Wiremock API is very useful for mocking application layer behaviour and RESTful APIs. However it is not that useful for testing lower level networking and protocol behaviour. With this PR we are able to marry the two aspects into a single class. 

Benefits

* My immediate use-case is to force test origins to require TLS1.2 from incoming connections, and to reject all other versions.

* Currently we have different test origin implementations depending on the use case. Hopefully with this we can reduce the number of different test server classes and consolidate then into one.


system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/MockOriginServer.java
```diff
@@ -0,0 +1,218 @@
...
+    private MockOriginServer(String appId, String originId, int adminPort, int serverPort, HttpServer adminServer, HttpServer mockServer) {
+        this.appId = appId;
+        this.originId = originId;
+        this.adminPort = adminPort;
+        this.serverPort = serverPort;
+        this.adminServer = adminServer;
+        this.mockServer = mockServer;
+    }
+
+    private static WireMockApp wireMockApp() {
```
kvosper: Could we move this method further down the class? I think it makes more sense for public methods to be the first thing the viewer sees, as they convey how the class can be used.

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/MockOriginServer.java
```diff
@@ -0,0 +1,218 @@
...
+
+    public static MockOriginServer create(String appId, String originId, int adminPort, HttpsConnectorConfig httpsConfig) {
+        WireMockApp wireMockApp = wireMockApp();
+        HttpServer adminServer = createAdminServer(originId, adminPort, wireMockApp);
+        HttpServer mockServer = HttpServers.createHttpsServer(
+                "mock-stub-" + originId,
+                httpsConfig,
+                mockHandler(wireMockApp, new WireMockConfiguration()));
+        int serverPort = httpsConfig.port();
+        System.out.println(format("Creating HTTPS MockOriginServer - %d, %d", adminPort, serverPort));
```
kvosper: If this is required it should use the `LOGGER`. `System.out.println` usually causes checkstyle violations - was this not the case here?

Also, if we log for HTTPS, we should log for HTTP too.

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/MockOriginServer.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    private static HttpServer createAdminServer(String originId, int adminPort, WireMockApp wireMockApp) {
+        return HttpServers.createHttpServer(
+                "mock-admin-" + originId,
+                new HttpConnectorConfig(adminPort),
+                adminHandler(wireMockApp));
+    }
+
+    public static MockOriginServer create(String appId, String originId, int adminPort, HttpConnectorConfig httpConfig) {
```
kvosper: There is a lot of code duplication between the two `create` methods - is it possible to neatly refactor them to reduce this?

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/MockOriginServer.java
```diff
@@ -0,0 +1,218 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+import static com.google.common.base.Optional.absent;
+import static com.hotels.styx.servers.WiremockResponseConverter.toStyxResponse;
+import static java.lang.String.format;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Collections.emptyMap;
+import static rx.Observable.just;
+
+public final class MockOriginServer {
+    private static final String FILES_ROOT = "__files";
```
kvosper: This constant can be imported from `WireMockServer.FILES_ROOT`, it needn't be copied here.

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/WiremockResponseConverter.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.github.tomakehurst.wiremock.http.Response;
+import com.hotels.styx.api.HttpResponse;
+import io.netty.buffer.ByteBuf;
+import io.netty.handler.codec.http.HttpResponseStatus;
+
+import static io.netty.buffer.Unpooled.EMPTY_BUFFER;
+import static io.netty.buffer.Unpooled.copiedBuffer;
+
+public class WiremockResponseConverter {
```
kvosper: Classes with only static methods should be final, with a private constructor.

kvosper: Also, this class could be package-private, will access be needed from outside this package in the future?

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/WiremockResponseConverter.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public static HttpResponse toStyxResponse(Response response) {
+        HttpResponseStatus status = HttpResponseStatus.valueOf(response.getStatus());
+        com.hotels.styx.api.HttpHeaders headers = toStyxHeaders(response.getHeaders());
+        ByteBuf content = toNettyContent(response);
+
+        return HttpResponse.Builder.response(status).headers(headers).body(content).build();
```
kvosper: Please use static imports. There is no need to fully-qualify `HttpResponse.Builder.response`.

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/WiremockStyxRequestAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+
+import static io.netty.handler.codec.http.HttpHeaderNames.CONTENT_TYPE;
+import static io.netty.handler.codec.http.HttpHeaderNames.HOST;
+
+public class WiremockStyxRequestAdapter implements Request {
+    private final HttpRequest styxRequest;
+    private final String body;
+
+    public WiremockStyxRequestAdapter(HttpRequest styxRequest, String body) {
+        this.styxRequest = styxRequest;
```
kvosper: Please please use null checks in constructors with `Objects.requireNonNull`.

Null checks provide very helpful fail-fast behaviour when used this way.

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/WiremockStyxRequestAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+    @Override
+    public ContentTypeHeader contentTypeHeader() {
+        return styxRequest.header(CONTENT_TYPE)
+                .map(ContentTypeHeader::new)
+                .orElse(ContentTypeHeader.absent());
+    }
+
+    @Override
+    public HttpHeaders getHeaders() {
+        List<HttpHeader> list = StreamSupport.stream(styxRequest.headers().spliterator(), false)
```
kvosper: Please use static-imports to reduce verbosity. `stream`, `httpHeader` and `toList` can all be statically-imported.

(Note that we will accept verbosity in the case of something like `ImmutableList.of` where `of` does not tell us anything by itself).

system-tests/e2e-testsupport/src/test/java/com/hotels/styx/servers/MockOriginServerTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+    private ValueMatchingStrategy valueMatchingStrategy(String matches) {
+        ValueMatchingStrategy strategy = new ValueMatchingStrategy();
+        strategy.setMatches(matches);
+        return strategy;
+    }
+
+    private HostnameVerifier disableHostNameVerification() {
+        HostnameVerifier old = HttpsURLConnection.getDefaultHostnameVerifier();
+        HttpsURLConnection.setDefaultHostnameVerifier(new HostnameVerifier() {
```
kvosper: Please statically import `setDefaultHostnameVerifier`.
Also, the `HostnameVerifier` instantiated here can be a lambda.

system-tests/e2e-testsupport/src/test/java/com/hotels/styx/servers/MockOriginServerTest.java
```diff
@@ -0,0 +1,144 @@
...
+                .stub(WireMock.get(urlMatching("/.*")), aResponse()
+                        .withStatus(200)
+                        .withHeader("a", "b")
+                        .withBody("Hello, World!"));
+
+        HttpResponse response = send(client,
+                get(format("http://localhost:%d/mock", serverPort))
+                        .header("X-Forwarded-Proto", "http")
+                        .build())
+                .responseBuilder()
```
kvosper: Instead of calling ` .responseBuilder().build()` in your test methods, these could be moved inside your `send` method. 
Note also that the responseBuilder has no content by default, which means the response body has been removed. I recommend either fixing this and adding an assertion to verify, or remove the `.withBody("Hello, World!")` from the mock response.

system-tests/e2e-testsupport/src/test/java/com/hotels/styx/servers/WiremockStyxRequestAdapterTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    @Test
+    public void adaptsContainsHeader() throws Exception {
+        assertThat(adapter.containsHeader("Foo-Bar"), is(false));
+        assertThat(adapter.containsHeader("Host"), is(true));
+        assertThat(adapter.containsHeader("host"), is(true));
+    }
+
+    @Test(enabled = false)
```
kvosper: Disabled tests should not be added to the codebase.

system-tests/e2e-testsupport/src/test/java/com/hotels/styx/servers/WiremockStyxRequestAdapterTest.java
```diff
@@ -0,0 +1,179 @@
...
+        assertThat(contentType.encodingPart(), is(Optional.of("UTF-8")));
+        assertThat(contentType.mimeTypePart(), is("application/json"));
+    }
+
+    @Test
+    public void adaptsQueryParameter() throws Exception {
+        QueryParameter msg = adapter.queryParameter("msg");
+
+        assertThat(msg.key(), is("msg"));
+        assertThat(msg.values(), is(ImmutableList.of("6198.1")));
```
kvosper: Please use `Matchers.contains` (statically imported, of course) to verify list contents.

system-tests/e2e-testsupport/src/test/java/com/hotels/styx/servers/WiremockStyxRequestAdapterTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void adaptsNonExistantQueryParameterToNull() throws Exception {
+        QueryParameter msg = adapter.queryParameter("foobar");
+
+        assertThat(msg.key(), is("foobar"));
+        assertThat(msg.values(), is(Collections.emptyList()));
```
kvosper: There is no need for the assertion around `msg.values()` since we're testing that method for returning an exception.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/support/configuration/StyxBackend.scala
```diff
@@ -33,6 +34,12 @@ trait ImplicitOriginConversions {
     host = "localhost",
     port = fakeServer.port())
 
+  implicit def mockOrigin2Origin(server: MockOriginServer): Origin = Origin(
```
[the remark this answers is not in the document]
mikkokar: It constructs an `Origin` object from `MockOriginServer` instance. That is, a function from MockOrigin to Origin. Number 2 is often used in place of word "to" because they sound similar. Using it here for the sake of consistency. The other conversion routines in the class follow similar naming.

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/MockOriginServer.java
```diff
@@ -0,0 +1,218 @@
...
+    private final String appId;
+    private final String originId;
+    private final int adminPort;
+
+    private int serverPort;
+    private final HttpServer adminServer;
+    private final HttpServer mockServer;
+
+    static {
+        System.setProperty("org.mortbay.log.class", "com.github.tomakehurst.wiremock.jetty.LoggerAdapter");
```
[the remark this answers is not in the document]
mikkokar: Hmm... TBH I am not sure what this is supposed to do. I believe it can be removed. However I don't know if this would affect any logging originated by the Wiremock classes. 

We should follow up later.

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/MockOriginServer.java
mikkokar: Ok. I have moved all private methods to the end of the class.

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/MockOriginServer.java
mikkokar: The bulkiest part of the code is in the middle: a call to static `createHttpServer`, or `createHttpsServer`. To have a meaningful outcome we should look into refactoring those two static methods instead.

Personally I am happy with this code as it is, as I need this PR in for additional TLS related features. I would refactor in future once we'll have a more pressing reason to do so. But please feel free to contribute if you have some good ideas.

system-tests/e2e-testsupport/src/main/java/com/hotels/styx/servers/MockOriginServer.java
mikkokar: I have removed. The print line statement t shouldn't have been there.

system-tests/e2e-testsupport/src/test/java/com/hotels/styx/servers/WiremockStyxRequestAdapterTest.java
mikkokar: I have fixed the tests.

kvosper: Except for the one directly above these comments.

PR #36, opened 2017-11-23
Add cipher suite controls to backend services configuration.

mikkokar:


components/client/src/main/java/com/hotels/styx/client/ssl/SslContextFactory.java
```diff
@@ -68,9 +68,18 @@ private static SslContext createSslContext(TlsSettings tlsSettings) throws IOExc
                 .sslProvider(SslProvider.valueOf(tlsSettings.sslProvider()))
                 .trustManager(trustManagerFactory(tlsSettings))
                 .protocols(toNettyProtocols(tlsSettings.protocols()))
+                .ciphers(toNettyCiphers(tlsSettings.cipherSuites()))
                 .build();
     }
 
+    private static List<String> toNettyCiphers(List<String> strings) {
+        if (strings == null || strings.isEmpty()) {
```
alobodzki: ImmutableList.copyOf(..) would throw NPE if this was null

PR #40, opened 2017-11-27
Add FullHttpRequest and FullHttpResponse classes to Styx API

kvosper:
These are some classes representing full (aggregated/decoded) HTTP messages.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
+++ b/components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
@@ -1,0 +1,622 @@
+/**
+ * Copyright (C) 2013-2017 Expedia Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.styx.api.messages;
+
+import com.google.common.collect.ImmutableList;
+import com.hotels.styx.api.HttpCookie;
+import com.hotels.styx.api.HttpHeaders;
+import com.hotels.styx.api.HttpMessageSupport;
+import com.hotels.styx.api.Url;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.handler.codec.http.HttpMethod;
+import io.netty.handler.codec.http.HttpVersion;
+
+import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+
+import static com.google.common.base.Objects.toStringHelper;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.hotels.styx.api.HttpHeaderNames.CONNECTION;
+import static com.hotels.styx.api.HttpHeaderNames.CONTENT_LENGTH;
+import static com.hotels.styx.api.HttpHeaderNames.HOST;
+import static com.hotels.styx.api.HttpHeaderValues.KEEP_ALIVE;
+import static com.hotels.styx.api.messages.HttpSupport.encodeBody;
+import static com.hotels.styx.api.support.CookiesSupport.isCookieHeader;
+import static io.netty.handler.codec.http.HttpMethod.CONNECT;
+import static io.netty.handler.codec.http.HttpMethod.DELETE;
+import static io.netty.handler.codec.http.HttpMethod.GET;
+import static io.netty.handler.codec.http.HttpMethod.HEAD;
+import static io.netty.handler.codec.http.HttpMethod.OPTIONS;
+import static io.netty.handler.codec.http.HttpMethod.PATCH;
+import static io.netty.handler.codec.http.HttpMethod.POST;
+import static io.netty.handler.codec.http.HttpMethod.PUT;
+import static io.netty.handler.codec.http.HttpMethod.TRACE;
+import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
+import static java.lang.Integer.parseInt;
+import static java.net.InetSocketAddress.createUnresolved;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.requireNonNull;
+import static java.util.UUID.randomUUID;
+
+/**
+ * Request.
+ *
+ * @param <T> content type
+ */
+public class FullHttpRequest<T> implements FullHttpMessage<T> {
+    private final Object id;
+    private final InetSocketAddress clientAddress;
+    private final HttpVersion version;
+    private final HttpMethod method;
+    private final Url url;
+    private final HttpHeaders headers;
+    private final boolean secure;
+    private final T body;
+    private final List<HttpCookie> cookies;
+
+    FullHttpRequest(Builder<T> builder) {
+        this.id = builder.id();
+        this.clientAddress = builder.clientAddress();
+        this.version = builder.version();
+        this.method = builder.method();
+        this.url = builder.url();
+        this.secure = builder.secure();
+        this.headers = builder.headers();
+        this.body = builder.body();
+        this.cookies = builder.cookies();
+    }
+
+    @Override
+    public HttpVersion version() {
+        return this.version;
+    }
+
+    @Override
+    public HttpHeaders headers() {
+        return headers;
+    }
+
+    @Override
+    public List<HttpCookie> cookies() {
+        return cookies;
+    }
+
+    @Override
+    public List<String> headers(CharSequence name) {
+        return headers.getAll(name);
+    }
+
+    @Override
+    public T body() {
+        return body;
+    }
+
+    /**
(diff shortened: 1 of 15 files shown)


alobodzki:
I lookd at the change, looks good for me, Mikko?

components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
```diff
@@ -0,0 +1,694 @@
...
+
+    /**
+     * Creates a request with the GET method.
+     *
+     * @param uri URI
+     * @param body body
+     * @param <T> body type
+     * @return {@code this}
+     */
+    public static <T> Builder<T> get(String uri, T body) {
```
mikkokar: The GET and HEAD (below) methods are not intended for pushing content up to server. Therefore we should remove this overloaded variant and the another for `head(String, T)`.

components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
```diff
@@ -0,0 +1,694 @@
...
+
+    /**
+     * Creates a request with the DELETE method.
+     *
+     * @param uri URI
+     * @param body body
+     * @param <T> body type
+     * @return {@code this}
+     */
+    public static <T> Builder<T> delete(String uri, T body) {
```
mikkokar: I would remove this method as well. See my comments for `get(String, T)` above.

components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
```diff
@@ -0,0 +1,694 @@
...
+                .build();
+    }
+
+    /**
+     * Encodes a request with a body of type String into a streaming form, using a UTF-8 encoding.
+     *
+     * @param request a request
+     * @return an encoded (streaming) request
+     */
+    public static com.hotels.styx.api.HttpRequest toStreamingHttpRequest(FullHttpRequest<String> request) {
```
mikkokar: Could this be statically imported?

components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
```diff
@@ -0,0 +1,694 @@
...
+            this.clientAddress = request.clientAddress();
+            this.url = request.url();
+            this.secure = request.isSecure();
+            this.version = request.version();
+            this.headers = request.headers().newBuilder();
+            this.body = request.body();
+            this.cookies = new ArrayList<>(request.cookies());
+        }
+
+        private Builder(FullHttpRequest<?> request, boolean doNotCopyBody) {
```
mikkokar: Unused constructor.

components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
```diff
@@ -0,0 +1,694 @@
...
+         *
+         * @param uri URI
+         * @return {@code this}
+         */
+        public Builder<T> uri(String uri) {
+            return this.url(Url.Builder.url(uri).build());
+        }
+
+        public Builder<T> body(T content) {
+            setContentLength(content);
```
mikkokar: I would remove this for now. For the following reasons:
- The API consumer may want to set `Transfer-Encoding: Chunked` instead.
- For testing purposes, etc we may want to create HTTP messages which *deliberately do not* have content length set. In fact this is a perfectly valid use case for servers returning HttpResponses.
- The content length depends on the kind of encodings applied to the body, and it should be in-line with the other headers applied such as content type charset etc. This method doesn't attempt to set those headers, and also I am not confident the length calculation works in all situations.

components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
```diff
@@ -0,0 +1,694 @@
...
+            return header(CONNECTION, KEEP_ALIVE);
+        }
+
+        public FullHttpRequest<T> build() {
+            if (validate) {
+                ensureContentLengthIsValid();
+                ensureMethodIsValid();
+            }
+
+            setHostHeader();
```
mikkokar: This too will prevent some use cases where we deliberately might *not* want to have the host header present. Such as adapting wire-mock onto Styx server framework. Perhaps move into the *validate* block or something?

components/api/src/test/java/com/hotels/styx/api/messages/FullHttpResponseTest.java
```diff
@@ -0,0 +1,405 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.emptyIterable;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.collection.IsIterableContainingInOrder.contains;
+
+public class FullHttpResponseTest {
+    @Test
```
mikkokar: We need to have a test to ensure that all headers, status & other response attributes are converted correctly to streaming type:

eg.

```
assertThat(streaming.status(), is(response.status()));
assertThat(streaming.headers(), is(response.headers());
...
```

components/api/src/test/java/com/hotels/styx/api/messages/FullHttpResponseTest.java
```diff
@@ -0,0 +1,405 @@
...
+        HttpResponse streaming = response.toStreamingHttpResponse(string -> copiedBuffer(string, UTF_8));
+
+        TestSubscriber<ByteBuf> subscriber = TestSubscriber.create(0);
+        subscriber.requestMore(1);
+
+        streaming.body().content().subscribe(subscriber);
+
+        assertThat(subscriber.getOnNextEvents().size(), is(1));
+        ByteBuf buf = subscriber.getOnNextEvents().get(0);
+        assertThat(buf.toString(UTF_8), is("foobar"));
```
mikkokar: Also, it should be sufficient to use `streaming.body().decode(...)` to aggregate the content stream for assertions. 

IMO the fact that it is streamed as a single content chunk should be an implementation detail, not part of the method contract (assertion in line 77). That is, we could at any time choose to change the `toStreamingHttpResponse` implementation to split it into several smaller chunks if necessary without affecting the API contract.

Perhaps this could be even clarified in the javadoc comments.

components/api/src/test/java/com/hotels/styx/api/messages/FullHttpResponseTest.java
```diff
@@ -0,0 +1,405 @@
...
+        assertThat(response.body(), is(nullValue()));
+    }
+
+    @Test
+    public void setsTheContentType() {
+        assertThat(response().contentType(ANY_AUDIO_TYPE).build().contentType(), isValue(ANY_AUDIO_TYPE.toString()));
+    }
+
+    @Test
+    public void setsASingleOutboundCookie() throws Exception {
```
mikkokar: IntelliJ says the the Exception is never thrown.

components/api/src/test/java/com/hotels/styx/api/messages/FullHttpRequestTest.java
```diff
@@ -0,0 +1,442 @@
...
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+import static rx.Observable.just;
+
+
+public class FullHttpRequestTest {
+    @Test
+    public void encodesToStreamingHttpRequest() throws Exception {
```
mikkokar: IntelliJ says the Exception is never thrown.

mikkokar: Here we need to have tests to demonstrate that all HTTP request attributes (URL, path, protocol, headers) are converted accordingly by the call to `toStreamingHttpRequest()`. Not just the request body.

mikkokar: It should be sufficient to use `streaming.body().decode(...)` for content assertions. See my comments for the FullHttpResponseTest.

mikkokar: Copying my comment from the FullHttpResponseTest:

> IMO the fact that it is streamed as a single content chunk should be an implementation detail, not part of the method contract (assertion in line 74 below). That is, we could at any time choose to change the toStreamingHttpResponse implementation to split it into several smaller chunks if necessary without affecting the API contract.

components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
```diff
@@ -0,0 +1,680 @@
...
+         * Sets the (only) value for the header with the specified name.
+         * <p/>
+         * All existing values for the same header will be removed.
+         *
+         * @param name  The name of the header
+         * @param value The value of the header
+         * @return {@code this}
+         */
+        public Builder<T> header(CharSequence name, Object value) {
+            checkNotCookie(name);
```
mikkokar: The `checkNotCookie(name)` should be present only when validation is set to true.

We need to come up with a better approach for representing cookies. Going forward, we should build the structured cookie API on top of textual HTTP headers API, making it complementary, instead of enforcing its usage. 

This is important because enforcing structured cookie API restricts the applicability of the API (Think of the WireMock use case). Therefore we need to let API consumers to set/read cookies directly via headers.

For now, make the `checkNotCookie()` somehow optional, but leave any further API improvements to the next increment.

mikkokar: This comment also `addHeader(CharSequence, Object)` variant.

mikkokar: Hmm... looking at code. This alone is not sufficient. Let me think about this a bit.

PR #43, opened 2017-12-06
Stop NullPointerExceptions in HttpRequestMessageLogger

kvosper:


components/common/src/main/java/com/hotels/styx/common/logging/HttpRequestMessageLogger.java
```diff
@@ -36,11 +35,23 @@ public HttpRequestMessageLogger(String name, boolean longFormatEnabled) {
...
+        }
     }
 
     public void logResponse(HttpRequest request, HttpResponse response) {
-            logger.info("requestId={}, response={}", request.id(), information(response, longFormatEnabled));
+        if (response == null) {
+            logger.warn("Unable to log null response, request={}", request);
+        } else {
+            logger.info("requestId={}, response={}", id(request), information(response, longFormatEnabled));
+        }
```
mikkokar: Have you investigated how we end up calling `logResponse` and/or `logRequest` methods with null arguments?

components/common/src/main/java/com/hotels/styx/common/logging/HttpRequestMessageLogger.java
```diff
@@ -36,11 +35,23 @@ public HttpRequestMessageLogger(String name, boolean longFormatEnabled) {
...
+            logger.info("requestId={}, request={}", request.id(), information(request, origin, longFormatEnabled));
+        }
     }
 
     public void logResponse(HttpRequest request, HttpResponse response) {
-            logger.info("requestId={}, response={}", request.id(), information(response, longFormatEnabled));
+        if (response == null) {
+            logger.warn("Unable to log null response, request={}", request);
+        } else {
+            logger.info("requestId={}, response={}", id(request), information(response, longFormatEnabled));
```
mikkokar: The two messages should be somewhat consistent. Now we have:

> "Unable to log null response, request={...}"

vs

> "requestId={...}, response={...}, requestId, ...."

For consistency, when response==null, we would instead log:

> "requestId=NA, respone={...}"

But if it is possible for request to be `null`, it does beg a question can request also be sometimes `null`?

components/common/src/main/java/com/hotels/styx/common/logging/HttpRequestMessageLogger.java
```diff
@@ -36,11 +35,23 @@ public HttpRequestMessageLogger(String name, boolean longFormatEnabled) {
...
+            logger.warn("requestId=N/A, request=null, origin={}", origin);
+        } else {
+            logger.info("requestId={}, request={}", request.id(), information(request, origin, longFormatEnabled));
+        }
     }
 
     public void logResponse(HttpRequest request, HttpResponse response) {
-            logger.info("requestId={}, response={}", request.id(), information(response, longFormatEnabled));
+        if (response == null) {
+            logger.warn("requestId={}, response=null", id(request));
```
mikkokar: Hi @kvosper the warning on line 39 uses "N/A" in place of "null", but the warning on line 47 uses "null".
Let's be consistent with this and either use "N/A" or "null" everywhere, but let's not mix and match.

PR #44, opened 2017-12-07
Add streaming classes

kvosper:
Streaming classes to accompany the FullHttpRequest and FullHttpResponse, these can help us move towards an improved API in a later major version.


components/api/src/main/java/com/hotels/styx/api/messages/StreamingHttpRequest.java
```diff
@@ -0,0 +1,693 @@
...
+    public boolean isSecure() {
+        return secure;
+    }
+
+    /**
+     * Returns the remote client address that initiated the current request.
+     *
+     * @return the client address for this request
+     */
+    public InetSocketAddress clientAddress() {
```
mikkokar: Client address is conceptually not part of HTTP request/response objects. Therefore we should not have it our HTTP request abstraction. Please remove this. 

I recognise it appears in the existing HttpRequest/Response objects. So for the sake of backwards compatibility we can keep it as a private filed so the requests/responses can be converted between the formats without losing information. But nevertheless we should not have this as a part of public API.

The clientAddress should be exposed to the interceptors as a part of interceptor context instead.

components/api/src/main/java/com/hotels/styx/api/messages/StreamingHttpRequest.java
```diff
@@ -0,0 +1,693 @@
...
+     * This allows a new request to be made that will be identical to this one except for the properties
+     * overridden by the builder methods.
+     *
+     * @return new builder based on this request
+     */
+    public Builder newBuilder() {
+        return new Builder(this);
+    }
+
+    public <T> Observable<FullHttpRequest<T>> toFullHttpRequest(Function<ByteBuf, T> decoder, int maxContentBytes) {
```
mikkokar: See the current implementation of `FullHttpRequest`. For the latest developments consult the master branch. It has been merged as a part of PR #50.

In effect the `decoder` argument can now be removed.

components/api/src/main/java/com/hotels/styx/api/messages/StreamingHttpResponse.java
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    /**
+     * Creates an HTTP response builder with a given status and body.
+     *
+     * @param status response status
+     * @param body response body
+     * @return a new builder
+     */
+    public static Builder response(HttpResponseStatus status, Observable<ByteBuf> body) {
```
mikkokar: Is this a Netty `HttpResponseStatus`? We should aim to minimise Netty dependencies. Ideally use an `int` or introduce a Styx response status, but apart from ByteBuf let's keep this clean of Netty definitions.

Having said this I now realise this comment might also apply to FullHttpRequest/Response classes too.

components/api/src/main/java/com/hotels/styx/api/messages/StreamingHttpResponse.java
```diff
@@ -0,0 +1,448 @@
...
+
+    public HttpResponseStatus status() {
+        return status;
+    }
+
+    public boolean isRedirect() {
+        return status().code() >= 300 && status().code() < 400;
+    }
+
+    public <T> Observable<FullHttpResponse<T>> toFullHttpResponse(Function<ByteBuf, T> decoder, int maxContentBytes) {
```
mikkokar: Update as per latest `FullHttpResponse` without `T` type parameter. Essentially, the `decoder` parameter ought to be removed as unnecessary.

components/api/src/main/java/com/hotels/styx/api/messages/StreamingHttpResponse.java
```diff
@@ -0,0 +1,448 @@
...
+            this.cookies = new ArrayList<>(response.cookies());
+        }
+
+        /**
+         * Sets the response status.
+         *
+         * @param status response status
+         * @return {@code this}
+         */
+        public Builder status(HttpResponseStatus status) {
```
mikkokar: Netty class `HttpResponseStatus`?

components/api/src/main/java/com/hotels/styx/api/messages/StreamingHttpResponse.java
```diff
@@ -0,0 +1,448 @@
...
+            return this;
+        }
+
+        /**
+         * Sets the value of the {@code Content-Type} header.
+         * @param contentType content type
+         *
+         * @return {@code this}
+         */
+        public Builder contentType(MediaType contentType) {
```
mikkokar: Guava type appears in Styx API?

This comment probably applies to existing Full message types too.

components/api/src/main/java/com/hotels/styx/api/messages/StreamingHttpResponse.java
kvosper: I will change the status as part of a separate branch as it will affect a very large number of files.

components/api/src/main/java/com/hotels/styx/api/messages/FullHttpRequest.java
```diff
@@ -366,14 +359,26 @@ public Builder() {
             this.cookies = new ArrayList<>();
         }
 
-        public Builder(HttpMethod method, String uri) {
+        public Builder(String method, String uri) {
```
mikkokar: It would be better to introduce a `HttpMethod` type. It would be generally more type safe approach and ensure that only specific values of the `method` are meaningful. Not to mention it would contribute towards more self-commenting code.

components/api/src/main/java/com/hotels/styx/api/messages/FullHttpResponse.java
```diff
@@ -134,12 +133,12 @@ public Builder newBuilder() {
         return new Builder(this);
     }
 
-    public HttpResponseStatus status() {
+    public int status() {
```
mikkokar: This is a good step forward to remove Netty & Google dependencies. However I would still prefer to represent HTTP status code as a separate class rather than just `int`. This is because a class can hold both the status code number and the textual description in one unit, which will pay dividends for example in logging:

say:

` println("status=%s", response.status())`

Which would now show just `status=200` instead of `status=200 OK`.

components/api/src/main/java/com/hotels/styx/api/messages/HttpResponseStatus.java
```diff
@@ -0,0 +1,135 @@
...
+    public static final HttpResponseStatus SERVICE_UNAVAILABLE = new HttpResponseStatus(503, "Service Unavailable");
+    public static final HttpResponseStatus GATEWAY_TIMEOUT = new HttpResponseStatus(504, "Gateway Timeout");
+    public static final HttpResponseStatus HTTP_VERSION_NOT_SUPPORTED = new HttpResponseStatus(505, "HTTP Version Not Supported");
+    public static final HttpResponseStatus VARIANT_ALSO_NEGOTIATES = new HttpResponseStatus(506, "Variant Also Negotiates");
+    public static final HttpResponseStatus INSUFFICIENT_STORAGE = new HttpResponseStatus(507, "Insufficient Storage");
+    public static final HttpResponseStatus NOT_EXTENDED = new HttpResponseStatus(510, "Not Extended");
+    public static final HttpResponseStatus NETWORK_AUTHENTICATION_REQUIRED = new HttpResponseStatus(511, "Network Authentication Required");
+
+    private final int code;
+    private final String name;
```
mikkokar: IANA calls this `description`. So should we.

Ref: https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

components/api/src/main/java/com/hotels/styx/api/messages/HttpResponseStatus.java
```diff
@@ -0,0 +1,135 @@
...
+    private HttpResponseStatus(int code, String name) {
+        this.code = code;
+        this.name = name;
+    }
+
+    public int code() {
+        return code;
+    }
+
+    public String name() {
```
mikkokar: Let's call this `description`.

components/api/src/main/java/com/hotels/styx/api/messages/HttpResponseStatus.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+package com.hotels.styx.api.messages;
+
+import java.lang.reflect.Field;
+import java.util.stream.Stream;
+
+import static java.lang.reflect.Modifier.isStatic;
+
+/**
+ * Contains all HTTP response status codes as HttpResponseStatusCodesCodeseger constants as well as a method to go from a code to its name.
```
mikkokar: Please split this long comment on two lines.

PR #46, opened 2017-12-12
Splitting StyxHttpClient functionality into two separate logically ty…

alobodzki:
Splitting StyxHttpClient functionality into two separate logically tYpes. It is a first step in a refactoring process of decoupling the logic.


mikkokar:
Not necessary.

PR #48, opened 2017-12-14
String only full messages

kvosper:
Simplify body type of FullHttpRequest/FullHttpResponse to just String.


mikkokar:
Opened a separate PR with additional changes.

PR #49, opened 2017-12-15
Passing a dependency array into service factories

alobodzki:


components/api/src/main/java/com/hotels/styx/api/client/loadbalancing/spi/LoadBalancingStrategy.java
```diff
@@ -60,11 +62,20 @@ default double oneMinuteRateForStatusCode5xx(Origin origin) {
...
      * @return the origins sorted into a new order
      */
-    Iterable<ConnectionPool> vote(Iterable<ConnectionPool> origins, Context context);
+    Iterable<ConnectionPool> vote(Context context);
+
+    /**
+     * Returns a collection of {@link ConnectionPool}, without any guarantee of ordering.
+     * @return available, unordered origins
+     */
+    default Iterable<ConnectionPool> snapshot() {
```
mikkokar: Is this necessary only in the RetryPolicyContext? In that case please remove the `snapshot()` method, and either look into populating the `RetryPolicyContext` directly from the origins inventory. Given that value retry mechanism provides is so questionable I don't mind if we remove that feature for now at least until the more important refactoring are finished.

components/api/src/main/java/com/hotels/styx/api/client/loadbalancing/spi/LoadBalancingStrategyFactory.java
```diff
@@ -26,12 +27,24 @@
...
 public interface LoadBalancingStrategyFactory extends ServiceFactory<LoadBalancingStrategy> {
+
+    /**
+     * Creates a strategy.
+     *
+     * @param environment           Styx application environment
+     * @param strategyConfiguration configuration specific to load balancer
+     * @return a new load balancing strategy.
+     */
+    default LoadBalancingStrategy create(Environment environment, Configuration strategyConfiguration) {
```
mikkokar: Please remove this. We only need `create(Environment, Configuration, ActivieOrigins)` variant.

components/api/src/main/java/com/hotels/styx/api/configuration/ServiceFactory.java
```diff
@@ -32,4 +32,16 @@
...
+
+    /**
+     * Create an instance of the product.
+     *
+     * @param environment                 environment
+     * @param serviceConfiguration configuration specific to the factory product
+     * @param objects objects that service might depend on
+     * @return product instance
+     */
+    default E create(Environment environment, Configuration serviceConfiguration, Object... objects) {
```
mikkokar: Let's get rid of the `Objects...` array. However I suppose this is also used for the styx "server" class loading? In that case just duplicate this interface for now. Create a separate variant for "StyxClientService" etc which always takes an OriginInventory as a third parameter. Going forward we will remove the duplicated interface in the next increment. But at the moment we'll need running code.

components/api/src/main/java/com/hotels/styx/api/HttpClient.java
```diff
@@ -34,10 +34,4 @@
      *
      */
     Observable<HttpResponse> sendRequest(HttpRequest request);
-
-    default void close() {
-    }
-
-    default void registerStatusGauges() {
```
mikkokar: Good to see these being removed.

mikkokar:
Retargeting this to `spi-refactor` branch. Good work in decoupling origin inventory from client. However need to carry out further verifications to ensure backwards compatibility. Additional work to improve routing and SPI interface can be based on the `spi-refactor` branch while the verifications are pending.

PR #50, opened 2017-12-19
Store Full HTTP message content as byte array.

mikkokar:
Remove type argument T from the Full HTTP messages. Store the content instead as byte[].


mikkokar:
Remove type argument T from the Full HTTP messages. Store the content instead as byte[].

PR #58, opened 2018-01-09
Moving underlying request send logic from client implementstions to connections.

alobodzki:
This allows to decouple clients from particular implementation choices.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/client/Connection.java
+++ b/components/api/src/main/java/com/hotels/styx/api/client/Connection.java
@@ -21,7 +21,6 @@ import rx.Observable;
 
 import java.io.Closeable;
 import java.util.EventListener;
-import java.util.function.Supplier;
 
 /**
  * A connection to an origin.
@@ -60,15 +59,6 @@ public interface Connection extends Closeable {
          */
         Observable<Connection> createConnection(Origin origin, Settings connectionSettings);
     }
-
-    /**
-     * Executes an operation in the context of this connection, allowing it to collect metrics, handle errors, etc.
-     *
-     * @param operation an operation
-     * @param <R>       result type
-     * @return an observable that provides the operation result
-     */
-    <R> Observable<R> execute(Supplier<Observable<R>> operation);
 
     /**
      * Writes HTTP request to a remote peer in the context of this connection.
--- a/components/api/src/main/java/com/hotels/styx/api/client/Connection.java
+++ b/components/api/src/main/java/com/hotels/styx/api/client/Connection.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2013-2017 Expedia Inc.
+ * Copyright (C) 2013-2018 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,8 @@
  */
 package com.hotels.styx.api.client;
 
+import com.hotels.styx.api.HttpRequest;
+import com.hotels.styx.api.HttpResponse;
 import rx.Observable;
 
 import java.io.Closeable;
@@ -67,6 +69,14 @@ public interface Connection extends Closeable {
      * @return an observable that provides the operation result
      */
     <R> Observable<R> execute(Supplier<Observable<R>> operation);
+
+    /**
+     * Writes HTTP request to a remote peer in the context of this connection.
+     *
+     * @param request
+     * @return an observable that provides the response
+     */
+    Observable<HttpResponse> write(HttpRequest request);
 
     /**
      * Returns if the underlying connection is still active.
--- a/components/client/src/main/java/com/hotels/styx/client/connectionpool/CloseAfterUseConnectionDestination.java
+++ b/components/client/src/main/java/com/hotels/styx/client/connectionpool/CloseAfterUseConnectionDestination.java
@@ -21,7 +21,7 @@ import com.hotels.styx.api.client.ConnectionDestination;
 import com.hotels.styx.api.client.Origin;
 import com.hotels.styx.client.ConnectionSettings;
 import com.hotels.styx.client.HttpRequestOperationFactory;
-import com.hotels.styx.client.netty.HttpRequestOperation;
+import com.hotels.styx.client.netty.connectionpool.HttpRequestOperation;
 import com.hotels.styx.client.netty.connectionpool.NettyConnectionFactory;
 import com.hotels.styx.client.ssl.TlsSettings;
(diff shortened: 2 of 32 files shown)


components/api/src/main/java/com/hotels/styx/api/client/Connection.java
```diff
@@ -68,6 +70,14 @@
      */
     <R> Observable<R> execute(Supplier<Observable<R>> operation);
```
mikkokar: It should be OK to get rid of this now?

components/api/src/main/java/com/hotels/styx/api/client/Connection.java
```diff
@@ -68,6 +70,14 @@
...
 
+    /**
+     * Writes HTTP request to a remote peer in the context of this connection.
+     *
+     * @param request
+     * @return an observable that provides the response
+     */
+    Observable<HttpResponse> write(HttpRequest request);
+
```
mikkokar: The `write` method should also appear in the `Connection` interface. As for the `Connection` interface, it probably needs type parameters in place of HttpRequest and HttpResponse.

alobodzki: But this is a connection interface.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/ProxyToBackend.java
```diff
@@ -79,14 +79,28 @@ public HttpHandler2 build(List<String> parents, BuiltinHandlersFactory builder,
```
mikkokar: on line 76, variable `origins` is no longer used.

components/client/src/main/java/com/hotels/styx/client/connectionpool/CloseAfterUseConnectionDestination.java
```diff
@@ -126,6 +128,13 @@ public void connectionClosed(Connection connection) {
         private Connection.Settings connectionSettings = new ConnectionSettings(1000, 1000);
         private Connection.Factory connectionFactory;
         private TlsSettings tlsSettings;
+        private HttpRequestOperationFactory requestOperationFactory = request -> new HttpRequestOperation(
```
kvosper: Do these fields not need to be configured in the factory?

components/client/src/main/java/com/hotels/styx/client/connectionpool/stubs/StubConnectionFactory.java
```diff
@@ -56,6 +58,11 @@ public StubConnection(Origin origin) {
             return operation.get();
         }
 
+        @Override
+        public Observable<HttpResponse> write(HttpRequest request) {
+            throw new IllegalStateException("Not implemented");
```
kvosper: `java.lang.UnsupportedOperationException` is the correct exception for this scenario.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
```diff
@@ -37,11 +42,12 @@
  * A connection using a netty channel.
  */
 public class NettyConnection implements Connection, TimeToFirstByteListener {
-    public static final AttributeKey<Object> CLOSED_BY_STYX = AttributeKey.newInstance("CLOSED_BY_STYX");
+    private static final AttributeKey<Object> CLOSED_BY_STYX = AttributeKey.newInstance("CLOSED_BY_STYX");
 
     private final Origin origin;
     private final Object id;
     private final Channel channel;
+    private final Function<HttpRequest, HttpRequestOperation> requestOpFactory;
```
kvosper: Please use full words when naming fields.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
```diff
@@ -52,23 +58,14 @@
...
-     * @param origin  the origin connected to
-     * @param channel the netty channel used
-     */
-    public NettyConnection(Object id, Origin origin, Channel channel) {
-        this.id = id;
+    public NettyConnection(Origin origin, Channel channel, HttpRequestOperationFactory factory) {
+        this.id = new UUID();
         this.origin = checkNotNull(origin);
         this.channel = checkNotNull(channel);
+        this.requestOpFactory = factory::newHttpRequestOperation;
```
kvosper: Why not just use `HttpRequestOperationFactory` as the field type, so that this method reference isn't required?

components/api/src/main/java/com/hotels/styx/api/client/Connection.java
mikkokar: IMO it makes sense. Looking at `Connection` interface there is nothing that is inherently specific to HTTP protocol. And it is just our particular implementation of `write` which is a function `HttpRequest -> Observable<HttpResponse>` and this could be expressed in `Connection` in terms of type parameters.

But then again this is not really a top priority at the moment. So let's leave this as it is for now.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
```diff
@@ -37,11 +42,12 @@
  * A connection using a netty channel.
  */
 public class NettyConnection implements Connection, TimeToFirstByteListener {
```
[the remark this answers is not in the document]
mikkokar: *"not at the moment"* is the answer.

mikkokar:
Given that `HttpRequestOperation` is now executed *only* from the `NettyConnection.write`, it should be moved into the same package alongside with `NettyConnection`.

components/client/src/main/java/com/hotels/styx/client/connectionpool/CloseAfterUseConnectionDestination.java
alobodzki: So this line was moved from SimpleNettyHttpClient.Builder, that was doing exactly the same type of initialisation.

components/client/src/main/java/com/hotels/styx/client/connectionpool/stubs/StubConnectionFactory.java
alobodzki: Done.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
alobodzki: Done.

PR #66, opened 2018-01-12
Spi refactor

alobodzki:
Feature branch for several refactoring efforts combined.


mikkokar:
We are going to do this in two parts.

PR #68, opened 2018-01-16
Connection expiration.

alobodzki:
Adding connection expiration logic. Using additional configuration entries user will be able to
suggest to server that it should close and recreate an origin connection periodically.


components/client/src/test/unit/java/com/hotels/styx/client/connectionpool/SimpleConnectionPoolTest.java
```diff
@@ -543,10 +547,54 @@ public void countsTerminatedConnections() throws Exception {
...
+        Connection.Listener listener = mock(Connection.Listener.class);
+
+        Connection connectionExpired = borrowConnectionSynchronously();
+        connectionExpired.addConnectionListener(listener);
+        connectionPool.returnConnection(connectionExpired);
+
+        try {
+            Thread.sleep(SECONDS.toMillis(1));
+        } catch (InterruptedException e) {
+            fail();
```
kvosper: Why not just throw `InterruptedException`?

kvosper: Also, why are we using a delay here? If this behaviour depends on the passage of time, we have a mockable interface called `com.hotels.styx.api.Clock` that can be used for this. (There is a real implementation in `com.hotels.styx.api.Clocks`).

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionExpirationTracker.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Provides wrapper for connection, that stores a connection expiration time. Also provides a method for verification of
+ * a time pasted.
+ */
+class ConnectionExpirationTracker implements ConnectionUsageTracker {
+    private final long connectionExpirationSeconds;
+    private final Supplier<Ticker> tickerSupplier;
+
+    ConnectionExpirationTracker(long connectionExpirationSeconds, Supplier<Ticker> tickerSupplier) {
+        this.connectionExpirationSeconds = connectionExpirationSeconds;
```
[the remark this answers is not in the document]
alobodzki: In this case it can be whatever, non-positive values turn off the functionality.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionExpirationTracker.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    public ExpirableConnection decorate(Connection connection) {
+        return new TrackedConnectionAdapter(connection, connectionExpirationSeconds, tickerSupplier);
+    }
+
+    @Override
+    public boolean shouldTerminate(Connection connection) {
+        return ((ExpirableConnection) connection).isExpired();
```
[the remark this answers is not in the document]
alobodzki: Done

components/client/src/test/unit/java/com/hotels/styx/client/connectionpool/SimpleConnectionPoolTest.java
```diff
@@ -543,10 +547,54 @@ public void countsTerminatedConnections() throws Exception {
...
+        this.connectionPool = connectionPool(
+                connectionPoolConfig()
+                        .maxConnectionsPerHost(1)
+                        .connectionExpirationSeconds(-1, SECONDS));
+        Connection.Listener listener = mock(Connection.Listener.class);
+        Connection listenedConnection = borrowConnectionSynchronously();
+        listenedConnection.addConnectionListener(listener);
+        connectionPool.returnConnection(listenedConnection);
+
+        for (int i = 0; i < 1000000; i++ ) {
```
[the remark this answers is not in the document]
alobodzki: Done.

docs/user-guide/configure-connection-pooling.md
```diff
@@ -28,6 +30,11 @@ The two other parameters, *connectTimeoutMillis* and *socketTimeoutMillis* descr
   * maximum amount of inactivity over the TCP connection before it should be closed
     (*socketTimeoutMillis*).
 
+In case there is a need for periodical termination of connections in the pool, *connectionExpirationSeconds* setting
+allows to specify upper threshold on amount of times the connection can be used before it should
+be closed and recreated. Specifing no-zero value turns off the functionality. This is useful when an origin host is 
+specified as a DNS domain name, and you want to ensure that domain names are re-resolved periodically.
+
 When the connection pool gets full, any additional requests
```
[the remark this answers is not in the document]
mikkokar: "`connectionExpirationSeconds` specifies the maximum lifetime, in seconds, for a pooled connection. After this time the connection is closed. This is useful when an origin host is specified as a DNS domain name, and you need to ensure the domain name is periodically re-resolved. A value of zero turns the feature off."

(Assuming a zero value turns it off? ^^^)

Might be worth mentioning, as an implementation note, that the connections are closed lazily. That is only when a new request arrives.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionUsageTracker.java
```diff
@@ -0,0 +1,64 @@
...
+     * @return decorated connection
+     */
+    Connection decorate(Connection connection);
+
+    /**
+     * Provides logic for checking if connection is usable.
+     * @param connection connection under test.
+     * @return value whether the connection has expired.
+     */
+    boolean shouldTerminate(Connection connection);
```
mikkokar: The `shouldTerminate` feels out of place. Like a leaking abstraction. So if we have a generic `ConnectionUsageTracker` interface behind which it is possible to implement arbitrary connection tracking logic, then why does it need `shouldTerminate` method?

I guess I am trying to say that `shouldTerminate` is very specific to a particular implementation of this interface. Therefore it shouldn't have place in an interface like this. Further evidence for this can be found in `ConnectionExpirationTracker` class where we need to resort to `instanceof` checks:

```
    @Override
    public boolean shouldTerminate(Connection connection) {
        return connection instanceof ExpirableConnection && ((ExpirableConnection) connection).isExpired();
    }
```

Perhaps better way to approach this problem is to perform the check in the `isConnected()` method of the adapting class (`TrackedConnectionAdapter`). Perform the check there and close the underlying channel if necessary. Then we could get away with this method and the associated instance checks.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionExpirationTracker.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public ExpirableConnection decorate(Connection connection) {
+        requireNonNull(connection);
+        return new TrackedConnectionAdapter(connection, connectionExpirationSeconds, tickerSupplier);
+    }
+
+    @Override
+    public boolean shouldTerminate(Connection connection) {
+        return connection instanceof ExpirableConnection && ((ExpirableConnection) connection).isExpired();
+    }
```
mikkokar: We need to remove this method as discussed in my other comments. Also, see below...

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionExpirationTracker.java
```diff
@@ -0,0 +1,102 @@
...
+        @Override
+        public void close() {
+            nettyConnection.close();
+        }
+
+        @Override
+        public boolean isExpired() {
+            return stopwatch.elapsed(SECONDS) >= connectionExpirationSeconds;
+        }
+    }
```
mikkokar: I think this class could perform the expiration check in the `isConnected()` method. Like so:

```
    if (isExpired()) {
        close();
        return true;
    } 
   ...
```

Then we could get rid of the above `shouldTerminate` method, and perhaps we could get rid of `ExpirableConnection` interface too?

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -393,8 +414,14 @@ public Factory recordStats(boolean recordStats) {
             return this;
         }
 
+        public Factory connectionDecorator(ConnectionDecorator connectionDecorator) {
+            this.connectionDecorator = connectionDecorator;
+            return this;
+        }
```
mikkokar: No longer used?

Also my editor says that `recordStats` method above is no longer used. Please review if they can be removed.

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -393,8 +414,14 @@ public Factory recordStats(boolean recordStats) {
...
+
         public SimpleConnectionPool create(Origin origin) {
-            return new SimpleConnectionPool(origin, connectionPoolSettings, connectionFactory, recordStats);
+            return new SimpleConnectionPool(origin, connectionPoolSettings, connectionFactory, recordStats,
+                    connectionDecorator);
         }
     }
 }
```
mikkokar: Now after the latest round of changes, it is apparent that the `SimpleConnectionPool` class doesn't need to change at all. The connections can be decorated by the `ConnectionFactory`. Please consider implementing a new connection factory that delegates connection creation into `NettyConnectionFactory` and decorates the connections with connection expiry behaviour.

PR #77, opened 2018-01-24
Fix for NPE in case of retry.

alobodzki:
In case of retry RetryOnErrorHandler was recreated without required fields.
Additionally this change contains refactoring to reuse OriginStatsFactory in
StyxHttpClient.

### DIFF
--- a/components/client/src/test/unit/java/com/hotels/styx/client/retry/RetryNTimesTest.java
+++ b/components/client/src/test/unit/java/com/hotels/styx/client/retry/RetryNTimesTest.java
@@ -1,0 +1,117 @@
+/**
+ * Copyright (C) 2013-2018 Expedia Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.styx.client.retry;
+
+import com.hotels.styx.api.client.ConnectionPool;
+import com.hotels.styx.api.client.loadbalancing.spi.LoadBalancingStrategy;
+import com.hotels.styx.api.client.retrypolicy.spi.RetryPolicy;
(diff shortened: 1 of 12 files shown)


components/client/src/test/unit/java/com/hotels/styx/client/retry/RetryNTimesTest.java
```diff
@@ -0,0 +1,117 @@
...
+        assertThat(retryOutcome.nextOrigin().get(), equalTo(connectionPool));
+    }
+
+    @Test
+    public void shouldReturnEmptyOriginList() {
+        RetryNTimes retryNTimes = new RetryNTimes(1);
+
+        RetryPolicy.Context context = mock(RetryPolicy.Context.class);
+        ConnectionPool connectionPool = mock(ConnectionPool.class);
+        LoadBalancingStrategy.Context lbContext = mock(LoadBalancingStrategy.Context.class);
```
mikkokar: The mock setup adds lots of verbosity in the tests in this class. Please consider:
1) adding a helper method to set up context
2) convert the mocks like `connectionPool`, `lbContext`, `loadBalancingStrategy` into fields and initialise in test setup method before each test.

components/client/src/test/unit/java/com/hotels/styx/client/retry/RetryNTimesTest.java
```diff
@@ -0,0 +1,109 @@
...
+        this.strategyMock = mock(LoadBalancingStrategy.class);
+        this.strategyContextMock = mock(LoadBalancingStrategy.Context.class);
+        this.connectionPool = mock(ConnectionPool.class);
+
+        when(retryPolicyContext.currentRetryCount()).thenReturn(0);
+        when(retryPolicyContext.lastException()).thenReturn(empty());
+    }
+
+    @Test
+    public void shouldRetry() {
```
kvosper: This test name is not informative. We need to make the cause+effect clear.

components/client/src/test/unit/java/com/hotels/styx/client/retry/RetryNTimesTest.java
```diff
@@ -0,0 +1,109 @@
...
+        when(retryPolicyContext.currentRetryCount()).thenReturn(1);
+        when(retryPolicyContext.lastException()).thenReturn(of(new TestException()));
+        RetryPolicy.Outcome retryOutcome = retryNTimesPolicy.evaluate(retryPolicyContext,
+                strategyMock, strategyContextMock);
+
+        assertThat(retryOutcome.shouldRetry(), equalTo(false));
+    }
+
+    @Test
+    public void shouldNotRetryBasedOnWrongException() {
```
kvosper: What makes an exception wrong? I think the test name should mention the marker interface.

PR #78, opened 2018-01-25
Refactor origin inventory builder

mikkokar:
Simplify builder class for the OriginsInventory.

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/OriginsInventory.java
+++ b/components/client/src/main/java/com/hotels/styx/client/OriginsInventory.java
@@ -22,7 +22,6 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import com.hotels.styx.api.Announcer;
-import com.hotels.styx.api.HttpClient;
 import com.hotels.styx.api.Id;
 import com.hotels.styx.api.client.ActiveOrigins;
 import com.hotels.styx.api.client.ConnectionPool;
@@ -31,14 +30,8 @@ import com.hotels.styx.api.client.OriginsInventorySnapshot;
 import com.hotels.styx.api.client.OriginsInventoryStateChangeListener;
 import com.hotels.styx.api.metrics.MetricRegistry;
 import com.hotels.styx.api.metrics.codahale.CodaHaleMetricRegistry;
-import com.hotels.styx.client.applications.BackendService;
-import com.hotels.styx.client.connectionpool.CloseAfterUseConnectionDestination;
-import com.hotels.styx.client.healthcheck.OriginHealthCheckFunction;
 import com.hotels.styx.client.healthcheck.OriginHealthStatusMonitor;
-import com.hotels.styx.client.healthcheck.OriginHealthStatusMonitorFactory;
-import com.hotels.styx.client.healthcheck.UrlRequestHealthCheck;
 import com.hotels.styx.client.healthcheck.monitors.NoOriginHealthStatusMonitor;
-import com.hotels.styx.client.netty.connectionpool.NettyConnectionFactory;
 import com.hotels.styx.client.origincommands.DisableOrigin;
 import com.hotels.styx.client.origincommands.EnableOrigin;
 import com.hotels.styx.client.origincommands.GetOriginsInventorySnapshot;
@@ -61,6 +54,7 @@ import static com.hotels.styx.client.OriginsInventory.OriginState.DISABLED;
 import static com.hotels.styx.client.OriginsInventory.OriginState.INACTIVE;
 import static com.hotels.styx.common.StyxFutures.await;
 import static java.util.Collections.emptyMap;
+import static java.util.Collections.emptySet;
 import static java.util.Collections.singleton;
 import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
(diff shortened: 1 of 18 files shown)


components/client/src/main/java/com/hotels/styx/client/OriginsInventory.java
```diff
@@ -476,54 +466,33 @@ private OriginState state() {
...
-
         public Builder connectionPoolFactory(ConnectionPool.Factory connectionPoolFactory) {
             this.connectionPoolFactory = connectionPoolFactory;
             return this;
         }
 
-        public Builder connectionFactory(Connection.Factory connectionFactory) {
-            this.connectionFactory = connectionFactory;
+        public Builder originHealthMonitor(OriginHealthStatusMonitor originHealthMonitor) {
+            this.originHealthMonitor = originHealthMonitor;
```
kvosper: We should check this for null here, to avoid throwing an exception in the `build` method.

components/client/src/main/java/com/hotels/styx/client/OriginsInventory.java
```diff
@@ -532,93 +501,28 @@ public Builder eventBus(EventBus eventBus) {
...
-
-        private OriginsInventory originsInventory(OriginHealthStatusMonitor originHealthStatusMonitor, HttpConfig httpConfig,
-                                                  MetricRegistry metricsRegistry, OriginStatsFactory originStatsFactory) {
-            await(originHealthStatusMonitor.start());
-
-            ConnectionPool.Factory hostConnectionPoolFactory = connectionPoolFactory(backendService.connectionPoolConfig(), httpConfig, metricsRegistry, originStatsFactory);
-            OriginsInventory originsInventory = new OriginsInventory(eventBus, backendService.id(), originHealthStatusMonitor, hostConnectionPoolFactory, metricsRegistry);
-            originsInventory.setOrigins(backendService.origins());
+            OriginsInventory originsInventory = new OriginsInventory(eventBus, appId, originHealthMonitor, connectionPoolFactory, metricsRegistry);
+            originsInventory.setOrigins(initialOrigins);
```
kvosper: I think the `setOrigins(Set<Origin>)` method can be private, and `setOrigins(Origin...)` can be package-private.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionPools.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public static ConnectionPool.Factory simplePoolFactory() {
+        return ConnectionPools::poolForOrigin;
+    }
+
+    public static ConnectionPool poolForOrigin(Origin origin, MetricRegistry metricRegistry) {
+        return new StatsReportingConnectionPool(
+                new SimpleConnectionPool(
+                        origin,
+                        defaultSettableConnectionPoolSettings(),
```
kvosper: Should we rename `defaultSettableConnectionPoolSettings` since "settable" is no longer in the class name?

components/proxy/src/main/java/com/hotels/styx/proxy/BackendServicesRouter.java
```diff
@@ -124,6 +155,39 @@ public void onChange(Registry.Changes<BackendService> changes) {
                         .close());
     }
 
+    private OriginHealthCheckFunction originHealthCheckFunction(
```
kvosper: This method can be static.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/ProxyToBackend.java
```diff
@@ -99,14 +102,33 @@ public HttpHandler2 build(List<String> parents, RouteHandlerFactory builder, Rou
...
                     .metricsRegistry(environment.metricRegistry())
-                    .connectionFactory(connectionFactory)
-                    .originStatsFactory(originStatsFactory)
+                    .connectionPoolFactory(connectionPoolFactory)
+                    .initialOrigins(backendService.origins())
                     .build();
             return new ProxyToBackend(clientFactory.createClient(backendService, inventory));
         }
+
+        private ConnectionPoolFactory connectionPoolFactory(
```
kvosper: This method can be static.

components/client/src/main/java/com/hotels/styx/client/OriginsInventory.java
mikkokar: ok

components/client/src/main/java/com/hotels/styx/client/OriginsInventory.java
mikkokar: Hi Kyle. The `setOrigins(Set<Origin>)` is nowadays public. This is because the inventory by design will manage the state for the origins, and the `setOrigins` provides a way to push more origins into the inventory.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionPools.java
mikkokar: Makes sense. Will do.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/ProxyToBackend.java
mikkokar: This method is just a wrapper around `new` keyword. I will remove it.

PR #80, opened 2018-01-26
Separate StyxService from AbstractRegistry.

mikkokar:
Improve modularity of classes by separating StyxService from AbstractRegistry. It is no longer necessary for registries to implement StyxService interface.


components/proxy/src/main/java/com/hotels/styx/infrastructure/AbstractRegistry.java
```diff
@@ -50,10 +50,8 @@
     }
 
     protected void notifyListeners(Changes<T> changes) {
-        if (!changes.isEmpty()) {
```
kvosper: The check has been removed here, but not added to `set(Iterable<T>)`?

PR #84, opened 2018-01-30
Connection expiration

alobodzki:
Adding connection expiration logic. Using additional configuration entries user will be able to
suggest to server that it should close and recreate an origin connection periodically.


components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionFactoryTrackerDecorator.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.function.Supplier;
+
+/**
+ * Decorator for a connection factory that wraps newly created connections into a decorator that tracks and expires the connection
+ * on expiration time.
+ */
+public class ConnectionFactoryTrackerDecorator implements Connection.Factory {
+
+    private static Supplier<Ticker> systemTicker = Ticker::systemTicker;
```
kvosper: Static fields should be final to prevent potential thread-safety issues.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionFactoryTrackerDecorator.java
```diff
@@ -0,0 +1,49 @@
...
+    private final long connectionExpirationSeconds;
+    private Connection.Factory connectionFactory;
```
kvosper: This field should also be final.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionFactoryTrackerDecorator.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public ConnectionFactoryTrackerDecorator(long connectionExpirationSeconds, Connection.Factory connectionFactory) {
+        this.connectionExpirationSeconds = connectionExpirationSeconds;
+        this.connectionFactory = connectionFactory;
```
kvosper: This parameter should be null-checked.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionFactoryTrackerDecorator.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private Observable<Connection> decorate(Observable<Connection> connection) {
```
kvosper: Generally in our codebase we have been putting private methods below the public API methods.

components/client/src/main/java/com/hotels/styx/client/connectionpool/TrackedConnectionDecorator.java
```diff
@@ -0,0 +1,82 @@
...
+ * a time pasted.
+ */
+class TrackedConnectionDecorator implements Connection {
+    private final Connection nettyConnection;
+    private final long connectionExpirationSeconds;
+    private final Stopwatch stopwatch;
+
+    TrackedConnectionDecorator(Connection nettyConnection, long connectionExpirationSeconds, Supplier<Ticker> tickerSupplier) {
+        this.connectionExpirationSeconds = connectionExpirationSeconds;
+        this.nettyConnection = nettyConnection;
```
kvosper: Please check parameters for nulls as this prevents NullPointerExceptions at runtime.

docs/user-guide/configure-connection-pooling.md
```diff
@@ -28,6 +30,11 @@ The two other parameters, *connectTimeoutMillis* and *socketTimeoutMillis* descr
   * maximum amount of inactivity over the TCP connection before it should be closed
     (*socketTimeoutMillis*).
 
+In case there is a need for periodical termination of connections in the pool, *connectionExpirationSeconds* setting
+will create a timer that will be assigned to each connection and initiate a procedure that will close the connection, when it expires.
+This is useful when an origin host is specified as a DNS domain name, and you want to ensure that domain names are re-resolved periodically.
+Specifing non-positive value turns off the functionality. Closing of the connection is performed lazily on incoming request.
+
```
kvosper: A suggestion for a more straightforward way of expressing this:
```
The *connectionExpirationSeconds* setting allows connections to be terminated after a configured period of time has elapsed since the connection was created.
This is useful when an origin host is specified as a DNS domain name, and you want to ensure that domain names are re-resolved periodically.
If the value of the setting is non-positive, connections will not expire. 
Connection age is checked on each incoming request, so connections may live longer than their expiration time if they do not serve any requests.
```

system-tests/e2e-suite/src/test/scala/com/hotels/styx/client/ExpiringConnectionSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+      .build()
+
+    val response1 = waitForResponse(client.sendRequest(request))
+
+    assertThat(response1.status(), is(OK))
+
+    styxServer.metricsSnapshot.gauge(s"origins.appOne.generic-app-01.connectionspool.available-connections").get should be(1)
+    styxServer.metricsSnapshot.gauge(s"origins.appOne.generic-app-01.connectionspool.connections-closed").get should be(0)
+
+    Thread.sleep(1000)
```
kvosper: Perhaps we should allow the expiration time to be set in milliseconds, so that we don't have to add such a large sleep into our integration test? What do you think?

mikkokar: From end-user's perspective the most sensible unit of time for connection expiration is probably a second, or even a minute. Remember that this feature is for mainly re-resolving DNS entries. They are not going to change that often.

I would leave the time unit as it is.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionFactoryTrackerDecorator.java
```diff
@@ -0,0 +1,49 @@
...
+        return connection.map(conn -> new TrackedConnectionDecorator(conn, connectionExpirationSeconds, systemTicker));
+    }
+
+    @Override
+    public Observable<Connection> createConnection(Origin origin, Connection.Settings connectionSettings) {
+        Observable<Connection> connection = connectionFactory.createConnection(origin, connectionSettings);
+        return decorate(connection);
+    }
```
mikkokar: I would personally just inline all these methods:

```
    @Override
    public Observable<Connection> createConnection(Origin origin, Connection.Settings connectionSettings) {
        return connectionFactory
                .createConnection(origin, connectionSettings)
                .map(conn -> new TrackedConnectionDecorator(conn, connectionExpirationSeconds, systemTicker));
    }
```

But I will leave it up to you.

mikkokar: Or actually, you should really use a map operator in the observable to decorate:

```
    private Connection decorate(Connection connection) {
        return new TrackedConnectionDecorator(connection, connectionExpirationSeconds, systemTicker);
    }

    @Override
    public Observable<Connection> createConnection(Origin origin, Connection.Settings connectionSettings) {
        return connectionFactory
                .createConnection(origin, connectionSettings)
                .map(this::decorate);
    }
```

This promotes the idea that decoration is a synchronous operation that acts on the connection object itself rather than on the whole observable.

components/client/src/main/java/com/hotels/styx/client/connectionpool/TrackedConnectionDecorator.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.function.Supplier;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Provides wrapper for connection, that stores a connection expiration time. Also provides a method for verification of
+ * a time pasted.
+ */
+class TrackedConnectionDecorator implements Connection {
```
mikkokar: It feels to me that this decorator name should be more accurate. Why don't we just call it a `TimedConnectionExpiryDecorator`, or `TimedConnectionDecorator`, or something similar.

There is nothing really in this class that "tracks" the connections. It just closes them after awhile.

components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionFactoryTrackerDecorator.java
```diff
@@ -0,0 +1,49 @@
...
+import com.hotels.styx.api.client.Origin;
+import rx.Observable;
+
+import java.util.function.Supplier;
+
+/**
+ * Decorator for a connection factory that wraps newly created connections into a decorator that tracks and expires the connection
+ * on expiration time.
+ */
+public class ConnectionFactoryTrackerDecorator implements Connection.Factory {
```
mikkokar: This class should also be called something like `TimedConnectionFactory` etc.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/client/ExpiringConnectionSpec.scala
alobodzki: I agree with Mikko.

PR #93, opened 2018-02-20
Upgrading to newest metrics version

alobodzki:
Moving to a newest 3.2.6 dropwizard metrics.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
+++ b/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
@@ -208,10 +208,6 @@ public class GraphiteReporter extends ScheduledReporter {
         long timestamp = clock.getTime() / 1000;
 
         try {
-            if (!graphite.isConnected()) {
-                graphite.connect();
-            }
-
             gauges.forEach((name, gauge) ->
                     doReport(name, gauge, timestamp, this::reportGauge));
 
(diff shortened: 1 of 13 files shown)


components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
```diff
@@ -230,12 +226,20 @@ public void report(SortedMap<String, Gauge> gauges,
...
 
     private <M extends Metric> void doReport(String name, M metric, long timestamp, MetricReportingAction<M> consumer) {
         try {
+            if (!graphite.isConnected()) {
+                graphite.connect();
+            }
+
             consumer.execute(name, metric, timestamp);
         } catch (Exception e) {
             LOGGER.error("Error reporting metric '" + name + "': " + e.getMessage(), e);
```
mikkokar: Hi @alobodzki could you provide some background for this change? Mainly for the record please.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/MetricsHandlerTest.java
```diff
@@ -44,6 +44,6 @@ public void respondsToRequestWithJsonResponse() {
     public void exposesRegisteredMetrics() {
         metricRegistry.counter("foo").inc();
         FullHttpResponse response = waitForResponse(handler.handle(get("/metrics").build()));
-        assertThat(response.bodyAs(UTF_8), is("{\"version\":\"3.0.0\",\"gauges\":{},\"counters\":{\"foo\":{\"count\":1}},\"histograms\":{},\"meters\":{},\"timers\":{}}"));
+        assertThat(response.bodyAs(UTF_8), is("{\"version\":\"3.1.3\",\"gauges\":{},\"counters\":{\"foo\":{\"count\":1}},\"histograms\":{},\"meters\":{},\"timers\":{}}"));
```
mikkokar: Just curious as why it is reporting version 3.1.3 when we are upgrading to 3.2.6?

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
alobodzki: Our implementation of GraphiteReport tries to reconnect between failures in a single upload, but the implementation of reconnectIfNecessary reconnects only with IOException, there is no default option in that method, so theoretically we could enter doReport method without connected GraphiteConsole. Dropwizard implementation doesn't have this issue, as they don't try to recover, they just fail fast.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/MetricsHandlerTest.java
alobodzki: I assumed it's JSON scheme version: com.codahale.metrics.json.MetricsModule#VERSION

kvosper: It's definitely not the library version as that one was 3.1.1 before, but the JSON version was 3.0.0

PR #94, opened 2018-02-22
Add power of 2 load balancing strategy.

mikkokar:
Implements power of 2 load balancing strategy to styx. 

There is a know short-come in BusyConnectionStrategy: it can start favouring origins returning large quantities of quick 5xx responses. Sometimes a fault in origin may cause them to fail very early on in the business logic. This result a 5xx error being sent with a relatively low latency compared to successful requests. If such failure becomes persistent, an origin like that appears better compared to the others and therefore will receive more traffic.

A power of 2 strategy lessens this effect. This is because there is a larger element of randomness in its choice. Compare this to BusyConnectionStrategy which always chooses the "best" option.

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategy.java
+++ b/components/client/src/main/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategy.java
@@ -1,0 +1,140 @@
+/**
+ * Copyright (C) 2013-2018 Expedia Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.styx.client.loadbalancing.strategies;
+
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import com.hotels.styx.api.Environment;
+import com.hotels.styx.api.client.ActiveOrigins;
+import com.hotels.styx.api.client.ConnectionPool;
+import com.hotels.styx.api.client.RemoteHost;
+import com.hotels.styx.api.client.loadbalancing.spi.LoadBalancingStrategy;
+import com.hotels.styx.api.client.loadbalancing.spi.LoadBalancingStrategyFactory;
+import com.hotels.styx.api.configuration.Configuration;
+
+import java.util.List;
+import java.util.Random;
+
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.StreamSupport.stream;
+
+
+/**
+ * A Power Of Two load balancing strategy.
+ *
+ * Randomly choose two origins, and return the one with less ongoing requests.
+ */
+public class PowerOfTwoStrategy implements LoadBalancingStrategy {
+    private ActiveOrigins activeOrigins;
+    private Random rng;
+
+    public PowerOfTwoStrategy(ActiveOrigins activeOrigins) {
+        this(activeOrigins, new Random());
+    }
+
+    @VisibleForTesting
+    PowerOfTwoStrategy(ActiveOrigins activeOrigins, Random rng) {
+        this.activeOrigins = activeOrigins;
+        this.rng = rng;
+    }
+
+    /**
+     * A load balancing strategy that favours the origin with the least response time.
+     */
+    public static class Factory implements LoadBalancingStrategyFactory {
+        @Override
+        public LoadBalancingStrategy create(Environment environment, Configuration strategyConfiguration, ActiveOrigins activeOrigins) {
+            return new PowerOfTwoStrategy(activeOrigins);
+        }
+    }
+
+    @Override
+    public Iterable<RemoteHost> vote(Context context) {
+        List<ConnectionPoolStatus> poolsList = stream(activeOrigins.snapshot().spliterator(), false)
+                .map(host -> new ConnectionPoolStatus(host, context))
+                .collect(toList());
+
+        if (poolsList.size() == 0) {
+            return ImmutableList.of();
+        }
+
+        if (poolsList.size() == 1) {
+            return ImmutableList.of(poolsList.get(0).host());
+        }
+
+        ConnectionPoolStatus poolStatus1 = choose(poolsList);
+        ConnectionPoolStatus poolStatus2 = choose(poolsList, poolStatus1);
+
+        return betterOf(poolStatus1, poolStatus2);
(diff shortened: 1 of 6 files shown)


components/client/src/main/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategy.java
```diff
@@ -0,0 +1,140 @@
...
+import static java.util.stream.StreamSupport.stream;
+
+
+/**
+ * A Power Of Two load balancing strategy.
+ *
+ * Randomly choose two origins, and return the one with less ongoing requests.
+ */
+public class PowerOfTwoStrategy implements LoadBalancingStrategy {
+    private ActiveOrigins activeOrigins;
```
kvosper: These fields can be final.

components/client/src/main/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategy.java
```diff
@@ -0,0 +1,140 @@
...
+    private Random rng;
+
+    public PowerOfTwoStrategy(ActiveOrigins activeOrigins) {
+        this(activeOrigins, new Random());
+    }
+
+    @VisibleForTesting
+    PowerOfTwoStrategy(ActiveOrigins activeOrigins, Random rng) {
+        this.activeOrigins = activeOrigins;
```
kvosper: These parameters should be null-checked. Please add these checks *before* submitting a pull request.

components/client/src/main/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategy.java
```diff
@@ -0,0 +1,140 @@
...
+        int index = rng.nextInt(poolsList.size());
+        return poolsList.get(index);
+    }
+
+    private static class ConnectionPoolStatus {
+        private final RemoteHost host;
+        private final int busyConnectionCount;
+        private final int pendingConnectionCount;
+
+        public ConnectionPoolStatus(RemoteHost host, Context context) {
```
kvosper: Context is not used.

components/client/src/main/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategy.java
```diff
@@ -0,0 +1,140 @@
...
+        } else {
+            return ImmutableList.of(poolStatus2.host(), poolStatus1.host());
+        }
+    }
+
+    private ConnectionPoolStatus choose(List<ConnectionPoolStatus> poolsList, ConnectionPoolStatus another) {
+        int i = 0;
+        ConnectionPoolStatus second = choose(poolsList);
+
+        while (i < 20 && another.host().id() == second.host.id()) {
```
kvosper: This seems a little arbitrary. Why iterate 20 times? Do we need this trial and error approach? It seems like there is no need for this repeating trial-and-error approach, we simply need to exclude the "another" ID when we `choose`.

Some pseudo-code:
```
int indexOfOther = find(pool in pools)
int randomIndex = random(pools.size)
if(randomIndex >= indexOfOther) {
  randomIndex++
}
```

components/client/src/test/unit/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategyTest.java
```diff
@@ -0,0 +1,121 @@
...
+import static com.hotels.styx.api.support.HostAndPorts.localHostAndFreePort;
+import static java.util.Arrays.asList;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+
+public class PowerOfTwoStrategyTest {
+    final Origin ORIGIN_ONE = newOriginBuilder(localHostAndFreePort()).id("one").build();
```
kvosper: 1. These field names are capitalised, which indicates static fields in Java, but these fields are not static.
2. We should have consistent visibility between our constants. Here, some are private, some are default.

components/client/src/test/unit/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategyTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private long RNG_SEED = 5;
+
+
+    @Test
+    public void choosesBetterOfTwoRandomChoices() {
+        ActiveOrigins activeOrigins = mock(ActiveOrigins.class);
+        when(activeOrigins.snapshot()).thenReturn(allOrigins);
+
+        Random rng = new Random(RNG_SEED);
```
kvosper: Perhaps the output of the RNG should be proven in the `@BeforeMethod` setUp?

components/client/src/test/unit/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategyTest.java
```diff
@@ -0,0 +1,121 @@
...
+        int first = rng.nextInt(4);
+        assert(first == 2);
```
kvosper: Why are these not using hamcrest matchers? We should be consistent.

components/client/src/test/unit/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategyTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+        int second = rng.nextInt(4);
+        assert(second == 0);
+
+        PowerOfTwoStrategy loadBalancer = new PowerOfTwoStrategy(activeOrigins, new Random(RNG_SEED));
+        Iterable<RemoteHost> chosenOne = loadBalancer.vote(mock(LoadBalancingStrategy.Context.class));
+
+        assertThat(getFirst(chosenOne, null), is(betterOf(allOrigins.get(first), allOrigins.get(second))));
```
kvosper: It would be clearer to have the expected result stated explicitly, rather than generated by a function.
We could add a brief (one line) comment to explain why it is the expected result if necessary.

components/client/src/main/java/com/hotels/styx/client/loadbalancing/strategies/PowerOfTwoStrategy.java
```diff
@@ -0,0 +1,140 @@
...
+            return ImmutableList.of(poolsList.get(0).host());
+        }
+
+        ConnectionPoolStatus poolStatus1 = choose(poolsList);
+        ConnectionPoolStatus poolStatus2 = choose(poolsList, poolStatus1);
+
+        return betterOf(poolStatus1, poolStatus2);
+    }
+
+    private Iterable<RemoteHost> betterOf(ConnectionPoolStatus poolStatus1, ConnectionPoolStatus poolStatus2) {
```
kvosper: The function to decide which pool is better could be decoupled and passed in - since the concept of choosing the better of two randoms is independent of what you consider "better" to mean.

mikkokar:
This work is contained within #96.

PR #97, opened 2018-03-01
Origins file monitor

mikkokar:
Add an option to monitor changes to the origin configuration file, and to automatically reload the file when it changes.

### DIFF
--- a/components/common/src/main/java/com/hotels/styx/common/Files.java
+++ b/components/common/src/main/java/com/hotels/styx/common/Files.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright (C) 2013-2018 Expedia Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.styx.common;
+
+import com.google.common.hash.HashCode;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.file.Path;
+
+import static com.google.common.hash.Hashing.md5;
+import static com.google.common.io.ByteStreams.toByteArray;
+
+public final class Files {
+    private Files() {
+    }
+
+    public static HashCode fileContentMd5(Path path) {
+        try {
+           return md5().hashBytes(toByteArray(new FileInputStream(path.toFile())));
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+}
--- a/components/common/src/main/java/com/hotels/styx/common/Files.java
+++ b/components/common/src/main/java/com/hotels/styx/common/Files.java
@@ -19,6 +19,7 @@ import com.google.common.hash.HashCode;
 
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Path;
 
 import static com.google.common.hash.Hashing.md5;
@@ -29,8 +30,8 @@ public final class Files {
     }
 
     public static HashCode fileContentMd5(Path path) {
-        try {
-           return md5().hashBytes(toByteArray(new FileInputStream(path.toFile())));
+        try (InputStream stream = new FileInputStream(path.toFile())) {
+           return md5().hashBytes(toByteArray(stream));
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
--- a/components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
+++ b/components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
@@ -108,6 +108,7 @@ public class FileBackedBackendServicesRegistry extends AbstractStyxService imple
         return super.stop();
     }
 
+    @VisibleForTesting
     FileMonitor monitor() {
         return fileChangeMonitor;
     }
@@ -142,11 +143,8 @@ public class FileBackedBackendServicesRegistry extends AbstractStyxService imple
                     newResource(originsFile),
                     new YAMLBackendServicesReader());
 
-            if (monitorSettings.enabled()) {
-                return new FileBackedBackendServicesRegistry(fileBackedRegistry, new FileChangeMonitor(originsFile));
-            } else {
-                return new FileBackedBackendServicesRegistry(fileBackedRegistry, FileMonitor.DISABLED);
-            }
+            FileMonitor monitor = monitorSettings.enabled() ? new FileChangeMonitor(originsFile) : FileMonitor.DISABLED;
+            return new FileBackedBackendServicesRegistry(fileBackedRegistry, monitor);
         }
 
         private static String requireNonEmpty(String originsFile) {
--- a/components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
+++ b/components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
@@ -26,6 +26,7 @@ import com.hotels.styx.client.applications.BackendServices;
 import com.hotels.styx.infrastructure.FileBackedRegistry;
 import com.hotels.styx.infrastructure.Registry;
 import com.hotels.styx.infrastructure.YamlReader;
+import com.hotels.styx.proxy.backends.file.FileChangeMonitor.FileMonitorSettings;
 
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
@@ -40,21 +41,31 @@ import static java.util.Objects.requireNonNull;
 /**
  * File backed {@link com.hotels.styx.client.applications.BackendService} registry.
  */
-public class FileBackedBackendServicesRegistry extends AbstractStyxService implements Registry<BackendService> {
+public class FileBackedBackendServicesRegistry extends AbstractStyxService implements Registry<BackendService>, FileChangeMonitor.Listener {
     private final FileBackedRegistry<BackendService> fileBackedRegistry;
+    private final FileMonitor fileChangeMonitor;
 
     @VisibleForTesting
     FileBackedBackendServicesRegistry(FileBackedRegistry<BackendService> fileBackedRegistry) {
-        super(format("FileBackedBackendServiceRegistry(%s)", fileBackedRegistry.fileName()));
-        this.fileBackedRegistry = requireNonNull(fileBackedRegistry);
+        this(fileBackedRegistry, FileMonitor.DISABLED);
     }
 
+    @VisibleForTesting
+    FileBackedBackendServicesRegistry(FileBackedRegistry<BackendService> fileBackedRegistry, FileMonitor fileChangeMonitor) {
+        super(format("FileBackedBackendServiceRegistry(%s)", fileBackedRegistry.fileName()));
+        this.fileBackedRegistry = requireNonNull(fileBackedRegistry);
+        this.fileChangeMonitor = requireNonNull(fileChangeMonitor);
+    }
 
     public static FileBackedBackendServicesRegistry create(String originsFile) {
+        return create(originsFile, FileMonitor.DISABLED);
+    }
+
+    static FileBackedBackendServicesRegistry create(String originsFile, FileMonitor fileChangeMonitor) {
         FileBackedRegistry<BackendService> fileBackedRegistry = new FileBackedRegistry<>(
                 newResource(originsFile),
                 new YAMLBackendServicesReader());
-        return new FileBackedBackendServicesRegistry(fileBackedRegistry);
+        return new FileBackedBackendServicesRegistry(fileBackedRegistry, fileChangeMonitor);
     }
 
     @Override
@@ -79,6 +90,13 @@ public class FileBackedBackendServicesRegistry extends AbstractStyxService imple
 
     @Override
     protected CompletableFuture<Void> startService() {
+        try {
+            fileChangeMonitor.start(this);
+        } catch (Exception e) {
+            CompletableFuture<Void> x = new CompletableFuture<>();
+            x.completeExceptionally(e);
+            return x;
+        }
         return this.fileBackedRegistry.reload()
                 .thenAccept(result -> {
                     // Swallow the result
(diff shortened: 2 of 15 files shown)


components/common/src/main/java/com/hotels/styx/common/Files.java
```diff
@@ -0,0 +1,38 @@
...
+import static com.google.common.hash.Hashing.md5;
+import static com.google.common.io.ByteStreams.toByteArray;
+
+public final class Files {
+    private Files() {
+    }
+
+    public static HashCode fileContentMd5(Path path) {
+        try {
+           return md5().hashBytes(toByteArray(new FileInputStream(path.toFile())));
```
dvlato: We are not closing the FileInputStream here.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+                LOGGER.debug("Monitored file does not exist. Path={}", monitoredFile);
+                return;
+            }
+
+            if (!isReadable(monitoredFile)) {
+                LOGGER.debug("Monitored file is no longer readable. Path={}", monitoredFile);
+                return;
+            }
+
+            if (modificationTimeChanged(monitoredFile)) {
```
dvlato: We might want to refactor this into a new method " fileHasChanged" . 

Potentially, we might want to include a "file size" check just after the timestamp; that way, we could avoid having to perform the md5sum over the whole content when the file size has changed.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
```diff
@@ -90,27 +108,45 @@ public static FileBackedBackendServicesRegistry create(String originsFile) {
         return super.stop();
     }
 
+    FileMonitor monitor() {
```
kvosper: Is this package-private because it is for a test? Let's use the `@VisibleForTesting` annotation as we have done in other classes.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
```diff
@@ -90,27 +108,45 @@ public static FileBackedBackendServicesRegistry create(String originsFile) {
...
+        private static Registry<BackendService> registry(String originsFile, FileMonitorSettings monitorSettings) {
             requireNonEmpty(originsFile);
 
             FileBackedRegistry<BackendService> fileBackedRegistry = new FileBackedRegistry<>(
                     newResource(originsFile),
                     new YAMLBackendServicesReader());
 
-            return new FileBackedBackendServicesRegistry(fileBackedRegistry);
+            if (monitorSettings.enabled()) {
+                return new FileBackedBackendServicesRegistry(fileBackedRegistry, new FileChangeMonitor(originsFile));
```
kvosper: Only the file monitor varies. We shouldn't duplicate the rest of the code. We could do
```
FileMonitor monitor = monitorSettings.enabled() ? new FileChangeMonitor(originsFile) : FileMonitor.DISABLED;
return new FileBackedBackendServicesRegistry(fileBackedRegistry, monitor);

File: components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java [hunk 5]
```diff
@@ -0,0 +1,155 @@
...
+ * Monitors a file system object and notifies the consumer of any changes.
+ */
+public class FileChangeMonitor implements FileMonitor {
+    private final Path monitoredFile;
+    private final ScheduledExecutorService executor = newSingleThreadScheduledExecutor();
+    private final long pollPeriod;
+    private final TimeUnit timeUnit;
+    private ScheduledFuture<?> monitoredTask;
+
+    private AtomicReference<FileTime> lastChangedTime = new AtomicReference<>(FileTime.fromMillis(0));
```
kvosper: These `AtomicReference` fields can be final.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+    private AtomicReference<HashCode> hashCode = new AtomicReference<>();
+    private volatile boolean performHashCheck;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FileChangeMonitor.class);
+
+    @VisibleForTesting
+    FileChangeMonitor(String monitoredFile, long pollPeriod, TimeUnit timeUnit) {
+        requireExists(requireNonNull(monitoredFile));
+        this.monitoredFile = Paths.get(monitoredFile);
+        this.pollPeriod = pollPeriod;
```
kvosper: We can encapsulate some of these fields into our `Schedule` class if we move it out of `com.hotels.styx.client.healthcheck`, maybe into `common`? Just an idea.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+        synchronized (this) {
+            if (monitoredTask != null) {
+                throw new IllegalStateException(format("File monitor for '{}' is already started", monitoredFile));
+            }
+
+            monitoredTask = executor.scheduleAtFixedRate(newPollerTask(listener), pollPeriod, pollPeriod, timeUnit);
+        }
+    }
+
+    private Runnable newPollerTask(Listener listener) {
```
kvosper: I think it would be better if we named our method after what we wanted the task to do on each call, rather than simply naming it a "poller task".

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+        return () -> {
+            if (!exists(monitoredFile)) {
+                LOGGER.debug("Monitored file does not exist. Path={}", monitoredFile);
+                return;
```
kvosper: Could we not just use `else`?

```
if(condition) {
  // code

} else if(otherCondition) {
  // more code

} else ...

File: components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java [hunk 9]
```diff
@@ -0,0 +1,155 @@
...
+
+    private boolean modificationTimeChanged(Path monitoredFile) {
+        try {
+            FileTime current = getLastModifiedTime(monitoredFile);
+            FileTime previous = lastChangedTime.getAndSet(current);
+            boolean changed = !previous.equals(current);
+            LOGGER.debug("modificationTimeChange probe. Changed={}", changed);
+            return changed;
+        } catch (IOException e) {
+            LOGGER.debug("modificationTimeChange probe failed. Cause={}", e);
```
kvosper: Why not add this information into the exception message?

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public FileChangeMonitor(String monitoredFile) {
+        this(monitoredFile, 1, SECONDS);
+    }
+
+    @Override
+    public void start(Listener listener) {
+        synchronized (this) {
+            if (monitoredTask != null) {
+                throw new IllegalStateException(format("File monitor for '{}' is already started", monitoredFile));
```
kvosper: Incorrect syntax. It should be `%s%` not `{}`.

components/proxy/src/test/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistryFactoryTest.java
```diff
@@ -72,9 +105,34 @@ public void requiresOriginsFileToBeNonEmpty() {
         new FileBackedBackendServicesRegistry.Factory().create(environment, configuration);
     }
 
-    private Configuration mockConfiguration(Optional<String> s) {
+    @Test
+    public void fileMonitorIsTurnedOffByDefault() {
+        Environment environment = new com.hotels.styx.Environment.Builder().build();
```
kvosper: If `environment` is not test-specific, it can go in the set-up, to minimise clutter in the tests themselves.

components/proxy/src/test/java/com/hotels/styx/proxy/backends/file/FileChangeMonitorTest.java
```diff
@@ -0,0 +1,149 @@
...
+        verify(listener, timeout(3000).times(1)).fileChanged();
+    }
+
+
+    @Test
+    public void notifiesListenersOnFileChange() throws Exception {
+        monitor.start(listener);
+        verify(listener, timeout(3000).times(1)).fileChanged();
+
+        Thread.sleep(1000);
```
kvosper: We are adding a lot of sleeps in these tests, which all add up to make the total build time longer. Is there no way to avoid it?

system-tests/e2e-suite/src/test/scala/com/hotels/styx/admin/FileBasedOriginsFileChangeMonitorSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+  val reqToApp01 = get(s"http://localhost:${styxServer.proxyHttpAddress().getPort}/app01/x").build()
+  val reqToApp02 = get(s"http://localhost:${styxServer.proxyHttpAddress().getPort}/app02/x").build()
+
+
+  override protected def beforeAll(configMap: ConfigMap) {
+    styxServer.isRunning should be(true)
+  }
+
+  it("Responds with BAD_REQUEST when the origins cannot be read") {
+    waitForResponse(client.sendRequest(reqToApp01.toStreamingRequest)).status() should be (HttpResponseStatus.OK)
```
kvosper: We should import these status codes so we don't need to include the class name.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
mikkokar: In principle I think it would be a good idea. However in this instance I would caution against it. For the following reasons:

1) The hash code is computed only once on the next poll cycle, regardless of the file size change. This is to ensure we always pick up the latest changes if the file changes incrementally.

2) It adds little bit more code which is really quite tricky to test.

For these reasons I would prefer for the simpler option and to keep the length it out.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
mikkokar: Hmm... we can think about this. I'll make a mental note of it.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
mikkokar: `detectFileChangesTask`?

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
mikkokar: Matter of taste. IMO the function would look very crowded with  if/else branching.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
kvosper: Sounds good to me.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
mikkokar: Hmm... perhaps you are correct.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileChangeMonitor.java
kvosper: Well here's a comparison:

**Before:**
```
            if (!exists(monitoredFile)) {
                LOGGER.debug("Monitored file does not exist. Path={}", monitoredFile);
                return;
            }

            if (!isReadable(monitoredFile)) {
                LOGGER.debug("Monitored file is no longer readable. Path={}", monitoredFile);
                return;
            }

            if (modificationTimeChanged(monitoredFile)) {
                hashCode.set(fileContentMd5(monitoredFile));
                performHashCheck = true;

                listener.fileChanged();
                return;
            }

            if (performHashCheck && contentHashChanged(monitoredFile)) {
                listener.fileChanged();
            }
```
**After:**
```
            if (!exists(monitoredFile)) {
                LOGGER.debug("Monitored file does not exist. Path={}", monitoredFile);

            } else if (!isReadable(monitoredFile)) {
                LOGGER.debug("Monitored file is no longer readable. Path={}", monitoredFile);

            } else if (modificationTimeChanged(monitoredFile)) {
                hashCode.set(fileContentMd5(monitoredFile));
                performHashCheck = true;

                listener.fileChanged();

            } else if (performHashCheck && contentHashChanged(monitoredFile)) {
                listener.fileChanged();
            }
```

I don't think there's a huge difference, but the latter removes some unnecessary lines/statements.

components/proxy/src/test/java/com/hotels/styx/proxy/backends/file/FileChangeMonitorTest.java
mikkokar: I have reduced sleeping. I couldn't think of any way of completely avoiding sleeps without more extensive changes. At the moment we are adding about 2.25 seconds to the overall build time.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/admin/FileBasedOriginsFileChangeMonitorSpec.scala
mikkokar: Changed the test case name.

components/proxy/src/test/java/com/hotels/styx/proxy/backends/file/FileChangeMonitorTest.java
kvosper: Yeah, I have a feeling in this particular case we may not be able to easily avoid it, but the reason we try to is that when we have lots of tests adding just a couple of seconds (added gradually over months/years), those seconds add up to minutes.
I think we'll just have to accept it for this class though.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
mikkokar: That is correct. Will fix it.

components/proxy/src/test/java/com/hotels/styx/proxy/backends/file/FileChangeMonitorTest.java
dvlato: So, would it be acceptable in this case to simply change the modification time  manually (using Files.setLasModifiedTime() or File.setLastModified() )  and thus avoid having to sleep at all?

PR #101, opened 2018-03-02
Styx tests should pass in windows

dvlato:
Fixed several issues when trying to run the build in Windows.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/support/PathTrie.java
+++ b/components/api/src/main/java/com/hotels/styx/api/support/PathTrie.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (C) 2013-2018 Expedia Inc.
- * <p>
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * <p>
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/components/api/src/main/java/com/hotels/styx/api/support/PathTrie.java
+++ b/components/api/src/main/java/com/hotels/styx/api/support/PathTrie.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (C) 2013-2018 Expedia Inc.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -19,7 +19,13 @@ import org.slf4j.Logger;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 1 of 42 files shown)


components/api/src/test/java/com/hotels/styx/api/io/FileResourceIndexTest.java
```diff
@@ -27,14 +27,15 @@
...
+    private Path PLUGINS_FIXTURE_PATH = ResourcePaths.fixturesHome(FileResourceIndexTest.class, "/plugins");
+    private FileResourceIndex resourceIndex = new FileResourceIndex();
+    private String EXISTING_PLUGIN_RELPATH = "oneplugin/url-rewrite-1.0-SNAPSHOT.jar";
 
     @Test
     public void listsResourcesFromFileSystemDirectory() {
         Iterable<Resource> jars = resourceIndex.list(PLUGINS_FIXTURE_PATH.toString(), ".jar");
 
-        assertThat(jars, contains(resourceWithPath("oneplugin/url-rewrite-1.0-SNAPSHOT.jar")));
+        assertThat(jars, contains(resourceWithPath(EXISTING_PLUGIN_RELPATH)));
```
kvosper: We follow a principle of leaving literal strings in tests where possible so that readers do not need to jump around the file to understand the test.

components/api/src/test/java/com/hotels/styx/api/io/ResourceFactoryTest.java
```diff
@@ -26,7 +26,7 @@
     public void canAcquireClasspathResources() {
         Resource resource = ResourceFactory.newResource("classpath:com/hotels/styx/api/io/resource.txt");
         
-        assertThat(resource, contains("This is an example resource.\nIt has content to use in automated tests."));
+        assertThat(resource, contains("This is an example resource."+System.lineSeparator()+"It has content to use in automated tests."));
```
kvosper: We may want to consider using `String.format` (with static imports) for these line separator cases.

e.g. ` assertThat(resource, contains(format("This is an example resource.%nIt has content to use in automated tests.")));`

components/proxy/src/test/java/com/hotels/styx/StartupConfigTest.java
```diff
@@ -19,14 +19,12 @@
...
+import java.io.File;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
-import static com.hotels.styx.StartupConfig.CONFIG_FILE_LOCATION_VAR_NAME;
-import static com.hotels.styx.StartupConfig.LOGBACK_CONFIG_LOCATION_VAR_NAME;
-import static com.hotels.styx.StartupConfig.STYX_HOME_VAR_NAME;
-import static com.hotels.styx.StartupConfig.newStartupConfigBuilder;
+import static com.hotels.styx.StartupConfig.*;
```
kvosper: We don't use the `*` import in Styx as we like to only import exactly what we are using.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/json/JsonReformatterTest.java
```diff
@@ -107,7 +107,8 @@ public void createsMetricsJsonCorrectly() throws JsonProcessingException {
                 "\n" +
                 "  },\n" +
                 "  \"version\":\"3.1.3\"\n" +
-                "}"));
+                "}";
+        assertThat(after, is(expectedString.replace("\n", System.lineSeparator())));
```
kvosper: We should try to find a consistent way to handle the line-separator cross-platform issue, so that we are not cluttering the code too much.

components/proxy/src/test/java/com/hotels/styx/StartupConfigTest.java
dvlato: As discussed, that's due to IntelliJ settings. The IDE itself was changing the imports as per the "count to use (static) import with *"  If you could do an export of the Settings -> Editor -> Code Style to verify that I don't have any other settings configured that break  codestyle's expectations, that would be great.

components/api/src/main/java/com/hotels/styx/api/support/PathTrie.java
```diff
@@ -1,12 +1,12 @@
 /**
- * Copyright (C) 2013-2017 Expedia Inc.
- *
+ * Copyright (C) 2013-2018 Expedia Inc.
+ * <p>
```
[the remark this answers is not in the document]
dvlato: I have addressed these issues with the IDE in another comment. Other than that,  do you have any document with coding guidelines you follow? For instance, for the "use string literals" in tests part.

components/api/src/main/java/com/hotels/styx/api/support/PathTrie.java
```diff
@@ -249,10 +250,10 @@ private String lastName(List<String> components) {
         return components;
     }
 
-    private Path removeAsterisk(Path path) {
-        Path newPath = path;
-        if (path.endsWith(Paths.get("*"))) {
-            newPath = path.getParent();
+    private String removeAsterisk(String path) {
+        String newPath = path;
+        if (path.endsWith("*")) {
```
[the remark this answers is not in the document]
dvlato: So can the path have "*" anywhere other than after a slash?  There are no tests for that condition in PathTrieTest.  In Windows * is not a valid character for filenames, so I understood it was safe to remove it in any case. If it is not, I think we might want to make code more explicit regarding what Paths are valid.

PR #106, opened 2018-03-08
Document origin configuration and improve user guide.

dvlato:
Changes to the user guide:

 -Included a section regarding the origin configuration.
 -Status codes extracted to their own page.
 -Rest of sections reviewed.


docs/user-guide/basic-usage.md
```diff
@@ -2,34 +2,30 @@
 
 ## System Requirements
 
-Building Styx requires Java 1.8. It can be built with 1.8.0_45. Earlier maintenance
-releases may work, but are not guaranteed to.
-
-The build system requires Apache Maven. The Styx team uses Maven version 3.2.1
-for automated continuous integration builds. On Mac OSX, a version installed 
-by HomeBrew is satisfactory.
+Running Styx requires Java 1.8. It can be run with 1.8.0_45 or later versions. Earlier maintenance
```
mikkokar: Why is the remark about maven being removed?

docs/user-guide/basic-usage.md
```diff
@@ -90,7 +86,7 @@ configuration.
 By default, the logback file is loaded from *$STYX_HOME/conf/logback.xml*. You can specify an alternative
 logging configuration file using {-l} or {--logback} command line options:
```
mikkokar: There are some old formatting isssues with the page, such as `-l` and `—logback` above. Please address them too now as you are changing this file. Need to have a consistent formatting for this and the other pages.

docs/user-guide/configure-load-balancing.md
```diff
@@ -9,25 +9,26 @@ Styx supports three load balancing strategies:
 Styx also provides a mechanism to bypass the load balancer and force
 the origin at source.
 
+##Load balancing strategies
```
mikkokar: This heading is not correctly rendered.

docs/user-guide/configure-origins.md
```diff
@@ -0,0 +1,166 @@
...
+The value for this **originsFile** property accepts including environment variables (with the ${ENV_VAR} format) and can reference:
+ - A path in the filesystem by default.
+ - A resource in the classpath by using the prefix *classpath:*.
+
+###Examples:
+```yaml
+    originsFile=${STYX_HOME}/conf/env-development/origins.yaml
+    originsFile=classpath:/conf/origins.yaml
+ ```
+
```
mikkokar: This file is a great overall addition to our documentation. Thanks for putting it together.

However the above description about *originsFile* is slightly incomplete.

The origins file is actually specified in the `services/` section, under `backendServiceRegistry`, see: https://github.com/dvlato/styx/blob/5e18c02bdcf7e39e462f5c1232fe851fc3c1ed67/distribution/conf/default.yml#L59.

Note that this behaviour also depends on the advanced routing settings.

 The reason the above works is because of Styx configuration overrides.

docs/user-guide/configure-overview.md
```diff
@@ -1,14 +1,12 @@
-## Configuring Styx environment
-
-Styx supports a YAML-based configuration. A yaml file can be specified using a system property `YAML_CONFIG_LOCATION=your/file/path`.
-By default a file is taken from the classpath at `classpath:conf/environment/styx-config.yaml`.
-Regardless of file location, these properties can all be overwritten using system properties by the same name.
+## Configuring Styx
 
+Styx supports a YAML file-based configuration. The YAML file can be specified using the system property `YAML_CONFIG_LOCATION=your/file/path`.
+By default, the file is read from the classpath at `classpath:conf/environment/styx-config.yaml`.
+Additionally, all these properties can be overwritten using environment variables with the same name as the property.
```
mikkokar: This is badly outdated information. The preferred (and only?) way is to pass the configuration file as a command line argument to the startup script. Please update accordingly.

docs/user-guide/configure-tls.md
```diff
@@ -1,20 +1,19 @@
...
-Transport Layer Security (TLS) can be enabled on Styx network
-interfaces. You can enable it on the proxy server port, and also for
-each of the backend services, where necessary.
+Transport Layer Security (TLS) can be enabled on Styx both for the incoming interface
+(connections from clients) and the connections with the backend services.
 
-TLS can be enabled independently for each network interface. You can
-enable it independently for proxy server port, and for each of the
-back-end services. Styx will convert between the protocols where
+TLS can be enabled independently for each network resource. You can
```
mikkokar: The usage of term “network resource” is rether confusing. What we are trying to say is that the Styx server port and each app can have their TLS settings configured separately.

docs/user-guide/configure-tls.md
```diff
@@ -108,50 +107,54 @@ be specified as separate backends.
 
 ### Backend Application Attributes:
 
-  - trustAllCerts - Whether (true) or not (false) to authenticate server side.
+
+  - trustAllCerts - Turns off (true) or on (true) client-side TLS authentication. When `trustAllCerts` is false,
```
mikkokar: “Client-side TLS authentication” is potentially confusing. It can lead readers to think about TLS client authentication which is a completely different matter. Please paraphrase and remove “client side” etc from the description.

docs/user-guide/configure-tls.md
```diff
@@ -108,50 +107,54 @@ be specified as separate backends.
...
 
-      - *alias* - A symbolic name for the certificate. Used as an alias for keystore.
+      - *alias* - A symbolic name for the certificate. Used as an alias in the keystore.
 
       - *path*  - A path to the certificate file.
 
-  - *trustStorePath* - Styx can load the relevant secure material from Java truststore.
-    This attribute specifies a path to the keystore file containing relevant trust material.
+  - *trustStorePath* - Styx can load the list of trusted certificates from a Java JKS keystore.
+    This attribute specifies a path to the truststore file containing the trusted certificates.
```
mikkokar: By convention this file where trust material is stored is called *truststore* which technically is just a java keystore file. In this regard the description feels bit more confusing than it was before.

docs/user-guide/basic-usage.md
dvlato: The rest of the instructions were considering that we were downloading an already compiled binary, but here we were explaining how to build the app instead of how to run it.

Probably we can also add instructions to  build from source code.

docs/user-guide/configure-load-balancing.md
dvlato: Oh thanks! It's rendering  fine inside IntelliJ, let me see if I can fix that...

docs/user-guide/configure-overview.md
dvlato: Actually, specifying the configuration file with "CONFIG_FILE" (as the quickstart explains) also work. I hadn't noticed it wasn't the same property. 

Should I remove the reference to "CONFIG_FILE" from both documents?

docs/user-guide/configure-tls.md
dvlato: Ok.
The document used to say "for each network interface" which, to me, would mean that you can bind to different network cards / ip addresses.  I don't see any part of the code that allows for this.  If it's some feature we have lost or I've missed it, please tell me so.

docs/user-guide/configure-overview.md
kvosper: As far as I can tell `YAML_CONFIG_LOCATION` is not used by Styx at all, so it should not appear in the documentation.

docs/user-guide/basic-usage.md
mikkokar: Ok. Makes sense.

docs/user-guide/configure-load-balancing.md
mikkokar: Hi. This is still not rendering correctly. Please address this.

docs/user-guide/configure-origins.md
```diff
@@ -0,0 +1,180 @@
...
+## Configuration file
+
+Backend services and origins known to Styx are configured in their own YAML file, which is referenced by the main 
+configuration via the `originsFile` property in the services block.
+
+The value for this `originsFile` property accepts including environment variables (with the ${ENV_VAR} format) and can reference:
+ - A path in the filesystem by default.
+ - A resource in the classpath by using the prefix *classpath:*.
+
+###Examples:
```
mikkokar: This heading doesn't render correctly. I believe you need to add a whitespace character between the ### marks and the heading text.

docs/user-guide/configure-tls.md
mikkokar: You are correct. The original use of "network interface" was also a poor choice of words. But it reads much better now for what I can see 👍

docs/user-guide/configure-tls.md
```diff
@@ -108,50 +107,55 @@ be specified as separate backends.
 
 ### Backend Application Attributes:
 
-  - trustAllCerts - Whether (true) or not (false) to authenticate server side.
+
+  - *trustAllCerts* -  When `trustAllCerts` is false,origin servers need to provide a
```
mikkokar: A missing whitespace between comma and "origin".

docs/user-guide/configure-tls.md
```diff
@@ -108,50 +107,55 @@ be specified as separate backends.
...
 
-      - *alias* - A symbolic name for the certificate. Used as an alias for keystore.
+      - *alias* - A symbolic name for the certificate. Used as an alias in the keystore.
 
       - *path*  - A path to the certificate file.
 
-  - *trustStorePath* - Styx can load the relevant secure material from Java truststore.
-    This attribute specifies a path to the keystore file containing relevant trust material.
+  - *trustStorePath* - Styx can load the list of trusted certificates from a Java JKS truststore.
+    This attribute specifies a path to the truststore file containing the trusted certificates.
```
mikkokar: I would remove the "file containing the trusted certificates." part. That is, just leave: "Styx can load the list of trusted certificates from a Java truststore. This attribute specifies a path to the truststore."

docs/user-guide/configure-overview.md
```diff
@@ -1,14 +1,17 @@
-## Configuring Styx environment
+## Configuring Styx
```
mikkokar: Hi. I believe this file has incorrect information about load balancer configurations. In particular:

```
    loadBalancing:
      # Load balancer strategy type. Allowed values: "LEAST\_RESPONSE\_TIME", "ROUND\_ROBIN", and "ADAPTIVE". Defaults to "ROUND_ROBIN".
      strategy: "ADAPTIVE"
```

I believe you'd have to supply the full class name using the factory. See the load balancer config page for an example.

docs/user-guide/configure-connection-pooling.md
```diff
@@ -21,38 +21,42 @@ An example configuration block looks like this:
       connectionExpirationSeconds: 1000 # default value 0
 
 
-Connection pool size is given by *maxConnectionsPerHost* setting. 
-The pool initially has no established TCP connections. 
+##General settings.
```
mikkokar: Rendering issue above.

docs/user-guide/configure-connection-pooling.md
```diff
@@ -21,38 +21,42 @@ An example configuration block looks like this:
...
+* *socketTimeoutMillis*: maximum length of time a TCP connection can remain idle before being closed. 
+*Warning:* this property is reserved for future usage and not currently honoured by Styx.
 
-The *connectionExpirationSeconds* setting allows connections to be terminated after a configured period of time has elapsed since the connection was created.
+* *connectionExpirationSeconds*: allows connections to be terminated after the configured number of seconds has elapsed since the connection was created.
 This is useful when an origin host is specified as a DNS domain name, and you want to ensure that domain names are re-resolved periodically.
 If the value of the setting is non-positive, connections will not expire. 
 Connection age is checked on each incoming request, so connections may live longer than their expiration time if they do not serve any requests.
 
+##Connection pending settings.
```
mikkokar: Rendering issue.

PR #116, opened 2018-03-22
Styx server refactoring

kvosper:
The `StyxServer` class was not very organised, and in fact was just a place where lots of code was dumped. This PR changes it to be primarily focused around holding the Styx core services, not setting them up.


components/proxy/src/main/java/com/hotels/styx/startup/CoreMetrics.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Core metrics - JVM details, Styx version.
+ */
+public final class CoreMetrics {
+    private static final Logger LOG = getLogger(CoreMetrics.class);
+
+    private CoreMetrics() {
+    }
+
+    public static void registerCoreMetrics(Environment environment) {
```
mikkokar: Instead of passing in all of the environment, pass in only the objects that are actually used: 1) buildInfo, and 2) metrics registry. It makes it easier to follow the code at the `registerCoreMetrics` call site.

mikkokar: We should really have some unit tests for this class.

components/proxy/src/main/java/com/hotels/styx/startup/ProxyServerSetUp.java
```diff
@@ -0,0 +1,85 @@
+/**
```
mikkokar: We need to add some unit tests:

- Notifies plugins when starts.
- Notifies plugins when stops.
- Logs plugin startup failures.
- Attempts to start all plugins before fails.
- Logs plugin shutdown failures.

components/proxy/src/main/java/com/hotels/styx/startup/StyxPipelineFactory.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+import static com.hotels.styx.api.configuration.ConfigurationContextResolver.EMPTY_CONFIGURATION_CONTEXT_RESOLVER;
+import static java.util.stream.Collectors.toMap;
+
+/**
+ * Produces the pipeline for the Styx proxy server.
+ */
+public final class StyxPipelineFactory {
+    private StyxPipelineFactory() {
```
mikkokar: This class could be further broken down in order to improve testability.

components/proxy/src/main/java/com/hotels/styx/startup/StyxPipelineFactory.java
```diff
@@ -0,0 +1,143 @@
...
+                builtinInterceptorsFactory);
+
+        RouteHandlerFactory routeHandlerFactory = new RouteHandlerFactory(objectFactories, new ConcurrentHashMap<>());
+
+        return styxHttpPipeline(
+                config.environment().styxConfig(),
+                configuredPipeline(config.environment(), config.services(), config.plugins(), routeHandlerFactory));
+    }
+
+    private static HttpHandler2 styxHttpPipeline(StyxConfig config, HttpHandler2 interceptorsPipeline) {
```
mikkokar: This "unit" of code adds the default "built-in" interceptors around the configured HTTP pipeline. So we should have following unit tests:

 - Adds default interceptors in front of user-configured pipeline.
 - Adds logging interceptor when logging configuration is enabled
 - Enables long logging messages when ...

mikkokar: However, in order to make this easier, the `StandardHttpPipeline` class should be amalgamated into `HttpInterceptorPipeline`. Then you could expose its `interceptors` list as a `@VisibleForTesting` attribute, and ensure the interceptors are prepended as expected.

kvosper: This code was copied wholesale from the previous version of `StyxServer`, so I will take some time to wrap my head around all of it. It may have to be done as a separate increment - I don't know yet.

components/proxy/src/test/java/com/hotels/styx/startup/StyxServerComponentsTest.java
```diff
@@ -0,0 +1,133 @@
...
+    public void loadsPlugins() {
+        Chain chain = mock(Chain.class);
+
+        List<NamedPlugin> pluginsForBuilder = ImmutableList.of(
+                namedPlugin("plugin1", stubPlugin("MyResponse1")),
+                namedPlugin("plugin2", stubPlugin("MyResponse2")));
+
+        StyxServerComponents components = new StyxServerComponents.Builder()
+                .styxConfig(new StyxConfig())
+                .plugins(env -> pluginsForBuilder)
```
mikkokar: Please inline `pluginsForBuilder` for clarity. It is only used here.

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
```diff
@@ -0,0 +1,171 @@
...
+        return new Environment.Builder()
+                .configuration(styxConfig)
+                .metricsRegistry(new CodaHaleMetricRegistry())
+                .healthChecksRegistry(new HealthCheckRegistry())
+                .buildInfo(readBuildInfo())
+                .eventBus(new AsyncEventBus("styx", newSingleThreadExecutor()))
+                .build();
+    }
+
+    private static Version readBuildInfo() {
```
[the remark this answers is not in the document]
kvosper: I will fix this.

components/proxy/src/test/java/com/hotels/styx/startup/StyxServerComponentsTest.java
```diff
@@ -0,0 +1,133 @@
...
+                .build();
+
+        List<NamedPlugin> plugins = ImmutableList.copyOf(components.plugins());
```
mikkokar: Is there any good reason why the `components.plugins()` needs to return an `Iterable` instead of `List`? Now each consumer must always convert it to a list before it is any good.

components/proxy/src/test/java/com/hotels/styx/startup/StyxServerComponentsTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+        assertThat(names, contains("plugin1", "plugin2"));
+
+        List<String> responses = plugins.stream()
+                .map(plugin -> plugin.intercept(get("/").build(), chain))
+                .map(BlockingObservables::waitForResponse)
+                .map(resp -> resp.bodyAs(UTF_8))
+                .collect(toList());
+
+        assertThat(responses, contains("MyResponse1", "MyResponse2"));
```
mikkokar: The assertions to verify plugin responses are useless IMO. Please remove them. All this unit is doing is to load the plugins using `pluginLoader` and to expose the loaded plugins via `plugins()` method.

components/proxy/src/test/java/com/hotels/styx/startup/StyxServerComponentsTest.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Test
+    public void loadsServices() {
+        StyxService service1 = mock(StyxService.class);
+        StyxService service2 = mock(StyxService.class);
```
mikkokar: These two can be inlined AFAICS.

components/proxy/src/test/java/com/hotels/styx/startup/StyxServerComponentsTest.java
```diff
@@ -0,0 +1,133 @@
...
+                .map(plugin -> plugin.intercept(get("/").build(), chain))
+                .map(BlockingObservables::waitForResponse)
+                .map(resp -> resp.bodyAs(UTF_8))
+                .collect(toList());
+
+        assertThat(responses, contains("MyResponse1", "MyResponse2"));
+    }
+
+    @Test
+    public void loadsServices() {
```
mikkokar: This test is not testing for service loading aspect. It is only verifying that additional services are added to the exposed services list. Consider renaming this method to "exposesAdditionalServices".

components/proxy/src/test/java/com/hotels/styx/startup/StyxServerComponentsTest.java
```diff
@@ -0,0 +1,133 @@
...
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static rx.Observable.just;
+
+// TODO it is still difficult to decouple + test the configured plugins + services
```
mikkokar: Looks like you are having difficulties to test service loading. That is because the services are loaded by a separate static method which cannot be mocked.

components/proxy/src/test/java/com/hotels/styx/startup/CoreMetricsTest.java
```diff
@@ -0,0 +1,126 @@
...
+import com.hotels.styx.api.metrics.codahale.CodaHaleMetricRegistry;
+import org.testng.annotations.Test;
+
+import java.util.Map;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.is;
+
+public class CoreMetricsTest {
```
mikkokar: 👍

components/proxy/src/test/java/com/hotels/styx/startup/ProxyServerSetUpTest.java
```diff
@@ -0,0 +1,257 @@
...
+        PipelineFactory pipelineFactory = mock(PipelineFactory.class);
+        HttpHandler2 pipeline = mock(HttpHandler2.class);
+        when(pipelineFactory.create(components)).thenReturn(pipeline);
+
+        server = new ProxyServerSetUp(pipelineFactory).createProxyServer(components);
+
+        server.startAsync().awaitRunning(3, SECONDS);
+
+        assertThat(server.httpAddress().getPort(), is(httpPort));
+        assertThat(server.httpsAddress().getPort(), is(httpsPort));
```
mikkokar: Hmm... we really should be able to run most of the tests without actually having to bind to the TCP ports. Could you have a quick look if the HTTP server connectors could be somehow mocked for testing?

components/proxy/src/test/java/com/hotels/styx/startup/StyxServerComponentsTest.java
kvosper: I will look into this.

components/proxy/src/test/java/com/hotels/styx/startup/ProxyServerSetUpTest.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    @Test
+    public void createsServer() throws TimeoutException {
+        StyxServerComponents components = new StyxServerComponents.Builder()
+                .styxConfig(new StyxConfig(config))
+                .build();
+
+        PipelineFactory pipelineFactory = mock(PipelineFactory.class);
```
mikkokar: You could create a `mockPipelineFactory` method to construct the HTTP pipeline object. It could save about 3 lines of code per test.

components/proxy/src/test/java/com/hotels/styx/startup/StyxServerComponentsTest.java
kvosper: It's a way to verify that the plugins are what we expect. I can refactor to just check equality, unless you have a preferred alternative?

components/proxy/src/test/java/com/hotels/styx/startup/AdminServerSetUpTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+import static com.hotels.styx.api.support.HostAndPorts.freePort;
+import static java.util.Collections.emptyList;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class AdminServerSetUpTest {
```
mikkokar: I would get rid of this test, because
- The AdminServerSetup has only one static method that initialises the admin server builder.
- This really is an integration test, not an unit test. We ought to have the integration tests in e2e suite.
- I doubt this adds any more value over the e2e suite we already have in place.

mikkokar:
Looks good so far. Please finalise the rebase and I will have another look.

PR #120, opened 2018-03-27
Fixes #118 Improve FlowControllingHttpContentProducer state machine.

### LINKED ISSUE #118: Improve FlowControllingHttpContentProducer state machine
## The problem

Confusing warning messages from the FlowControllingHttpContentProducer when origin server closes the TCP connection:

```
    WARN  com.hotels.styx.client.netty.connectionpool.FlowControllingHttpContentProducer 2018-03-03 02:02:02,000 [Styx-Client-Worker-14-Thread] 
    - message="Inappropriate event=ChannelIdleEvent", prefix=myserver.com/1.1.5.1:22222 -> /2.1.1.2:55555, Request(method=POST, url=/some/page.html, i0=00000000-0000-0000-0000-000000000000), state=TERMINATED, receivedChunks=0, receivedBytes=0, emittedChunks=0, emittedBytes=0
```

This suggests the TCP connection is closed due to idleness. But in fact the warning also pops up when the channel has been subjected to a "delayed closure". This happens when the origin server closes the TCP connection to indicate response content length (See _RFC 7230, Chapter 3.3.3.  Message Body Length, item 7._) Tomcat apps have been seen to behave this way when they respond with HTTP error status code (4xx or 5xx). 

A clarification regarding "delayed closure". It is the Flow Controller state machine associated to the TCP connection that is subjected to the delayed closure. The TCP connection itself is closed immediately. The Flow Controller state machine regulates the delivery of the HTTP message body to the rest of the styx honouring the back pressure and Rx.Java observable protocol. The purpose of the delayed closure is to allow styx server some time to consume the response in order to proxy it back.

## Detailed description

Back in the days when we realised it was necessary to implement the delayed closure as described above, we hijacked the FlowController's "idle state event" signal to inform the state machine to tear down any resources. This is in `NettyToStyxResponsePropagator.java`:

```
    @Override
    public void channelInactive(ChannelHandlerContext ctx) throws Exception {
        ensureContentProducerIsCreated(ctx);

        ctx.channel().eventLoop().schedule(
                () -> contentProducer.ifPresent(producer -> producer.idleStateEvent(
                        new ResponseTimeoutException(
                                origin,
                                "channelInactive",
                                producer.receivedBytes(),
                                producer.receivedChunks(),
                                producer.emittedBytes(),
                                producer.emittedChunks()))),
                idleTimeoutMillis,
                MILLISECONDS);

        TransportLostException cause = new TransportLostException(ctx.channel().remoteAddress(), origin);
        contentProducer.ifPresent(producer -> producer.channelInactive(cause));
    }
```

The misleading error message is caused by this re-use of the idle state event, and the time has come to address this issue.

## Acceptance criteria
* Introduce a new `delayed teardown` (or something similar) event to replace the usage of `idle state event` to tear down the Flow Controller state machine.
* The handling of `delayed teardown` in the Flow Controller state machine:
  * The 'delayed teardown' must be handled in `BUFFERING_COMPLETED` and `EMITTING_BUFFERED_CONTENT` states. 
  * In `BUFFERING_COMPLETED` state: release all buffered content and transition to `TERMINATED` state.
  * In `EMITTING_BUFFERED_CONTENT` state, release all buffered content, emit error on content observable, and transition to `TERMINATED` state.
  * In `COMPLTED` state, swallow the event to prevent an unnecessary warning message.
  * Leave it unhandled in all other states, so that an informational warning message is logged.
* The 'idle state event' is only used to inform the state machine about prolonged inactivity on the TCP connection.

dvlato:
New event, DelayedTearDownEvent, created to replace IdleStateEvent.  
Currently there's no handling of idle events in the FSM in `FlowControllingHttpContentProducer`, so the logic has been kept untouched.

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/FlowControllingHttpContentProducer.java
+++ b/components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/FlowControllingHttpContentProducer.java
@@ -47,12 +47,13 @@ import static rx.internal.operators.BackpressureUtils.getAndAddRequest;
 class FlowControllingHttpContentProducer {
     private static final Logger LOGGER = LoggerFactory.getLogger(FlowControllingHttpContentProducer.class);
 
-    private final StateMachine<ProducerState> stateMachine;
+     private final StateMachine<ProducerState> stateMachine;
     private final String loggingPrefix;
 
     private final Runnable askForMore;
     private final Runnable onCompleteAction;
     private final Consumer<Throwable> onTerminateAction;
+    private Runnable delayedTearDownAction;
 
     private final Queue<ByteBuf> readQueue = new ConcurrentLinkedDeque<>();
     private final AtomicLong requested = new AtomicLong(Long.MAX_VALUE);
@@ -60,6 +61,8 @@ class FlowControllingHttpContentProducer {
     private final AtomicLong receivedBytes = new AtomicLong(0);
     private final AtomicLong emittedChunks = new AtomicLong(0);
     private final AtomicLong emittedBytes = new AtomicLong(0);
+
+    private final Origin origin;
 
     private volatile Subscriber<? super ByteBuf> contentSubscriber;
 
@@ -76,11 +79,14 @@ class FlowControllingHttpContentProducer {
             Runnable askForMore,
             Runnable onCompleteAction,
             Consumer<Throwable> onTerminateAction,
+            Runnable delayedTearDownAction,
             String loggingPrefix,
             Origin origin) {
         this.askForMore = requireNonNull(askForMore);
         this.onCompleteAction = requireNonNull(onCompleteAction);
         this.onTerminateAction = requireNonNull(onTerminateAction);
+        this.delayedTearDownAction = requireNonNull(delayedTearDownAction);
+        this.origin = requireNonNull(origin);
 
         this.stateMachine = new StateMachine.Builder<ProducerState>()
                 .initialState(BUFFERING)
@@ -93,7 +99,7 @@ class FlowControllingHttpContentProducer {
 
                 .transition(BUFFERING_COMPLETED, RxBackpressureRequestEvent.class, this::rxBackpressureRequestBeforeSubscription)
                 .transition(BUFFERING_COMPLETED, ContentChunkEvent.class, this::spuriousContentChunkEvent)
-                .transition(BUFFERING_COMPLETED, ChannelInactiveEvent.class, s -> BUFFERING_COMPLETED)
+                .transition(BUFFERING_COMPLETED, ChannelInactiveEvent.class, s-> scheduleTearDown(BUFFERING_COMPLETED))
                 .transition(BUFFERING_COMPLETED, ChannelExceptionEvent.class, s -> BUFFERING_COMPLETED)
                 .transition(BUFFERING_COMPLETED, DelayedTearDownEvent.class, this::tearDown)
                 .transition(BUFFERING_COMPLETED, ContentSubscribedEvent.class, this::contentSubscribedEventWhileBufferingCompleted)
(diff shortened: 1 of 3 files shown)


components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyToStyxResponsePropagator.java
```diff
@@ -120,20 +120,16 @@ private RuntimeException toStyxException(Throwable cause) {
...
         ctx.channel().eventLoop().schedule(
-                () -> contentProducer.ifPresent(producer -> producer.idleStateEvent(
-                        new ResponseTimeoutException(
-                                origin,
-                                "channelInactive",
-                                producer.receivedBytes(),
-                                producer.receivedChunks(),
-                                producer.emittedBytes(),
-                                producer.emittedChunks()))),
+                () -> contentProducer.ifPresent(producer -> producer.tearDownResources(cause)),
```
mikkokar: The `tearDownResources` event is scheduled in a controlled fashion. That is, the `tearDownResources` is always scheduled after a short time to ensure state machine is orderly brought down. Therefore `cause` doesn't have to be passed in.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/FlowControllingHttpContentProducer.java
```diff
@@ -88,23 +87,23 @@
...
+                .transition(BUFFERING_COMPLETED, DelayedTearDownEvent.class, this::releaseAndTerminate)
                 .transition(BUFFERING_COMPLETED, ContentSubscribedEvent.class, this::contentSubscribedEventWhileBufferingCompleted)
                 .transition(BUFFERING_COMPLETED, ContentEndEvent.class, s -> BUFFERING_COMPLETED)
 
                 .transition(STREAMING, RxBackpressureRequestEvent.class, this::rxBackpressureRequestEventWhileStreaming)
                 .transition(STREAMING, ContentChunkEvent.class, this::contentChunkEventWhileStreaming)
                 .transition(STREAMING, ChannelInactiveEvent.class, e -> emitErrorAndTerminate(e.cause()))
                 .transition(STREAMING, ChannelExceptionEvent.class, e -> emitErrorAndTerminate(e.cause()))
-                .transition(STREAMING, ChannelIdleEvent.class, e -> emitErrorAndTerminate(new ResponseTimeoutException(origin)))
+                .transition(STREAMING, DelayedTearDownEvent.class, e -> emitErrorAndTerminate(e.cause()))
```
mikkokar: I would use a different method for the state transition handler. For example:

```
+                .transition(STREAMING, DelayedTearDownEvent.class, e -> tearDown())
```

And then in the `tearDown` method:

```
private void tearDown() {
   emitErrorAndTerminate(new TimedOutExceptionWhatever());  // Choose an appropriate exception here.
}
```

So I would have a dedicated method specifically for the `tearDown` transition. This exposes the state transition logic more explicitly.

dvlato: That sounds like an improvement, but in your example we are emitting an error, I assume, to signal an invalid combination of state+event.  Doesn't it make more sense to use a new 'tearDown' method for the cases in which we are calling 'releaseAndTerminate' ?

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyToStyxResponsePropagator.java
dvlato: I would say that if we have to publish an error, it will be more informative if it is the same error that caused the "channelInactive"  so they can be easily linked when reading the logs, that if we make up an Exception as in your example below: TimedOutExceptionWhatever. 
But I haven't seen the logs to be honest...

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyToStyxResponsePropagator.java
```diff
@@ -117,23 +117,18 @@ private RuntimeException toStyxException(Throwable cause) {
...
+    private void scheduleResourcesTearDown(ChannelHandlerContext ctx) {
         ctx.channel().eventLoop().schedule(
-                () -> contentProducer.ifPresent(producer -> producer.idleStateEvent(
-                        new ResponseTimeoutException(
-                                origin,
-                                "channelInactive",
-                                producer.receivedBytes(),
-                                producer.receivedChunks(),
-                                producer.emittedBytes(),
-                                producer.emittedChunks()))),
```
mikkokar: Please ensure these stats are passed in to the ResponseTimeoutException.
Even better please add a test that verifies that the counters are present in the exception when it is torn down.

mikkokar: Or instead log a warning using `FlowControllingHttpContentProducer.warningMessage`. This is the preferred option IMO.

mikkokar: The log messages can be tested with `LoggingTestSupport`. Please look for examples of usage in the codebase. It is quite easy to use.

mikkokar: Alternatively raise a separate GitHub enhancement issue.

PR #121, opened 2018-03-27
Plugin documentation and example pom.xml for plugins reviewed.

dvlato:
* Plugin documentation updated to use the latest version of the schema for the configuration of plugins.
* Simplified the pom.xml in the plugin example.
* Changed the scope of the styx-api dependency in the plugin pom,  so styx JARs (and third party dependencies such as guava or rxjava) are not bundled with the plugin.

### DIFF
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -9,25 +9,25 @@ Each plugin is contained within its own JAR file, located according to the confi
 ## Configuration example
 
     plugins:
-      active: rewrite
+      active: addheaders
       all:
-        - name: rewrite
+        addheaders:
           factory:
-            class: "com.hotels.styx.RewritePluginFactory"
+            class: "com.hotels.styx.ExamplePluginFactory"
             classPath: "<path-to-plugin>/plugin-examples-1.0-SNAPSHOT.jar"
           config:
-            oldUri: "/olduri"
-            newUri: "/newuri"
+            requestHeaderValue: "requestheader"
+            responseHeaderValue: "responseheader"
       
 ### Configuration example explanation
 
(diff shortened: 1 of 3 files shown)


docs/developer-guide/plugins.md
```diff
@@ -9,25 +9,25 @@ Each plugin is contained within its own JAR file, located according to the confi
...
 * **all** contains a list of configuration for all the plugins deployed with the Styx instance. Inside each list item:
-    * **name** is the name of the plugin
+    * **addheaders** is the name of the plugin
     * **factory** configures a factory object that can produce the plugin:
         * **class** contains the name of the factory class, which must extend `com.hotels.styx.api.plugins.spi.PluginFactory`
         * **classPath** provides the location of the JAR file
-    * **config** contains custom configuration for the particular plugin. Its value can be of any type. See below for how to write a custom config java class.
+    * **config** contains custom configuration for the particular plugin. Its value can be of any type. See below how to write a custom config java class.
 
 When Styx starts, it sets up the HTTP interceptor chain as follows:
```
mikkokar: There is a chapter called `Custom Configuration` that is somewhat out of place. I would move that further down under the `Developing Plugin` chapter.

docs/developer-guide/plugins.md
```diff
@@ -9,25 +9,25 @@ Each plugin is contained within its own JAR file, located according to the confi
 ## Configuration example
```
mikkokar: Hi David. At the top of the file it claims that plugins can be "...implemented in any JVM language..". Well, that is only true if the JVM language can talk to Java. Because we don't have any particular language adapters besides Java I would remove this note from the document.

PR #126, opened 2018-04-06
Startup script does not fail if the order of the arguments is incorrect #125

### LINKED ISSUE #125: Startup script does not fail if the order of the arguments is incorrect
## The problem
When running the startup script with several arguments (for instance, adding -e) the passed in config filename is ignored silently if it is not the last argument.
## Detailed description
The startup script expects the config filename to be the last argument for the command, but if it is not, instead of failing it will just ignore that config filename completely.
Example:

- `bin/startup -e conf/styx-env.sh conf2/environment/default.yaml` starts the server correctly  using the chosen config files.

- `bin/startup conf2/environment/default.yaml -e conf/styx-env.sh` starts the server but it ignores the specified config filename and uses the default one (conf/default.yml

The script should be fixed so it will return an error when the order of the arguments is incorrect.

## Acceptance criteria

* When the config filename is not the last argument for the command, an error plus the usage instructions  will be returned: 
 `bin/startup conf2/environment/default.yaml -e conf/styx-env.sh`  will return an error immediately.

dvlato:
Fixes #125 
Startup script fixed so it correctly returns an error when an argument is unknown.

### DIFF
--- a/distribution/bin/startup
+++ b/distribution/bin/startup
@@ -59,7 +59,7 @@ do
       STYX_ENV_FILE=$2
       shift
       ;;
-    -*)
+    *)
       echo "Unknown option: $1 ($@)"
       echo
       usage


kvosper:
Could you provide a shell command we can execute to compare the master version with this branch?

PR #128, opened 2018-04-06
Add Styx fallback configuration

kvosper:


components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -87,11 +84,7 @@ private static StyxServer createStyxServer(String[] args) {
             LOG.info("Styx configFileLocation={}", startupConfig.configFileLocation());
             LOG.info("Styx logConfigLocation={}", startupConfig.logConfigLocation());
 
-            Configuration configFromFile = new ConfigurationParser.Builder<YamlConfiguration>()
-                    .format(YAML)
-                    .overrides(System.getProperties())
-                    .build()
-                    .parse(configSource(startupConfig.configFileLocation()));
+            Configuration configFromFile = configurationFromFile(startupConfig, "classpath:/configuration/default.yml");
```
dvlato: Do you want to hardcode the path the fallback file?

distribution/conf/env-development/styx-config.yml
```diff
@@ -1,5 +1,3 @@
-include: ${STYX_HOME:classpath:}/conf/default.yml
-
```
mikkokar: Not sure why this is removed?

I understand we will still support configuration inheritance, and this would have served at least as an example how it is used?

components/proxy/src/main/resources/configuration/default.yml
```diff
@@ -87,3 +87,4 @@ url:
     unwiseCharactersToEncode:
 
 #originRestrictionCookie: yourCookieNameHere
+
```
mikkokar: Any commented should be removed from the default file.

mikkokar: Now the question is if the default values are sensible. Like. `bossThreadCount: 1`? Perhaps that is OK.

How about proxy and admin connectors? Is possible to start styx in HTTPS only?

mikkokar: Also, are we able to unset the `backendServiceRegistry` at all?

mikkokar: My problem with this change is that it results in confusing behaviour from end-user's point of view. When I'll start the server with following config, I would expect styx to open up HTTPS port *only* but it starts the insecure HTTP port too:

```
proxy:
  connectors:
    https:
      sslProvider: JDK
      certificateFile: ${STYX_HOME:classpath:}/conf/tls/testCredentials.crt
      certificateKeyFile: ${STYX_HOME:classpath:}/conf/tls/testCredentials.key
      cipherSuites:
        - TLS_RSA_WITH_AES_128_GCM_SHA256
```

The styx now started with:

```
"proxy": {
  "connectors": {
    "http": {
      "port": 8080
    },
    "https": {
      "port": 8443,
      "sslProvider": "JDK",
      "certificateFile": "/Users/mikko/src/github/styx-mikko/styx-0.7-SNAPSHOT/conf/tls/testCredentials.crt",
      "certificateKeyFile": "/Users/mikko/src/github/styx-mikko/styx-0.7-SNAPSHOT/conf/tls/testCredentials.key",
      "sessionTimeoutMillis": 300000,
      "sessionCacheSize": 20000,
      "cipherSuites": [
         "TLS_RSA_WITH_AES_128_GCM_SHA256"
      ]
    }
},```

kvosper:
This PR will not be merged (see Mikko's comments). This does not invalidate the idea of default values for unspecified configuration, but tying it into the YAML-based configuration system introduces too many potential problems.

Issue #130, opened 2018-04-06
Add 4xx server response codes to styx documentation


mikkokar:
## The problem

The 400, 408, etc (if any others?) response code is not documented in https://github.com/HotelsDotCom/styx/blob/master/docs/user-guide/response-codes.md.

## Acceptance criteria

The user guide response codes page contains the following information:
* The circumstances under which styx responds with 400.
* The circumstances under which styx responds with 408.

The relevant user manual page: https://github.com/HotelsDotCom/styx/blob/master/docs/user-guide/response-codes.md.

kvosper:
We should create a separate page for this documentation, since 5xx response codes indicate a server error, whereas 4xx response codes are client errors.

The significance of this distinction is that server errors indicate that there is a problem that someone running Styx should look into, whereas client errors are likely to be caused by the end-users' actions.

mikkokar:
@kvosper Please remember to close the issue after merging.

PR #135, opened 2018-04-10
Add extra documentation

kvosper:
See #130


docs/user-guide/response-codes-client-error.md
```diff
@@ -0,0 +1,22 @@
...
+Occurs when a request is somehow incorrectly formatted. Examples:
+
+* Unrecognised HTTP verb
+* Multiple `Host` headers.
+
+
+## 408 Request Timeout
+
+Occurs if the channel from the client to Styx becomes idle, for example if the client starts sending a request
+but does not complete it.
```
mikkokar: Please elaborate these aspects:

- More specifically, this can be thought of an "origin inactivity timeout". That is:
  - While origin is sending a HTTP Request, it must send something within `proxy.requestTimeoutMillis` (I believe this timeout can be configured for the admin interface too). Any inactivity exceeding this will trigger a 408 response code.
  - The timer applies only while HttpRequest is being transmitted "over the wire". That is, the timer is stopped when styx has fully received request. Therefore any delay or inactivity introduced by origin responding slowly will not trigger this timeout.

docs/user-guide/response-codes-client-error.md
```diff
@@ -0,0 +1,22 @@
...
+they may introduce 4xx statuses for other reasons not listed here.
+
+
+## 400 Bad Request
+
+Occurs when a request is somehow incorrectly formatted. Examples:
+
+* Unrecognised HTTP verb
+* Multiple `Host` headers.
+
```
mikkokar: The word `somehow` sounds bit unprofessional. Perhaps say `Occurs when the HTTP request format does not conform to HTTP 1.1 protocol specification.` etc?

Also, add `etc` at the end of the list to emphasise that the list of possible causes is not exhaustive.

docs/user-guide/response-codes-client-error.md
kvosper: At first, when you mention "origin" do you mean the client? This section is purely for client-errors and should have nothing to do with the origins.

mikkokar: Hi yes. That is correct. My mistake.

docs/user-guide/response-codes.md
```diff
@@ -1,15 +1,46 @@
 # Styx Response Codes
 
+Styx may return an HTTP error status code if it is unable to proxy a request.
+
+## Client Errors (4xx codes)
+
+When a client behaves in an unexpected way toward Styx, Styx will respond with specific error codes.
```
mikkokar: I would simplify along the lines of:

> "Styx responds with a specific HTTP error code when client's behaviour is unexpected or not conforming to HTTP protocol".

docs/user-guide/response-codes.md
```diff
@@ -1,15 +1,46 @@
...
+This refers only to responses generated by Styx itself, not responses being proxied from downstream.
```
mikkokar: This seems out of place:

> "This refers only to responses generated by Styx itself, not responses being proxied from downstream."

Perhaps move it up under "Styx Response Codes"?

docs/user-guide/response-codes.md
```diff
@@ -1,15 +1,46 @@
 # Styx Response Codes
 
+Styx may return an HTTP error status code if it is unable to proxy a request.
```
mikkokar: Perhaps:


> "Styx may return an HTTP error status code if it is unable to proxy a request.  These responses are generated by Styx itself, not responses being proxied from downstream. 
>
> This section describes those responses and the specific conditions when styx responds with them.
>
> "Note that interceptor plugins can also be programmed to modify response codes, so they may introduce 4xx or 5xx statuses for other reasons not listed here."

docs/user-guide/response-codes.md
```diff
@@ -1,15 +1,46 @@
...
+
+## Client Errors (4xx codes)
+
+When a client behaves in an unexpected way toward Styx, Styx will respond with specific error codes.
+This refers only to responses generated by Styx itself, not responses being proxied from downstream.
+For a general explanation of 4xx (Server Error) codes, please see the HTTP Spec.
+
+Note that interceptor plugins can also be programmed to modify response codes, so 
+they may introduce 4xx statuses for other reasons not listed here.
```
mikkokar: I would move this up into the "Styx Response Codes" section.

docs/user-guide/response-codes.md
```diff
@@ -1,15 +1,46 @@
...
+there is an unrecognised HTTP verb or multiple `Host` headers.
+
+
+### 408 Request Timeout
+
+While origin is client a HTTP Request, it must send something within `proxy.requestTimeoutMillis`.
+Any inactivity exceeding this will trigger a 408 response code. The timer applies only while the 
+HTTP request is being transmitted "over the wire". The timer is stopped when styx has fully 
+received request, so any delay or inactivity introduced by origin responding slowly will not 
+trigger this timeout.
```
mikkokar: > "While origin is client a HTTP Request"

???

kvosper: Looks like I replaced the wrong word while editing.

docs/user-guide/response-codes.md
```diff
@@ -1,15 +1,44 @@
...
+
+
+### 408 Request Timeout
+
+While the client a HTTP Request, it must send something within `proxy.requestTimeoutMillis`.
+Any inactivity exceeding this will trigger a 408 response code. The timer applies only while the 
+HTTP request is being transmitted "over the wire". The timer is stopped when styx has fully 
+received request, so any delay or inactivity introduced by origin responding slowly will not 
+trigger this timeout.
+
```
mikkokar: I would paraphrase:

>Styx responds with 408 Request Timeout error when a threshold of inactivity is observed on the underlying TCP connection while receiving an HTTP request.
>
>When a remote client **is sending** an HTTP Request, it must demonstrate some presence by continuously >sending the bytes over the network. If the client stops sending the request half way through, or slows down too >much, Styx will observe a duration of inactivity on the underlying TCP connection and rejects the request with >408 Request Timeout. 
>
>The `proxy.requestTimeoutMillis` server configuration parameter controls the amount of inactivity is tolerated.
>
>The timer applies only while the HTTP request is being transmitted "over the wire". The timer is stopped when >the request is fully received, so any delay or inactivity introduced by the upstream origins will not trigger this >timeout.

PR #138, opened 2018-04-12
Multiple jar plugin

### LINKED ISSUE #56: Plugin cannot consist of multiple JAR files
## The problem
It is not possible to load a plugin that consists of multiple JAR files.

## Detailed description
The implementation is bugged because it loads each JAR (i.e. dependency) into a separate classloader (and the classloaders do not have visibility of each other). This forces the plugin developer to merge all of the JARs together in order to make it work.

## Acceptance criteria
* Load all JARs in the plugin JAR directory into a single classloader
* A Plugin classloader should use the system classloader as its parent, not the Styx classloader (so that it uses the plugin's version of its dependencies)

kvosper:
Fixes #56


components/proxy/src/main/java/com/hotels/styx/spi/ObjectFactories.java
```diff
@@ -53,60 +52,44 @@ private ObjectFactories() {
...
-                LOGGER.error("error loading class={}", className);
-                return Optional.empty();
-            }
-        }
-
-        int size = instances.size();
-        if (size == 0) {
+        ClassLoader classLoader = pathOptional
+                .map(path -> classLoader(readResourcesFrom(path)))
+                .map(ClassLoader.class::cast)
```
dvlato: Why don't you just changed= the return type of the classLoader method to be "ClassLoader" instead of UrlClassLoader and get rid of this cast?

components/proxy/src/main/java/com/hotels/styx/spi/ObjectFactories.java
```diff
@@ -53,60 +52,44 @@ private ObjectFactories() {
...
+
+        try {
+            Class<?> aClass = Class.forName(className, false, classLoader);
+            T instance = type.cast(aClass.newInstance());
+            return Optional.of(instance);
+        } catch (ClassNotFoundException e) {
             LOGGER.error("no class={} is found in the specified classpath={}", className, pathOptional.orElse(Paths.get("")));
             return Optional.empty();
+        } catch (InstantiationException | IllegalAccessException e) {
+            LOGGER.error("error loading class={}", className);
```
dvlato: I'm not too sure that using Optionals here is the best option. Right now, when we return empty we just throw an Exception, but that Exception doesn't include the real cause, which we are logging separatedly. Isn't it easier to just propagate the Exception and log the full error? 
Thoughts?

system-tests/example-styx-plugin-dependencies/src/main/java/depend/ExampleDependency.java
```diff
@@ -0,0 +1,26 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package depend;
+
+/**
+ * A project can attempt to reference this to prove the dependency relationship is in effect.
+ */
+public final class ExampleDependency {
+    public static final String EXAMPLE_DEPENDENCY_PROPERTY = "Example-Dependency-Available";
```
dvlato: final Strings are not loaded dynamically but rather compiled inside the dependent byte code.  That is, this line : 
LOGGER.info(EXAMPLE_DEPENDENCY_PROPERTY);  will not try to load the dependent JAR file and obtain the  EXAMPLE_DEPENDENCY_PROPERTY at all:  ff you decompile the bytecode for  TestPluginModule.java  you will see it is just doing "logger.info("Example-Dependency-Available")

dvlato:
I have just tested it, and the scalatest includes the plugin jars  in the classpath despite being 'provided dependencies' ( the jars included in the classpath are listed in the TEST-com.hotels.styx.plugins.PluginPipelineSpec.xml), so we are effectively loading the plugin from the application classpath. 

We should find a way around that if we really want to test we are loading the plugins correctly.

kvosper:
Note: I have added some refactoring, this is not intended to address @dvlato 's most recent comment, but to make unit-testable classes (mocking the classloading).

kvosper:
@dvlato Regarding your comment "the scalatest includes the plugin jars in the classpath" - is it the *contents* of the JAR or only the JAR itself? Does the normal classloader look inside the JARs or just regard them as non-class files?

dvlato:
If a jar is listed in the classpath, it is available. Otherwise only its
folder would be listed.
Classpaths usally reference:

1) Folders. The classloader will find classes inside the folder.
2) individual jar files. The classloader will find classes inside the jar
file.

That's why in startup scripts we list all the jar files one by one. That's
obvously different from other kinds of packaging (war, ear, rar files) that
can contain nested jar files inside.

I would assume that can add precondition assertions before the test : try
to load the plugin with the current (application/system) classloader and
assert that it throws a ClassNotFoundException.

system-tests/e2e-suite/src/test/java/com/hotels/styx/E2EJarAccessTest.java
```diff
@@ -0,0 +1,98 @@
...
+import static com.hotels.styx.ExamplePluginJarLocation.examplePluginJarLocation;
+import static java.lang.ClassLoader.getSystemClassLoader;
+import static java.nio.file.Files.list;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+// This test is primarily ensuring that our supporting classes and modules are fit-for-purpose
+// i.e. that our example plugin and its dependency are only accessible by legitimately loading their
+// jar files and not just included in the e2e-suite module
+public class E2EJarAccessTest {
```
dvlato: This is not an end to end test at all. Is it?

system-tests/e2e-suite/src/test/java/com/hotels/styx/ExamplePluginJarLocation.java
```diff
@@ -0,0 +1,75 @@
...
+    static Path exampleDependencyJarLocation() throws IOException {
+        return jarLocation("example-styx-plugin-dependencies");
+    }
+
+    // module must be adjacent to e2e-suite
+    private static Path jarLocation(String module) throws IOException {
+        Path systemRoot = Paths.get("/");
+        Path classPathRoot = Paths.get(getSystemClassLoader().getResource("").getFile());
+
+        Path parent = systemRoot
```
dvlato: I don't think using 'systemRoot' here is correct. For instance, it would fail in windows when using different drives. I would rather use classPathRoot.getRoot()
Additionally, the code is difficult to understand because "subpath" is a bit strange in that it drops the root (and it's not clear in the documentation).  Maybe we can encapsulate it in a different way so the code is more readable?

system-tests/e2e-suite/src/test/java/com/hotels/styx/E2EJarAccessTest.java
kvosper: The name comes from the fact that it is checking the behaviour inside the `e2e-suite` module. I am happy to change it to something else.

PR #140, opened 2018-04-18
Prevent duplicate backend service paths

### LINKED ISSUE #139: Do not allow backend services to load or reload when path prefix is duplicated
## The problem
When Styx tries to load at start-up, or when a reload is executed, it should check that every backend service has its own unique path prefix and refuse to start otherwise.

## Acceptance criteria
* If there is a duplicate path prefix at start-up Styx does not start and logs an error
* If there is a duplicate path prefix at reload, Styx:
    * does not accept the new backend-service config
    * continues using the old config
    * sends an error response (assuming HTTP command)
    * logs an error

kvosper:
Fixes #139 .

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/service/spi/AbstractRegistry.java
+++ b/components/api/src/main/java/com/hotels/styx/api/service/spi/AbstractRegistry.java
@@ -22,12 +22,16 @@ import com.hotels.styx.api.Id;
 import com.hotels.styx.api.Identifiable;
 import org.slf4j.Logger;
 
+import java.util.Collection;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Predicate;
 
+import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.Maps.difference;
 import static com.google.common.collect.Maps.filterKeys;
 import static java.util.Collections.emptyList;
+import static java.util.Objects.requireNonNull;
 import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.StreamSupport.stream;
@@ -43,6 +47,15 @@ public abstract class AbstractRegistry<T extends Identifiable> implements Regist
 
     private final Announcer<Registry.ChangeListener> announcer = Announcer.to(ChangeListener.class);
     private final AtomicReference<Iterable<T>> snapshot = new AtomicReference<>(emptyList());
+    private final Predicate<Collection<T>> resourceConstraint;
+
+    public AbstractRegistry() {
+        this(any -> true);
+    }
+
+    public AbstractRegistry(Predicate<Collection<T>> resourceConstraint) {
+        this.resourceConstraint = requireNonNull(resourceConstraint);
+    }
 
     @Override
     public Iterable<T> get() {
@@ -63,6 +76,9 @@ public abstract class AbstractRegistry<T extends Identifiable> implements Regist
 
     public void set(Iterable<T> newObjects) {
         ImmutableList<T> newSnapshot = ImmutableList.copyOf(newObjects);
+
+        checkState(resourceConstraint.test(newSnapshot));
+
         Iterable<T> oldSnapshot = snapshot.get();
         snapshot.set(newSnapshot);
 
--- a/components/api/src/main/java/com/hotels/styx/api/service/spi/AbstractRegistry.java
+++ b/components/api/src/main/java/com/hotels/styx/api/service/spi/AbstractRegistry.java
@@ -49,10 +49,19 @@ public abstract class AbstractRegistry<T extends Identifiable> implements Regist
     private final AtomicReference<Iterable<T>> snapshot = new AtomicReference<>(emptyList());
     private final Predicate<Collection<T>> resourceConstraint;
 
+    /**
+     * Constructs an abstract registry with no resource constraint.
+     */
     public AbstractRegistry() {
         this(any -> true);
     }
 
+    /**
+     * Constructs an abstract registry that will not allow resources to be set if they do not
+     * satisfy a particular constraint.
+     *
+     * @param resourceConstraint a constraint to be satisfied
+     */
     public AbstractRegistry(Predicate<Collection<T>> resourceConstraint) {
         this.resourceConstraint = requireNonNull(resourceConstraint);
     }
@@ -74,7 +83,13 @@ public abstract class AbstractRegistry<T extends Identifiable> implements Regist
         return this;
     }
 
-    public void set(Iterable<T> newObjects) {
+    /**
+     * Sets the contents of this registry to the supplied resources.
+     *
+     * @param newObjects new contents
+     * @throws IllegalStateException if the resource constraint is not satisfied
+     */
+    public void set(Iterable<T> newObjects) throws IllegalStateException {
         ImmutableList<T> newSnapshot = ImmutableList.copyOf(newObjects);
 
         checkState(resourceConstraint.test(newSnapshot));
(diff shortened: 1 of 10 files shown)


system-tests/e2e-suite/src/test/scala/com/hotels/styx/admin/OriginsPathDuplicationOnReloadSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+  it("Responds with INTERNAL_SERVER_ERROR when the backend-services have a duplicate path") {
+    val fileBasedBackendsRegistry = FileBackedBackendServicesRegistry.create(styxOriginsFile.toString)
+    styxServer = StyxConfig().startServer(fileBasedBackendsRegistry)
+
+    styxServer.isRunning should be(true)
+
+    Files.copy(originsDuplicatePath, styxOriginsFile, REPLACE_EXISTING)
+
+    val resp = decodedRequest(post(styxServer.adminURL("/admin/tasks/origins/reload")).build())
+    resp.status() should be(INTERNAL_SERVER_ERROR)
```
mikkokar: Why would we respond with INTERNAL SERVER ERROR when this is fundamentally a client error?

system-tests/e2e-suite/src/test/scala/com/hotels/styx/admin/OriginsPathDuplicationOnReloadSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+  val originsOk = fixturesHome(classOf[OriginsPathDuplicationOnReloadSpec], "/conf/origins/origins-correct.yml")
+  val originsDuplicatePath = fixturesHome(classOf[OriginsPathDuplicationOnReloadSpec], "/conf/origins/origins-duplicate-path.yml")
+  val styxOriginsFile = Paths.get(tempDir.toString, "origins.yml")
+  var styxServer: StyxServer = _
+
+  it("Responds with INTERNAL_SERVER_ERROR when the backend-services have a duplicate path") {
+    val fileBasedBackendsRegistry = FileBackedBackendServicesRegistry.create(styxOriginsFile.toString)
+    styxServer = StyxConfig().startServer(fileBasedBackendsRegistry)
+
+    styxServer.isRunning should be(true)
```
mikkokar: I believe `styxServer.isRunning` is a pointless assertion.  This line is never reached if/when styx server fails to start.

components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
```diff
@@ -122,11 +131,12 @@ public String fileName() {
     }
 
     private boolean updateResources(byte[] content, HashCode hashCode) {
-
-        Iterable<T> resources = reader.read(content);
+        Collection<T> resources = reader.read(content);
         Changes<T> changes = changes(resources, get());
 
         if (!changes.isEmpty()) {
+            checkState(resourceConstraint.test(resources));
```
mikkokar: IMO this check should take place in the abstract registry `set` method. That way the check is not limited to the file based providers only.

mikkokar: Also, it is not checking for resource constrains, such as limits in RAM, disk or CPU. It is checking that the configuration is internally consistent

So perhaps this check needs to be renamed.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/admin/OriginsPathDuplicationOnReloadSpec.scala
kvosper: It's not a client error. The client sent a valid request for the origins to be reloaded. It was the state of the origin config that was invalid.

I presume you made the same observation when you changed the response code in your PR #108 ?

components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
kvosper: The name `resourceConstraint` is consistent with the variable `resources` that it is checking. What would you name the objects that registries contain?

I will look into moving the functionality to `AbstractRegistry.set`.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/admin/OriginsPathDuplicationOnReloadSpec.scala
mikkokar: Aah right.

components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
mikkokar: Ok I see.  Regardless, thinking from `AbstractRegistry` point of view, it doesn't have a concept of resources "resources", just "registry". Perhaps call it a `registryConstraint`?

kvosper: `AbstractRegistry` has a method ` protected static <T extends Identifiable> Changes<T> changes(Iterable<T> newResources, Iterable<T> currentResources)`, so the term is already used. We can still rename them to something else though.
`registryConstraint` is not really informative though, that would be naming it after the object it lives in, rather than the objects it constrains.
If we rename "resources" throughout the code, I will rename the constraint accordingly.

components/api/src/main/java/com/hotels/styx/api/service/spi/AbstractRegistry.java
```diff
@@ -43,6 +47,24 @@
 
     private final Announcer<Registry.ChangeListener> announcer = Announcer.to(ChangeListener.class);
     private final AtomicReference<Iterable<T>> snapshot = new AtomicReference<>(emptyList());
+    private final Predicate<Collection<T>> resourceConstraint;
+
+    /**
+     * Constructs an abstract registry with no resource constraint.
```
mikkokar: Should it be a "registry" constraint now when it is in `AbstractRegistry`?

Resources are specific to the file based providers, right?

kvosper: I will copy + paste my previous comment:
`AbstractRegistry` has a method `protected static <T extends Identifiable> Changes<T> changes(Iterable<T> newResources, Iterable<T> currentResources)`, so the term is already used. We can still rename them to something else though.
`registryConstraint` is not really informative though, that would be naming it after the object it lives in, rather than the objects it constrains.
If we rename "resources" throughout the code, I will rename the constraint accordingly.

mikkokar: Well, our `Resource` interface has methods like `path()`, `url()`, `inputStream()` etc which the `Identifiable`s stored in the Abstract registry do not have. So the question still remains what are the resources stored in the AbstractRegistry?

Maybe we can just decide that an identifiable stored in the registry in this instance can be called a resource and then we are sorted?

Another issue I have with this class is that it takes a type parameter T even if in practice it will never be anything else than a BackendService.

kvosper: We can't change the name unless we have a good alternative, so maybe let's leave that for later.

As for the type parameter, there is no point coupling this class to another class unnecessarily. Its purpose is to define a "Registry", so nothing else should be present.

mikkokar: Right. Well I will approve.

Issue #132, opened 2018-04-09
Remove fasterxml objects from the BackendService etc. classes


mikkokar:
## The problem
The BackendService API object exposes `com.fasterxml` annotations:

```
import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
```

Remove them from the interface.

## Detailed description
The yaml serialisation ought to move out of the API module. This can be implemented somewhere in styx-proxy module. We can use an intermediate form, perhaps a java bean based, to perform deserialisation (from yaml configuration) and serialisation (to admin interface).

## Acceptance criteria
- The `com.fasterxml.*` imports no longer appear in BackendService & related API classes.
- The admin interface remains unaffected.

kvosper:
A possible implementation is to replace the fasterxml annotations with new ones created inside the Styx codebase. This ought to make it easier to adapt to JSON (de)serialisation without needing to programmatically add every property into some translation layer.

PR #141, opened 2018-04-20
Removing fasterxml objects from BackendService

alobodzki:
Resolves issue:
Remove fasterxml objects from the BackendService etc. classes
https://github.com/HotelsDotCom/styx/issues/132


components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/BackendServiceDeserializer.java
```diff
@@ -0,0 +1,295 @@
...
+            return this;
+        }
+
+        /**
+         * Builds the application.
+         *
+         * @return the application
+         */
+        public BackendServiceDeserializer build() {
+            BackendService.Builder builder = BackendService.newBackendServiceBuilder();
```
kvosper: Builder interfaces are fluent. There is no need to repeat `builder.`
This code could be:
```
 public BackendServiceDeserializer build() {
            BackendService.Builder builder = BackendService.newBackendServiceBuilder()
                    .id(this.id)
                    .path(this.path)
                    .origins(this.origins)
                    .connectionPoolConfig(this.connectionPoolSettings)
                    .healthCheckConfig(this.healthCheckConfig)
                    .stickySessionConfig(this.stickySessionConfig)
                    .rewrites(this.rewrites)
                    .responseTimeoutMillis(this.responseTimeoutMillis)
                    .https(this.tlsSettings)
                    .httpsOld(this.tlsSettings);
            return new BackendServiceDeserializer(builder);
        }
```

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/BackendServiceDeserializer.java
```diff
@@ -0,0 +1,295 @@
...
+        private StickySessionConfig stickySessionConfig = stickySessionDisabled();
+        private HealthCheckConfig healthCheckConfig = noHealthCheck();
+        private List<RewriteConfig> rewrites = emptyList();
+        public int responseTimeoutMillis = DEFAULT_RESPONSE_TIMEOUT_MILLIS;
+        private TlsSettings tlsSettings;
+
+        public Builder() {
+        }
+
+        private Builder(BackendService backendService) {
```
kvosper: Unused private constructor.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/BackendServiceSerializer.java
```diff
@@ -0,0 +1,165 @@
...
+        return Optional.ofNullable(this.tlsSettings);
+    }
+
+    @JsonProperty("tlsSettings")
+    private TlsSettings getTlsSettings() {
+        return this.tlsSettings().orElse(null);
+    }
+
+    public BackendService.Protocol protocol() {
+        if (tlsSettings == null) {
```
kvosper: Using static-imports, this code could be `return tlsSettings == null ? HTTP : HTTPS;`

components/proxy/src/test/java/com/hotels/styx/admin/handlers/OriginsHandlerTest.java
```diff
@@ -87,6 +90,10 @@ public void respondsWithEmptyArrayWhenNoOrigins() throws IOException {
     }
 
     private static Iterable<BackendService> unmarshalApplications(String content) throws IOException {
-        return MAPPER.readValue(content, new TypeReference<Iterable<BackendService>>(){});
+        Iterable<BackendServiceDeserializer> backendServices = MAPPER.readValue(content, new TypeReference<Iterable<BackendServiceDeserializer>>() {
+        });
+        return StreamSupport.stream(backendServices.spliterator(), false)
```
kvosper: Let's statically-import `StreamSupport.stream` here too.

mikkokar:
Also looking into `BackendServices` class, there is a `JsonProperty` for "applications". This goes against the plan of modularising the backend service providers:

- The particular backend service storage format & how to read that in should be specific to a particular backend service provider implementing the SPI interface.

- The `JsonProperty` "applications" inside `BackendServices` is specific to the YAML based origins provider. 

Therefore the `BackendServices` class, if that still is necessary, should move to `proxy.backends.file` package. Perhaps it even should be a package private class if it is only used for reading the "applications" object from the yaml configuration file?

Please look into this issue.

components/proxy/src/main/java/com/hotels/styx/applications/yaml/YamlApplicationsProvider.java
```diff
@@ -32,14 +33,16 @@
...
 import static com.hotels.styx.client.applications.BackendServices.newBackendServices;
 import static java.lang.String.format;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.StreamSupport.stream;
 
 /**
  * Provides applications by reading from a YAML file.
  *
  */
 public class YamlApplicationsProvider implements ApplicationsProvider {
```
mikkokar: Apart from tests, is the origins starter app only place where this class is used from? 

As such, can you investigate if it can be removed?

PR #145, opened 2018-04-27
Improve warning message regarding idle persistent connections.

mikkokar:
The existing warning message when closing idle persistent connections is misleading. This PR provides more informative warning message.

Also, this PR adds the `keepAliveTimeoutMillis` configuration parameter in end-user documentation.


components/proxy/src/main/java/com/hotels/styx/proxy/ProxyConnectorFactory.java
```diff
@@ -214,7 +214,7 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
                     IdleStateEvent e = (IdleStateEvent) evt;
                     if (e.state() == ALL_IDLE && !httpTransactionOngoing) {
                         if (ctx.channel().isActive()) {
-                            LOGGER.warn("Closing suspected leaked connection={}", ctx.channel().remoteAddress());
+                            LOGGER.warn("Closing an idle persistent connection={}", ctx.channel().remoteAddress());
```
dvlato: Maybe change this to 'closing idle connection' so the message is never misleading.

PR #146, opened 2018-04-27
Fix for #142: Improve type detection for config validator.

### LINKED ISSUE #142: Config validator: Improve field type detection
## The problem

(1) The config schema validator relies on fasterxml `JsonNode` class to determine data types of the deserialised JSON objects. Sadly, fasterxml doesn't always get it correct, resulting perfectly valid JSON documents failing the validation. This has been seen with boolean fields, which erroneously have been mistaken for a string:

```
Unexpected field type. Field 'request-logging.inbound.enabled' should be BOOLEAN, but it is STRING
```

(2) Styx configuration supports placeholders. The placeholder substitution may change the apparent data type. For example an integer value may be mistaken for a string, invalidating just like in the above example for booleans.

## The solution

Whenever a String value is encountered, but a boolean is expected, the validator should attempt to parse the string value as a boolean. If successful, the field type is considered as boolean.

Whenever a STRING value is encountered, but an integer is expected, the validator should attempt to parse the string value as an integer. If successful, the field type is considered as integer.

## Acceptance criteria

(/) Accepts YAML boolean values as per http://yaml.org/type/bool.html.

(/) Placeholder replacement can yield integer and boolean values.

(/) When incorrect type is detected, print the actual field value from the deserialised yaml object into the error message. For example:

```
Unexpected field type. Field 'request-logging.inbound.enabled' should be BOOLEAN, but it is STRING (enabled='true')
```

mikkokar:
Fixes #142.  Config schema validator now accepts strings as integer/booleans when they parse into appropriate types.

That is, configurations like:

```
   myInt: "5"
```

and

```
  myBool: "true"
```

is now accepted as integer and boolean fields although the underlying JSON parser detects them as strings.

### DIFF
--- a/components/common/src/main/java/com/hotels/styx/config/validator/DocumentFormat.java
+++ b/components/common/src/main/java/com/hotels/styx/config/validator/DocumentFormat.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.hotels.styx.config.schema.InvalidSchemaException;
 import com.hotels.styx.config.schema.Schema;
+import com.hotels.styx.config.schema.Schema.FieldType;
 import com.hotels.styx.config.schema.SchemaValidationException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -30,7 +31,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
+import static java.lang.Integer.parseInt;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
@@ -46,6 +49,7 @@ import static java.util.Objects.requireNonNull;
  */
 public class DocumentFormat {
     private static final Logger LOGGER = LoggerFactory.getLogger(DocumentFormat.class);
+    private static final Pattern YAML_BOOLEAN_VALUES = Pattern.compile("(?i)true|false");
 
     private final Schema rootSchema;
     private final Map<String, Schema> schemas;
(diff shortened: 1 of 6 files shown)


components/common/src/main/java/com/hotels/styx/config/validator/DocumentFormat.java
```diff
@@ -222,15 +226,37 @@ private static String displayExpectedType(Schema.FieldType expectedType, Schema.
...
+        FieldType actualType = toFieldType(value);
+
+        if (FieldType.STRING.equals(actualType)) {
+            if (canParseAsInteger(value)) {
+                return;
+            }
+            if (canParseAsBoolean(value)) {
+                return;
+            }
+        }
```
kvosper:
```
        if (FieldType.STRING.equals(actualType) && (canParseAsInteger(value) || canParseAsBoolean(value))) {
            return;
        }
```

components/common/src/test/java/com/hotels/styx/config/validator/DocumentFormatTest.java
```diff
@@ -171,6 +170,23 @@ public void checksIntegerFieldTypes() throws Exception {
         assertThat(result, is(true));
     }
 
+    @Test
+    public void convertsStringsToIntegerValues() throws Exception {
+        JsonNode rootObject = YAML_MAPPER.readTree(
```
kvosper: You can get the IntelliJ IDEA code formatter to align your concatenated strings properly if you start the with an empty string on the initial line, like:
```
 JsonNode rootObject = YAML_MAPPER.readTree(""
                + "root: \n"
                + "  myInt: '5' \n");
```
It's a bit weird, but will save you from wrestling with the formatter or getting misaligned code by accident.

components/common/src/test/java/com/hotels/styx/config/validator/DocumentFormatTest.java
```diff
@@ -207,6 +223,34 @@ public void checksBoolFieldTypes() throws Exception {
         assertThat(result, is(true));
     }
 
+    @Test
+    public void convertsStringToBooleanValues() throws Exception {
+        JsonNode rootObject = YAML_MAPPER.readTree(
+                "root: \n"
```
kvosper: See comment above about alignment.

plugin-examples/src/main/java/com/hotels/styx/ExamplePlugin.java
```diff
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2018 Expedia Inc.
+  Copyright (C) ${license.git.copyrightYears} Expedia Inc.
```
kvosper: Where did this come from? Appears in 3 more classes below too.

components/common/src/test/java/com/hotels/styx/config/validator/DocumentFormatTest.java
mikkokar: Ok. I applied this technique to all other tests in this file too.

components/common/src/test/java/com/hotels/styx/config/validator/DocumentFormatTest.java
mikkokar: 👍

plugin-examples/src/main/java/com/hotels/styx/ExamplePlugin.java
mikkokar: There is something in `maven` that keeps automatically changing these 4 files every time I'll do a build from the command line. I haven't had time to investigate this issue. But obviously I should have backed off the unintended changes before publishing a PR.

Well these unintended changes have now been backed off.

PR #147, opened 2018-04-27
Remove fasterxml annotations from api with mixins#132

alobodzki:
Resolves issue:
Remove fasterxml objects from the BackendService etc. classes
https://github.com/HotelsDotCom/styx/issues/132


components/api/pom.xml
```diff
@@ -51,11 +51,6 @@
       <artifactId>annotations</artifactId>
     </dependency>
 
-    <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-databind</artifactId>
-    </dependency>
-
```
mikkokar: Well done! 🥇🥇🥇

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/json/ObjectMappers.java
```diff
@@ -0,0 +1,66 @@
...
+import com.hotels.styx.infrastructure.configuration.json.mixins.OriginsSnapshotMixin;
+import com.hotels.styx.infrastructure.configuration.json.mixins.RewriteConfigMixin;
+import com.hotels.styx.infrastructure.configuration.json.mixins.StickySessionConfigMixin;
+import com.hotels.styx.infrastructure.configuration.json.mixins.TlsSettingsMixin;
+
+
+/**
+ * Collection of methods that helps to work with {@link ObjectMapper} objects in styx libraries.
+ */
+public class ObjectMappers {
```
mikkokar: For an utility class like this, we would as per convention:

  1) Mark is a `final` to prevent subclassing, and
  2) Add a private constructor to prevent instantiation.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
```diff
@@ -191,7 +190,7 @@ private static String requireNonEmpty(String originsFile) {
         }
 
         private BackendServices readBackendServices(byte[] content) throws Exception {
-            return newBackendServices(delegate.read(content, new TypeReference<List<BackendService>>() {
+            return BackendServices.newBackendServices(delegate.read(content, new TypeReference<List<BackendService>>() {
```
mikkokar: Please use static import as per style conventions.

Issue #151, opened 2018-05-01
Graphite Reporter: The graphite IP address gets cached.


mikkokar:
## The problem

Graphite IP address is resolved when the GraphiteReporterService is created, and cached ever since then. Thus the IP address never re-resolves if it changes behind the DNS name.

mikkokar:
Fixed by PR: #153

PR #153, opened 2018-05-01
Prevent Graphite IP address from being cached.

mikkokar:
Fixes issue #151, where Graphite IP address got cached, therefore preventing DNS host name from re-resolving.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
+++ b/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
@@ -27,7 +27,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
 import static com.codahale.metrics.MetricFilter.ALL;
-import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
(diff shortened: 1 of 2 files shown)


mikkokar:
Oh. Looks like my local master is off sync. I will address this now.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
```diff
@@ -41,16 +41,18 @@
...
+        this.host = requireNonNull(builder.host);
+        this.port = builder.port;
+        MetricRegistry registry = requireNonNull(builder.registry);
         GraphiteSender graphiteSender = Optional
                 .ofNullable(builder.graphiteSender)
-                .orElseGet(() -> new NonSanitizingGraphite(address));
+                .orElseGet(() -> new NonSanitizingGraphite(host, port));
 
-        String prefix = checkNotNull(builder.prefix);
+        String prefix = requireNonNull(builder.prefix);
```
dvlato: Are we following any pattern to decide when to use requireNonNull instead of checkNotNull?

kvosper: IIRC, `requireNonNull` was added in Java 8. The Styx project was began development in Java 7, so we were using Guava's `checkNotNull`. Since upgrading there is no longer a reason to use this third-party dependency when the functionality is the same.

mikkokar: Prior to Java 8 we were using `checkNotNull` from Guava. But since then we should use `requireNonNull` in its place.

PR #155, opened 2018-05-02
Fixes #152: GraphiteReporter retries connection to the server too often.

### LINKED ISSUE #152: Graphite Reporter: retries too often
## The problem

Graphite reporter retries are too aggressive. 

When the Styx GraphiteReporter fails to report the metrics due to an error (IO-error), it tries again resuming from the point it failed. This however can cause a storm of retries possibly overwhelming the Graphite Server.

The graphite reporter needs to limit the number of retries towards the Graphite server.
Additionally, it is worth nothing that keeping a permanent connection to the server would prevent us from distributing the load among different instances of the Carbon/graphite server using DNS round robin.

## Acceptance criteria
- Limit the number of retries 
- Logs an error message for failed attempts (it already does)
- Allow to close the connection after every 'batch of metrics' so the load can be distributed as expected.

dvlato:
-  Connection will be retried for a fixed number of times (5) before aborting the reporting process. 
 -  Connection will be opened/closed in each scheduled reporting process.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
+++ b/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
@@ -36,14 +36,12 @@ import java.util.concurrent.TimeUnit;
 
 /**
  * A reporter which publishes metric values to a Graphite server.
- * <p>
- * This reporter is <b>not</b> thread safe and is intended for use within a single thread.
  *
  * @see <a href="http://graphite.wikidot.com/">Graphite - Scalable Realtime Graphing</a>
  */
 public class GraphiteReporter extends ScheduledReporter {
     /**
-     * Returns a new {@link Builder} fGor {@link GraphiteReporter}.
+     * Returns a new {@link Builder} for {@link GraphiteReporter}.
      *
      * @param registry the registry to report
      * @return a {@link Builder} instance for a {@link GraphiteReporter}
@@ -179,14 +177,14 @@ public class GraphiteReporter extends ScheduledReporter {
 
     private final LoadingCache<String, MeteredPrefixes> meteredPrefixes = CacheBuilder.newBuilder().build(new CacheLoader<String, MeteredPrefixes>() {
         @Override
-        public MeteredPrefixes load(String name) throws Exception {
+        public MeteredPrefixes load(String name) {
             return new MeteredPrefixes(name);
         }
     });
 
     private final LoadingCache<String, TimerPrefixes> timerPrefixes = CacheBuilder.newBuilder().build(new CacheLoader<String, TimerPrefixes>() {
         @Override
-        public TimerPrefixes load(String name) throws Exception {
+        public TimerPrefixes load(String name) {
             return new TimerPrefixes(name);
         }
     });
--- a/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
+++ b/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
@@ -14,6 +14,7 @@ import com.codahale.metrics.Snapshot;
 import com.codahale.metrics.Timer;
 import com.codahale.metrics.graphite.GraphiteSender;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import org.slf4j.Logger;
@@ -26,7 +27,6 @@ import java.util.concurrent.TimeUnit;
 import static com.codahale.metrics.Clock.defaultClock;
 import static com.codahale.metrics.MetricFilter.ALL;
 import static com.codahale.metrics.MetricRegistry.name;
-import static com.google.common.cache.CacheBuilder.newBuilder;
 import static com.hotels.styx.metrics.reporting.graphite.IoRetry.tryTimes;
 import static java.util.Locale.US;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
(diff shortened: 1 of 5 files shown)


components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
```diff
@@ -348,12 +360,18 @@ private void doSend(String name, double value, long timestamp) {
...
+        attemptWithRetryAndReconect(
+                () -> graphite.send(name, value, timestamp));
+    }
+
+    private void attemptWithRetryAndReconect(IoRetry.IoRunnable operation) {
+        IoRetry.tryTimes(
+                MAX_RETRIES,
+                operation,
+                this::reconnectIfNecessary
+        );
```
mikkokar: It looks to me that this approach will not address the issue. This is because `doSend` and `attemptWithRetryAndReconnect` are called once for each metric, resetting the retry count to `MAX_RETRIES` for each metric. As a result it will end up attempting each metric `MAX_RETRIES` times. 

Instead the `MAX_RETRIES` should be set to each reporting cycle such that the report function tolerates maximum of `MAX_RETRIES` before giving up.

mikkokar:
The GraphiteReporter class seems bit difficult to follow. Part of the reason is that the stateful behaviour like retrying graphite connections is hidden inside helper methods like `doReport`, `reportGauge`, and so on. So looking at the top level of `report()` it is impossible to say anything about the retry behaviour.

This obviously is an existing issue with this class and nothing to do with this PR. However, as a part of this PR we should take this opportunity to rectify these issues.

So could we instead refactor the `report()` along the lines of:

```
            Stream<? extends Metric> metrics =
                    Stream.of(
                            gauges.values().stream(),
                            counters.values().stream(),
                            histograms.values().stream(),
                            meters.values().stream(),
                            timers.values().stream()
                    ).flatMap(x -> x);
            
            retries = 0;
            metrics.forEach(metric -> {
                try {
                    if (retries < MAX_RETRIES) {
                        reportMetric(metric);
                    }
                } catch (UncheckedIOException e) {
                    retries++;
                }
            });
```

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/IoRetry.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.function.Consumer;
+
+import static java.lang.String.format;
+
+class IoRetry {
```
kvosper: Utility classes should be `final`, with a `private` constructor.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/IoRetry.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @FunctionalInterface
+    public interface IoRunnable {
```
kvosper: Seems to duplicate the `GraphiteReporter.IOAction` interface. Perhaps that could be raised to package-private or something?

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/IoRetry.java
```diff
@@ -0,0 +1,58 @@
...
+        void run() throws IOException;
+    }
+
+    private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(GraphiteReporter.class);
```
kvosper: `org.slf4j.Logger` can be imported instead of written fully-qualified.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/IoRetry.java
```diff
@@ -0,0 +1,58 @@
...
+    public static void tryTimes(int times, IoRunnable task, Consumer<Exception> onError)
+            throws UncheckedIOException {
+        int retries = 0;
+
+        while (true) {
+            try {
+                task.run();
+                return;
+            } catch (IOException e) {
+                try {
```
kvosper: I suspect there may be a way to write this that is a little more readable, with fewer nested blocks. I will try to come back with a more specific suggestion.

components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/IoRetryTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static org.mockito.Mockito.verify;
+import static org.testng.Assert.fail;
+
+
+public class IoRetryTest {
+
+    private IoRunnable task;
+
+    @BeforeMethod
+    public void initialize() {
```
kvosper: I think `createMock` would be a better method name.

components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/IoRetryTest.java
```diff
@@ -0,0 +1,84 @@
...
+    public void retryUpToMaxTimesForIoException() throws IOException {
+        doThrow(new IOException()).when(task).run();
+        try {
+            tryTimes(3, task, (e) -> assertThat(e, is(instanceOf(IOException.class))));
+            fail("An exception should have been thrown");
+        } catch (UncheckedIOException e) {
+            assertThat(e.getMessage(), containsString("Operation failed after"));
+        }
+        verify(task, times(3)).run();
+    }
```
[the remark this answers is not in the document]
dvlato: I have implemented this as requested. But I believe the request can be linked to an issue we have in our code base,  fragility and duplication in our tests.

dvlato:
I think I have somehow addressed  all your comments. Even though it's true that the class could be greatly improved by handling all the different meters in an uniform way, the problem I see in the current (previous to this PR) implementation and Mikko's code sample is that we are linking the I/O errors with the metric being reported and skipping a metric just because there was a connectivity issue (which will affect all the other metrics anyway).   Even if the code could be easily changed to do the retry logic  around the 'reportMetric' , it would still be mixing different concerns -what we really want to retry is the I/O operation itself and not any other logic.  I would rather completely decouple the I/o operations from the rest of the code.

Since this was an urgent issue to fix, maybe we can address that in a different issue altogether (I think the whole class and test can be completely refactored).

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/IoRetry.java
```diff
@@ -0,0 +1,87 @@
...
+
+import static java.lang.String.format;
+
+/**
+ * Generic retry class for tasks that perform I/O.
+ */
+final class IoRetry {
+
+    @FunctionalInterface
+    public interface IoRunnable {
```
kvosper: Our convention is to put inner classes/interfaces at the bottom of the class. 
Alternatively, did you consider my suggestion of replacing this and `GraphiteReporter.IOAction` with an interface in its own file?

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
```diff
@@ -236,17 +242,23 @@ public void report(SortedMap<String, Gauge> gauges,
...
-            reconnectIfNecessary(e);
         }
     }
 
+    @VisibleForTesting
+    void initConnection() {
+        IoRetry.tryTimes(
+                MAX_RETRIES,
+                graphite::connect,
+                (e) -> attemptErrorHandling(graphite::close)
```
kvosper: Single lambda parameters don't need to be wrapped in parentheses.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporter.java
```diff
@@ -236,17 +242,23 @@ public void report(SortedMap<String, Gauge> gauges,
 
     private <M extends Metric> void doReport(String name, M metric, long timestamp, MetricReportingAction<M> consumer) {
```
kvosper: This method can be static now.

components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterTest.java
```diff
@@ -143,8 +147,21 @@ public void continuesReportingMetricsAfterOneThrowsAnException() throws IOExcept
...
 
+
     @Test
-    public void continuesToSendAfterIOException() throws IOException {
+    public void initConnectionRetriesOnFailure() throws Exception {
+
+        doThrow(new UnknownHostException("UNKNOWN-HOST")).when(graphite).connect();
+        try {
+            reporter.initConnection();
+        } catch (UncheckedIOException e) {
```
kvosper: Empty `catch` blocks should have a comment explaining why.

components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/IoRetryTest.java
```diff
@@ -0,0 +1,95 @@
...
+        verify(task, times(2)).run();
+    }
+
+    @Test
+    public void noRetriesForOtherExceptions() throws IOException {
+        doThrow(new NullPointerException()).doNothing().when(task).run();
+        try {
+            tryTimes(3, task, null);
+            fail("An exception should have been thrown");
+        } catch (NullPointerException e) {
```
kvosper: Empty `catch` blocks should have a comment explaining why.

components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/IoRetryTest.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+
+    @Test
+    public void illegalArgumentExceptionAndNoAttemptsForInvalidArgument() throws IOException {
+        doNothing().when(task).run();
+        try {
+            tryTimes(0, task, null);
+            fail("An exception should have been thrown");
+        } catch (IllegalArgumentException e) {
```
kvosper: Empty `catch` blocks should have a comment explaining why.

PR #156, opened 2018-05-03
Mock dns server

mikkokar:
## Summary

Raising this as a PR to demonstrate a spike that implements and integrates a mock DNS server for Styx unit testing. The Mock server and the related concepts are rather simple, but the integration into the rest of the build system has proven problematic. @dvlato feel free to have a go at it.

* The test well when run from IntellJ or directly from the `styx-proxy` Maven subproject. 

* But the tests will fail when the styx is built from top level root project. This is because the mock-server doesn't get registered as a DNS provider.

## Details

This PR implements a simple `NameService` implementation that is loaded into the JVM alongside of its own DNS name service. Our name service is simply called `MockNameService`. It allows one to register a mock delegate for intercepting name service queries originating from the Java libraries. Styx unit tests can then perform verifications against the registered delegate.

### Service Loading

The MockNameService is loaded using the Java `ServiceLoader` infrastructure as described in https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html. To sum up, the service loading consist of:

* `MockNameService` implements the `NameService` interface.
* `LocalNameServiceDescriptor.java` that is used by Java to load the `MockNameService` class.
* `sun.net.spi.nameservice.NameServiceDescriptor` which declares the name service descriptor class so that Java service loader can find it.

### Enabling the Mocked Name Service

The mock name service is installed alongside Java's default name service implementation. We need to ensure that our implementation takes the priority over the Java default service. For this, we will set the system properties as follows:

```
             -Dsun.net.spi.nameservice.provider.1=dns,local-dns
             -Dsun.net.spi.nameservice.provider.2=default
```

The system properties are configured at the surefire plugin which runs the tests under Maven.

With this configuration Java first attempts to use "local-dns". If it fails to resolve the address, indicated by `UnknownHostException`, the next one (default) will be attempted. This PR takes advantage of this behaviour by deliberately falling back to default Java name service when mocking is not needed. 

### Disable DNS Caching

We also need to disable the DNS cache to ensure Java is consistently performing DNS lookups and our assertions against the mocked DNS service are predictable. The caching is disabled in the surefire plugin configuration:

```
             -Dsun.net.inetaddr.ttl=0
             -Dsun.net.inetaddr.negative.ttl=0
```

### Concurrency Issues

There is only one DNS service in the system. Thus any potentially concurrent usage must be serialised. We do this by modifying the Surefire `forkCount` attribute. This enables tests to run concurrently at the JVM level, rather than as threads inside the JVM.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
+++ b/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
@@ -59,7 +59,7 @@ public final class GraphiteReporterService extends AbstractStyxService {
     protected CompletableFuture<Void> startService() {
         return CompletableFuture.runAsync(() -> {
             this.reporter.start(reportingIntervalMillis, MILLISECONDS);
-            LOGGER.info("Graphite service started name=\"{}\"", serviceName());
+            LOGGER.info("Graphite service started, service name=\"{}\"", serviceName());
         });
     }
 
@@ -67,7 +67,7 @@ public final class GraphiteReporterService extends AbstractStyxService {
     protected CompletableFuture<Void> stopService() {
         return CompletableFuture.runAsync(() -> {
             this.reporter.stop();
-            LOGGER.info("Graphite service stopped name=\"{}\"", serviceName());
+            LOGGER.info("Graphite service stopped, service name=\"{}\"", serviceName());
         });
     }
 
--- a/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
+++ b/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
@@ -59,7 +59,7 @@ public final class GraphiteReporterService extends AbstractStyxService {
     protected CompletableFuture<Void> startService() {
         return CompletableFuture.runAsync(() -> {
             this.reporter.start(reportingIntervalMillis, MILLISECONDS);
-            LOGGER.info("Graphite service started name=\"{}\"", name);
+            LOGGER.info("Graphite service started name=\"{}\"", serviceName());
         });
     }
 
@@ -67,7 +67,7 @@ public final class GraphiteReporterService extends AbstractStyxService {
     protected CompletableFuture<Void> stopService() {
         return CompletableFuture.runAsync(() -> {
             this.reporter.stop();
-            LOGGER.info("Graphite service stopped name=\"{}\"", name);
+            LOGGER.info("Graphite service stopped name=\"{}\"", serviceName());
         });
     }
 
--- a/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
+++ b/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
@@ -16,13 +16,11 @@
 package com.hotels.styx.metrics.reporting.graphite;
 
 import com.codahale.metrics.MetricRegistry;
-import com.codahale.metrics.graphite.Graphite;
 import com.codahale.metrics.graphite.GraphiteSender;
 import com.google.common.annotations.VisibleForTesting;
-import com.hotels.styx.api.service.spi.StyxService;
+import com.hotels.styx.api.service.spi.AbstractStyxService;
 import org.slf4j.Logger;
 
-import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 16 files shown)


pom.xml
```diff
@@ -982,21 +983,6 @@
...
-                ${argLine}
-              </argLine>
-              <excludes>
-                <exclude>**/Abstract*</exclude>
-                <exclude>**/*$*</exclude>
-              </excludes>
-            </configuration>
-          </plugin>
           <plugin>
             <artifactId>maven-failsafe-plugin</artifactId>
```
dvlato: I know it's a bit unrelated, but do you think it's a good idea to remove the failsafe configuration as well?

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
```diff
@@ -63,66 +56,53 @@ private GraphiteReporterService(Builder builder) {
     }
 
     @Override
-    public CompletableFuture<Void> start() {
+    protected CompletableFuture<Void> startService() {
         return CompletableFuture.runAsync(() -> {
-            LOGGER.info("Graphite started on address=\"{}:{}\"", host, port);
             this.reporter.start(reportingIntervalMillis, MILLISECONDS);
+            LOGGER.info("Graphite service started name=\"{}\"", name);
```
kvosper: Maybe `LOGGER.info("Graphite service started: serviceName=\"{}\"", name);`

support/testsupport/src/main/java/com/hotels/styx/support/dns/MockNameService.java
```diff
@@ -0,0 +1,79 @@
...
+        LOGGER.info("Unsetting delegate");
+        this.delegate.set(null);
+    }
+
+    @Override
+    public InetAddress[] lookupAllHostAddr(String hostName) throws UnknownHostException {
+
+        if (delegate.get() != null) {
+            InetAddress[] result = delegate.get().lookupAllHostAddr(hostName);
+            LOGGER.info("lookup addresses for host: '{}' results={}", hostName, result.length);
```
dvlato: Probably we want to verify 'result' here and if it's null, we throw an unknown host exception. This way the 'default' nameServer will be called.  Otherwise we will get a nullpointerexception in the logger statement.

support/testsupport/src/main/java/com/hotels/styx/support/dns/LocalNameServiceDescriptor.java
```diff
@@ -0,0 +1,58 @@
...
+ * NOTE: This class is not suited for parallel tests.
+ */
+public class LocalNameServiceDescriptor implements NameServiceDescriptor {
+
+    private static final String dnsName = "local-dns";
+    private static Optional<MockNameService> nameService = Optional.empty();
+
+    @Override
+    public NameService createNameService() {
+        synchronized (this) {
```
kvosper: Why is there synchronization and Optional for a method that always returns `MockNameService.SELF`?

support/testsupport/src/main/java/com/hotels/styx/support/dns/LocalNameServiceDescriptor.java
```diff
@@ -0,0 +1,58 @@
...
+  limitations under the License.
+ */
+package com.hotels.styx.support.dns;
+
+import sun.net.spi.nameservice.NameService;
+import sun.net.spi.nameservice.NameServiceDescriptor;
+
+import java.util.Optional;
+
+/*
```
dvlato: Maybe we can document this in the Javadoc? I was trying to find out what are the recommendations for Javadoc docs regarding concurrency but I haven't found anything in a very very quick search... We could also want to use annotations such as "@NotThreadSafe".

support/testsupport/src/main/java/com/hotels/styx/support/dns/MockNameService.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public void unset() {
+        LOGGER.info("Unsetting delegate");
+        this.delegate.set(null);
+    }
+
+    @Override
+    public InetAddress[] lookupAllHostAddr(String hostName) throws UnknownHostException {
+
+        if (delegate.get() != null) {
```
kvosper: You should assign the return value of `delegate.get()` to a local variable and re-use it in order to keep the operation atomic.
Same comment for method below.

support/testsupport/src/main/java/com/hotels/styx/support/dns/LocalNameServiceDescriptor.java
mikkokar: I think the synchronisation can be removed.

As for the optional, I think I can refactor it away.

support/testsupport/src/main/java/com/hotels/styx/support/dns/MockNameService.java
mikkokar: ok

support/testsupport/src/main/java/com/hotels/styx/support/dns/LocalNameServiceDescriptor.java
mikkokar: Hmm... I assume you are referring to the `MockNameService` class?

I think the `LocalNameServiceDescriptor` itself is thread safe, but the consumers of the `MockNameService` should follow certain usage patterns.

I will add them in.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterService.java
mikkokar: I disagree. These messages are for our end users. They are not interested in seeing styx variable names, or log messages in the camel-case notation.

kvosper: I was more concerned with potential confusion when reading. 
It seems like the word "name" just suddenly appears, with no connection to the previous clause. Separating it with some kind of punctuation mark might do a lot to improve this.

mikkokar: I see. Makes sense. I will change this.

pom.xml
mikkokar: I would have to go through the pom files to ensure it is OK.

Therefore I would do this separately.

PR #158, opened 2018-05-14
Dont show health check if absent

### LINKED ISSUE #122: Admin page shows health check configuration even when it is absent
## The problem

The admin page for origins configuration (`/admin/configuration/origins`) shows the health check configuration block even when it is disabled:

```
  "healthCheck" : {
    "uri" : null,
    "intervalMillis" : 5000,
    "timeoutMillis" : 2000,
    "healthyThreshold" : 2,
    "unhealthyThreshold" : 2
  },
```

Well ok, it shows that `uri` is `null`, but really it shouldn't show this block at all.

## Acceptance criteria
- When health check configuration is absent, it shouldn't show up in the origins configuration page at all.

kvosper:
Fixes  #122


dvlato:
I would rather use a custom serializer for the configuration object so we can show:

 "healthCheck" : {  
    "enabled": false
}
or something similar and avoid having to deal with nulls.

Another option is that the config object itself has a different sutrcture when enabled = true than when it is disabled.

kvosper:
@dvlato Your comment seems to be a criticism of the original requirements in #122 rather than a review of this implementation - it might be better to discuss it on the issue page itself and see what @mikkokar says.

PR #159, opened 2018-05-14
Improve log messages when origin reload fails.

mikkokar:
More specifically, this PR adds a meaningful error message when origin reload fails due to multiple origins sharing the same path prefix.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
+++ b/components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
@@ -103,7 +103,11 @@ public class FileBackedRegistry<T extends Identifiable> extends AbstractRegistry
                         return reloaded(format("%s, File reloaded.", logPrefix));
                     }
                 } catch (Exception e) {
-                    LOG.error("Not reloading {} as there was an error reading content", configurationFile.absolutePath(), e);
+                    // Eliminate unhelpful error message when resource constraint exception fails.
+                    // This is a candidate for further refactoring.
+                    if (!"Resource constraint failure".equals(e.getMessage())) {
+                        LOG.error("Not reloading {} as there was an error reading content", configurationFile.absolutePath(), e);
+                    }
                     return failed(format("%s, Reload failure.", logPrefix), e);
                 }
             }
--- a/components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
+++ b/components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
@@ -36,7 +36,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 import static com.google.common.base.Objects.toStringHelper;
 import static com.google.common.base.Throwables.propagate;
@@ -47,6 +46,7 @@ import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toList;
 
 /**
  * File backed {@link BackendService} registry.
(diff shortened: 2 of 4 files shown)


components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
```diff
@@ -103,7 +103,11 @@ public String fileName() {
                         return reloaded(format("%s, File reloaded.", logPrefix));
                     }
                 } catch (Exception e) {
-                    LOG.error("Not reloading {} as there was an error reading content", configurationFile.absolutePath(), e);
+                    // Eliminate unhelpful error message when resource constraint exception fails.
+                    // This is a candidate for further refactoring.
+                    if (!"Resource constraint failure".equals(e.getMessage())) {
```
kvosper: Relying on this message not to change introduces an undesirable coupling. I think the ideal solution is to change the `AbstractRegistry` to throw a specific class, which we can then check for.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
```diff
@@ -200,11 +202,34 @@ private BackendServices readBackendServices(byte[] content) throws Exception {
...
-            return backendServices.stream()
-                    .collect(groupingBy(BackendService::path, counting()))
-                    .values()
+            Map<String, List<BackendService>> pathToApp = backendServices.stream()
+                    .collect(groupingBy(BackendService::path));
+
+            List<List<BackendService>> duplicateApps = pathToApp.values()
                     .stream()
-                    .noneMatch(count -> count > 1);
+                    .filter(bss -> bss.size() > 1)
```
kvosper: `bss` is not a meaningful variable name.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
```diff
@@ -200,11 +202,34 @@ private BackendServices readBackendServices(byte[] content) throws Exception {
...
+        private static String errorMessage(List<List<BackendService>> backendServices) {
+            return backendServices.stream()
+                    .map(RejectDuplicatePaths::duplicatePathsErrorForPath)
+                    .collect(joining(", "));
+        }
+
+        private static String duplicatePathsErrorForPath(List<BackendService> apps) {
+            String path = apps.get(0).path();
+            String appsList = apps.stream()
+                    .map(app -> format("'%s'", app.id().toString()))
```
kvosper: The `toString()` is redundant.

components/proxy/src/main/java/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistry.java
```diff
@@ -200,11 +202,34 @@ private BackendServices readBackendServices(byte[] content) throws Exception {
...
-                    .collect(groupingBy(BackendService::path, counting()))
-                    .values()
+            Map<String, List<BackendService>> pathToApp = backendServices.stream()
+                    .collect(groupingBy(BackendService::path));
+
+            List<List<BackendService>> duplicateApps = pathToApp.values()
                     .stream()
-                    .noneMatch(count -> count > 1);
+                    .filter(bss -> bss.size() > 1)
+                    .collect(Collectors.toList());
```
kvosper: Please statically import `toList`.

components/proxy/src/main/java/com/hotels/styx/infrastructure/FileBackedRegistry.java
mikkokar: I specifically wanted to avoid adding a new exception class as it would have to be a new public class inside the Styx API module. This could potentially open up backwards compatibility issues.

The error handling for the styx registry/service startup classes, not just this PR, ought to be redesigned anyway. In my opinion we ought to look into introducing a monadic error handling into these classes, but that is another story.

My PR is supposed top be a quick work-around for a particular problem. I don't have time at the moment to refactor this much further. Therefore I left a comment in the source code. But we should definitely discuss this going forward.

dvlato: Mikko and myself talked about this solution being quite hacky but agreed it could be ok for the time being.  Would you find using a constant in both places  instead of a String literal acceptable enough?  

Another solution would be  creating an Exception for this case that is a subclass of IllegalStateException so we don't have compatibility issues, but you might argue it's a bit of a misuse the inheritance system.

kvosper: Yes I would agree with using a constant.

PR #162, opened 2018-05-14
Refactor yamlreader

kvosper:
YamlReader is only used in origins file loading, but it has things related to config files, like overrides and placeholders. The required logic is actually much simpler, so the class has been deleted and the relevant code moved to its call site.


kvosper:
Accidentally branched from another branch.

PR #163, opened 2018-05-14
Refactor YamlReader

kvosper:
YamlReader is only used in origins file loading, but it has things related to config files, like overrides and placeholders. The required logic is actually much simpler, so the class has been deleted and the relevant code moved to its call site.


components/proxy/src/main/java/com/hotels/styx/infrastructure/YamlReader.java
```diff
@@ -1,151 +0,0 @@
...
-        this(emptyMap());
-    }
-
-    public YamlReader(Map<String, String> overrides) {
-        this.overrides = checkNotNull(overrides);
-    }
-
-    public T read(byte[] content, TypeReference<T> typeReference) throws Exception {
-        JsonNode rootNode = MAPPER.readTree(content);
-        List<String> includePaths = includedPaths(rootNode);
```
mikkokar: Do you know this `includePaths` handing does? I don't see that having been moved down tho the call site.

kvosper: In practice, it does nothing. The origins file is supposed to be a YAML representation of a `List` of `BackendService`s. If an `include` property were added, it would change the structure and no longer be a valid list.

PR #167, opened 2018-05-18
Cookie names are case sensitive.

taer:
This makes response.removeCookie() match request.removeCookie()

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
+++ b/components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
@@ -516,7 +516,7 @@ public final class HttpResponse implements HttpMessage {
          */
         public Builder removeCookie(String name) {
             cookies.stream()
-                    .filter(cookie -> cookie.name().equalsIgnoreCase(name))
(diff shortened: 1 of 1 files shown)


components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -516,7 +516,7 @@ public Builder addCookie(String name, String value) {
          */
         public Builder removeCookie(String name) {
             cookies.stream()
-                    .filter(cookie -> cookie.name().equalsIgnoreCase(name))
+                    .filter(cookie -> cookie.name().equals(name))
```
mikkokar: Now when you are at it, would you mind also fixing any other instances where the `equalsIgnoreCase` is used for cookie names?

I think there would be in the `HttpResponse`, `HttpRequest`, `FullHttpRequest`, `FullHttpResponse`, `StreamingHttpRequest`, and `StreamingHttpResponse` classes.

dvlato:
PR closed. New PR in https://github.com/HotelsDotCom/styx/pull/186

PR #172, opened 2018-05-25
Updated example plugin to 1.0 API

dvlato:


plugin-examples/src/main/java/com/hotels/styx/ExamplePluginConfig.java
```diff
@@ -34,8 +33,8 @@
      * @param responseHeaderValue
      */
     public ExamplePluginConfig(
-            @JsonProperty("requestHeaderValue") String requestHeaderValue,
-            @JsonProperty("responseHeaderValue") String responseHeaderValue) {
+            String requestHeaderValue,
+            String responseHeaderValue) {
```
mikkokar: This could break the plugin configuration. I doubt we have a test for the plugin config to verify this.

PR #174, opened 2018-05-25
Add proxy server start endpoint to admin interface, plus eventnexus

kvosper:
Creates an endpoint that can be queried to see if the proxy server has started.


components/proxy/src/main/java/com/hotels/styx/Environment.java
```diff
@@ -33,12 +35,14 @@
 public final class Environment implements com.hotels.styx.api.Environment {
     private final Version version;
     private final EventBus eventBus;
+    private final EventNexus eventNexus;
```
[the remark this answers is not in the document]
kvosper: It's definitely *not* necessary to have both. In fact the `EventNexus` can replace a lot of our current code.
In this PR I only used it to add new functionality in order to avoid making the PR too large.

There are three alternatives:
* Replace `EventBus` with `EventNexus` in a separate PR, also create PRs for other places it can be used
* Add more changes to this PR, either just to replace `EventBus` or every possible change
* Cancel this PR, create a separate PR to introduce `EventNexus` and change current code to use it, then do the proxy-start-endpoint feature as a new PR after that one is merged.

What do you think is the best approach?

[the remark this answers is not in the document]
mikkokar: Our specific requirement is to implement an admin interface endpoint for probing the proxy server status. That is the first priority.

You don't need to cancel this PR. Just back off the event nexus changes, and create a separate branch for them.

Regardless, it was good to see that code in this PR as it gives us some thoughts and direction to work on.

components/common/src/test/java/com/hotels/styx/events/SimpleEventNexusTest.java
```diff
@@ -0,0 +1,124 @@
...
+        assertThat(SimpleEventNexus.equalOrChild("foo", "foo"), is(true));
+        assertThat(SimpleEventNexus.equalOrChild("foo.bar", "foo"), is(true));
+        assertThat(SimpleEventNexus.equalOrChild("foo.bar", "foo.bar"), is(true));
+        assertThat(SimpleEventNexus.equalOrChild("foo.bar.baz", "foo.bar"), is(true));
+
+        assertThat(SimpleEventNexus.equalOrChild("foobar", "foo"), is(false));
+    }
+
+    @Test
+    public void eventNamesAreCaseInsensitive() {
```
mikkokar: Is there a specific reason why this needs to be case insensitive? I think it could be sensitive to case.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
```diff
@@ -0,0 +1,52 @@
...
+    private volatile boolean proxyServerStarted;
+
+    public ProxyStatusHandler(EventNexus eventNexus) {
+        eventNexus.events("server.started.proxy")
+                .map(Event::value)
+                .cast(Boolean.class)
+                .subscribe(started -> {
+                    proxyServerStarted = started;
+                    LOGGER.info("proxyServerStarted={}", started);
+                });
```
mikkokar: ProxyStatusHandler only needs to see the the proxy server state, which from `ProxyStatusHandler` point of view is just a read-only value.

Using an event bus for this use case is a bad practice due to:

1. It is a complicated way to communicate `proxyServerStarted` value to the handler.
2. The state management for "server.started.proxy" property is distributed to all listeners, instead of centralised for a single provider. This results in duplicated code: Every listener needs to implement the same 11 or 12 lines of code.
3. Opens up a possibility for data races. Consider what happens if server published the event "server.started.proxy" *before* some of the listeners was ready. That listener would miss the notification.

For these reasons, I would propose an opposing approach. Introduce a centralised configuration state variable for "server.started.proxy". Manage its state from the proxy server. Make it globally readable (read-only) via environment.

To clarify. I agree there definitely is a place for event based notifications within styx, but this endpoint is not one of them.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
```diff
@@ -0,0 +1,52 @@
...
+    }
```
mikkokar: Another issue with "server.started.proxy". Styx is capable of starting HTTP and HTTPS ports separately. There could be a possibility, or at least we need to consider a possibility, of them starting somewhat off-sync. Therefore we perhaps need to think about exposing them separately via this admin interface.

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -123,6 +124,7 @@ public HttpServer build() {
         httpRouter.add("/admin/origins/status", new OriginsInventoryHandler(environment.eventBus()));
         httpRouter.add("/admin/configuration/logging", new LoggingConfigurationHandler(styxConfig.startupConfig().logConfigLocation()));
         httpRouter.add("/admin/configuration/startup", new StartupConfigHandler(styxConfig.startupConfig()));
+        httpRouter.add("/admin/proxystatus", new ProxyStatusHandler(environment.eventNexus()));
```
mikkokar: I would consider adding more structure in the admin interface endpoint names. I would call this "/admin/proxy/server/status" etc.

Where:
- `/proxy` refers to the styx proxy application,
- `/proxy/server` refers to the `server` attribute inside proxy application,
- `/proxy/server/status` refers to the status attribute of the proxy server.

and so on.

@dvlato what do you think? I believe you are the most experienced amongst us in the web API design?

components/proxy/src/main/java/com/hotels/styx/Environment.java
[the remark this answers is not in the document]
kvosper: The Event Nexus provides the implementation, if it is backed off, the functionality will not work. Are you proposing an alternative implementation?

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
dvlato: I agree that feeding the endpoint directly with the events doesn't seem correct here.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
kvosper: The code is only considering the proxy started when *everything* is ready. I don't think there is a use-case for an endpoint saying when Styx is partially ready.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
kvosper: @mikkokar I understand your last two points in your list of three, but regarding your first, it is not clear - should we delay this feature until we have the "centralised configuration state variable[s]"? Otherwise it would require either: (1) using the existing `EventBus` which would seem to have the same issues, (2) implementing through some ad-hoc method which might be messy.

components/proxy/src/main/java/com/hotels/styx/Environment.java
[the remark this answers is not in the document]
dvlato: I image you have implemented EventNexus since it was necessary for this functionality. We might have to discuss that case... Otherwise I agree with mikko that replacing EventBus with the new class could be a separate PR.

In any case, as discussed below, it doesn't seem that that feeding the admin endpoint directly with the events is the best solution here.  At the very least, we'd need to make sure we  are not going to lose any events, but it seems that a 'centralised registry' could be a better option.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
mikkokar: Possibly not.

components/proxy/src/main/java/com/hotels/styx/Environment.java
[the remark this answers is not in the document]
mikkokar: Yeah, for this use case, a "centralised registry" is perfectly fine.

However for the general use case the consumers of the "centralised registry" may want to get notified of the state changes. So that is where & when I would bring back the event nexus.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
mikkokar: Going forward (not necessary for this change) the `StyxConfigStore` will need to provide notification services along the lines of:

```
   styxConfg.watch("server.started.proxy")
       .subscribe(value -> { ... });
```

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
dvlato: It sounds fine for me. I don't know if the status belongs only to the 'server attribute'  or we can consider it as the whole status for the whole 'proxy (server)'. I mean, would we have proxy/<something_different_from_server>/status ? And, in that case, do we need to aggregate the status to find out what's the global status of the proxy?

docs/user-guide/basic-usage.md
```diff
@@ -76,7 +76,15 @@ about the port numbers it is listening on.
 
 This tells you that the application has successfully started, and that proxy server is listening ports 8080 for HTTP, 
 8443 for HTTPS, and that admin interface has started on port 9000 (HTTPS).
+ 
     
+# Liveness check
+
+It is possible to check when Styx is ready for serve requests by accessing an admin endpoint at `/admin/proxystatus`.
```
mikkokar: As discussed, let's call this `/admin/styx/proxy/status`.

dvlato: As discussed, let's return  a JSON object for this status:
{ 
status: "UP|DOWN"
}

components/proxy/src/test/java/com/hotels/styx/StyxServerTest.java
```diff
@@ -100,6 +107,65 @@ public void invokesPluginLifecycleMethods() {
...
+                    sendGet(client, "http://localhost:%s/admin/proxystatus", styxServer.adminHttpAddress().getPort()),
+                    is(""
+                            + "{\n"
+                            + "  status:DOWN\n"
+                            + "}"
+                            + "\n"));
+        } finally {
+            stopIfRunning(styxServer);
+        }
+    }
```
mikkokar: Thinking about this, we now have two states, `UP` and `DOWN`. However, if styx server fails to start, it remains indefinitely in `DOWN` state. This could be confusing for the end-users because they cannot confidently say if styx has failed to start, or if it is just slow in starting.

For this reason let's consider adding a `FAILED` state.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
```diff
@@ -0,0 +1,57 @@
...
+        String resultDescription = configStore.get("server.started.proxy", Result.class)
+                .map(result -> (Result<String>) result)
+                .map(result -> result.mapSuccess(any -> "STARTED"))
+                .map(result -> result.defaultOnFailure(any -> "FAILED"))
+                .map(Result::successValue)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .orElse("INCOMPLETE");
+
+        String json = format("{%n  status:%s%n}%n", resultDescription);
```
mikkokar: Please ensure this endpoint returns syntactically valid JSON. According to jsonlint.com it should be:

```
{
	"status": "STARTED"
}
```

mikkokar: Also you need to set content-type header accordingly.

mikkokar: The admin interface remains unreachable until proxy server has started. You can verify this by adding a `Thread.sleep(30000)` in the `ProxyServerStartup.initialisePlugins` method.

I think this was one of the requirements for this change?

kvosper: I will investigate.

mikkokar: One approach is to use the new `ConfigStore`. Proxy Server could publish the plugin admin endpoints into the `ConfigStore`, and Admin Server can then read the endpoints when rendering the admin index page.

components/common/src/main/java/com/hotels/styx/configstore/ConfigStore.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Stores data about the current state of the system.
+ * Added to allow Styx to operate in a more dynamic, adaptive way.
+ */
+public class ConfigStore {
```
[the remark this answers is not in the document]
mikkokar: I am happy with `ConfigStore`. It is abstract enough not to specify if it is runtime configuration state or some static startup configuration.

PR #178, opened 2018-06-08
Refactor Styx Client API consumers

mikkokar:
Refactor to reduce dependencies to `HttpClient` interface. This allows us to eventually move the `HttpClient` interface away from the API module into styx-client module.

This PR replaces usages of `HttpClient` with `HttpHandler`. The API objects now see the `HttpHandler` everywhere. Where an API object needs a client, styx core provides it wrapped inside an HTTP handler.

I have not been fully able to remove the `HttpHandler` yet. It is still needed by the URL connection HTTP client. But once this work is merged, I will prepare another PR where the URL connection HTTP client is replaced with our Simple Netty HTTP client.


components/client/src/main/java/com/hotels/styx/client/OriginsInventory.java
```diff
@@ -22,13 +22,14 @@
...
 import com.hotels.styx.api.Announcer;
+import com.hotels.styx.api.HttpHandler;
 import com.hotels.styx.api.Id;
+import com.hotels.styx.api.client.RemoteHost;
 import com.hotels.styx.api.client.ActiveOrigins;
 import com.hotels.styx.api.client.ConnectionPool;
 import com.hotels.styx.api.client.Origin;
-import com.hotels.styx.api.client.OriginsSnapshot;
 import com.hotels.styx.api.client.OriginsChangeListener;
-import com.hotels.styx.api.client.RemoteHost;
```
dvlato: This is really picky, but if you are going to make other changes, maybe  you can try re-organizing the imports so they are ordered in alphabetical order again?

components/client/src/test/unit/java/com/hotels/styx/client/StyxHttpClientTest.java
```diff
@@ -448,6 +450,10 @@ public void prefersRestrictedOriginsOverStickyOriginsWhenBothAreConfigured() {
         assertThat(argPreferences.getValue().preferredOrigins(), is(Optional.of("Origin-Y")));
     }
 
+    private HttpHandler toHandler(StyxHostHttpClient hostClient) {
```
dvlato: Do you think we could improve the method name? I feel it's not clear enough but I don't know how we could improve it...

mikkokar: I wouldn't bother. It is a private method inside a test class.

components/client/src/main/java/com/hotels/styx/client/OriginsInventory.java
mikkokar: ok

PR #181, opened 2018-06-11
 Decouple admin server from plugins startup, add proxy status endpoint 

kvosper:


components/common/src/main/java/com/hotels/styx/common/Result.java
```diff
@@ -0,0 +1,98 @@
...
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Represents the outcome of attempting to do something.
+ * On a success a value is stored (defaults to "success" if not specified).
+ * On a failure an exception is stored (defaults to a new Exception if not specified).
+ *
+ * @param <T> type of success value
+ */
+public class Result<T> {
```
dvlato: I haven't checked the code that uses this class or thought about it considering we are doing Java, but I believe we wanted to implement this as an Either/Try monad?  So I would expect : map, flatMap...  Scala's try monad is nicer (as in simpler) https://mauricio.github.io/2014/02/17/scala-either-try-and-the-m-word.html , but  it seems there are a few implementations in Java such as  https://github.com/jasongoodwin/better-java-monads/blob/master/src/main/java/com/jasongoodwin/monads/Try.java

kvosper: The goal of this class was simply to add a representation of success or failure that can be stored as a value.

I can see a similarity with the `Try.java` linked, but I'm not sure it is worth adding a new dependency just for this class.

In terms of API, a difference I see is that the `Result` class can be instantiated directly with a call to `success` or `failure` without needing to wrap a block of code, so it is in that sense more decoupled from the try-catch structure and can be used in situations where try-catch does not appear (e.g. using result values in unit tests).

components/common/src/test/java/com/hotels/styx/configstore/ConfigStoreTest.java
```diff
@@ -0,0 +1,122 @@
...
+    @Test
+    public void listensOnSeparateThread() {
+        Latch unlockedByTestThread = new Latch(1);
+        Latch unlockedBySubscribeThread = new Latch(1);
+
+        configStore.watch("foo", String.class)
+                .subscribe(value -> {
+                    unlockedByTestThread.await(1, SECONDS);
+                    unlockedBySubscribeThread.countDown();
+                });
```
mikkokar: When registering a watch, a possibly existing value should be propagated straight away. This prevents unnecessary race conditions where a consumer is late to observe the event.

Please confirm if the `watch` behaves like this?

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    protected HttpResponse doHandle(HttpRequest request) {
+        String resultDescription = configStore.get("server.started.proxy", Result.class)
+                .map(result -> (Result<String>) result)
+                .map(result -> result.mapSuccess(any -> "STARTED"))
+                .map(result -> result.defaultOnFailure(any -> "FAILED"))
+                .map(Result::successValue)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .orElse("INCOMPLETE");
```
mikkokar: I wouldn't store anything complicated in the "server.started.proxy" attribute. Please consider just storing an enum of `STARTED`, `FAILED` or `INCOMPLETE` instead of `Result`

components/common/src/test/java/com/hotels/styx/configstore/ConfigStoreTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+        configStore.set("foo", "bar");
+        unlockedByTestThread.countDown();
+        unlockedBySubscribeThread.await(2, SECONDS);
```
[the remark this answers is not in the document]
kvosper: > Is it to show that the watch subscriptions run from a different thread from the config setter thread?

Yes.

> If so, does every watch subscriber run in their own thread? How far will it scale? Say I have say 10 watch subscribes on "foo", then on setting "foo" do they run simultaneously on 10 different threads? Or are they queued to a thread pool?

That code is already part of this PR. You can look at the `watch` method in the `ConfigStore` class. Please review.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
mikkokar: Also, a caused exception is stored in the result object (`failure(cause)`) in `NettyServer` java, just to be thrown away from the server status response. 

Therefore consider removing `Result` class. 

The detailed error cause can be logged.

components/common/src/test/java/com/hotels/styx/configstore/ConfigStoreTest.java
kvosper: This is not currently the case, I will fix it.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProxyStatusHandler.java
mikkokar: Also it is going to eliminate the above 6 or 7 lines of code :-)

components/common/src/test/java/com/hotels/styx/configstore/ConfigStoreTest.java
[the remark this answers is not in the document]
mikkokar: Ok this was just to probe your intentions regarding config store threading model. I think it is ok for now. Perhaps in future we may want to allow consumers to provide a custom executor.

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -82,68 +86,104 @@
...
+                .add("/admin/tasks/origins", new HttpMethodFilteringHandler(POST, new OriginsCommandHandler(environment.eventBus())))
+                ;
+
+        environment.configStore().watch("plugins", List.class)
+                .map(list -> (List<NamedPlugin>) list)
+                .subscribe(plugins -> {
+                    routesForPlugins(plugins).forEach(route -> httpRouter.add(route.path(), route.handler()));
+                    httpRouter.add("/admin/tasks/plugin/", new PluginToggleHandler(plugins));
+                    httpRouter.add("/admin/plugins", new PluginListHandler(plugins));
+                });
```
mikkokar: We don't necessarily need a watch functionality for this use case. You can just render the admin links every time a request comes in. This way you'll just query the config store and add the links to the end of the page.

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -82,68 +86,104 @@
...
+        private final Set<String> paths = new HashSet<>();
+
+        public RouterBuilder add(String path, HttpHandler2 handler) {
+            this.router.add(path, handler);
+            this.paths.add(path);
+            return this;
+        }
+    }
+
+    private static void verifyThatLinksInIndexMapToRealEndpoints(Iterable<IndexHandler.Link> links, Set<String> paths) {
```
mikkokar: Isn't this something we should have in the unit/e2e tests, rather than in the running code?

What would be the suggested remedial action for the end user if this exception ever fires?

components/proxy/src/main/java/com/hotels/styx/proxy/ProxyServerBuilder.java
```diff
@@ -54,28 +56,29 @@ public HttpServer build() {
...
     }
 
     private HttpResponse.Builder addInfoHeader(HttpResponse.Builder responseBuilder, HttpRequest request) {
         return responseBuilder.header(styxInfoHeaderName, responseInfoFormat.format(request));
     }
 
-    public ProxyServerBuilder httpHandler(HttpHandler2 httpHandler) {
+    public ProxyServerBuilder httpHandler(Supplier<HttpHandler2> httpHandler) {
         this.httpHandler = httpHandler;
         return this;
```
mikkokar: Can you remind me why the `httpHnadler` has to be loaded lazily? To prevent a blocking load of plugins before the admin interface starts?

kvosper: Yes that is the reason.

components/proxy/src/main/java/com/hotels/styx/startup/ProxyServerSetUp.java
```diff
@@ -66,18 +66,20 @@ private static void initialisePlugins(Iterable<NamedPlugin> plugins) {
         if (exceptions > 0) {
             throw new RuntimeException(format("%s plugins failed to start", exceptions));
         }
+
+        configStore.set("plugins", plugins);
```
mikkokar: Ideally this would move inside the loop:

```
        for (NamedPlugin plugin : plugins) {
            try {
                plugin.styxStarting();
                configStore.set("plugins", pluginsInitialisedSoFar);    // <----
            } catch (Exception e) {
                exceptions++;
                LOG.error("Error starting plugin '{}'", plugin.name(), e);
            }
        }
```

This would give a more instantaneous feedback about the plugins loaded so far. The end user could monitor the progress of plugins being initialised. Maybe something to consider in future.

However the current implementation is fine for now.

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
kvosper: This code does not render links, it builds the HttpRouter.

components/proxy/src/main/java/com/hotels/styx/startup/ProxyServerSetUp.java
kvosper: I had the same thought, but for now the current implementation is simpler.

mikkokar:
The Travis build has started failing. I assume you are still working on this PR?

There are three failing e2e tests in `OriginsReloadRepetitionSpec` and `OriginsReloadSpec` test suites. All of them show similar symptoms:

```
[31m  The code passed to eventually never returned normally. Attempted 60 times over 1.00299402 seconds. Last failure message: "{"appOne":{"appId":"appOne","activeOrigins":[],"inactiveOrigins":[],"disabledOrigins":[]}}" did not include substring "localhost:41693". (OriginsReloadSpec.scala:73)[0m
[31m- should reflect origin removal in origins status page *** FAILED ***[0m
```

That is, no origins are shown in the origin status response.

kvosper:
An update on failures - I cannot reproduce this failure locally, making it difficult to diagnose. I am going to try increasing timeouts on the tests in `OriginsReloadSpec` to see if it makes a difference. I will likely also add extra logging as I cannot see any difference in the logs when the tests pass compared to these failures.

kvosper:
Looks like the test `OriginsReloadSpec` causes two `OriginsSnapshot`s to be sent when it changes the backends. First to remove the old origins, then to add the new ones.
For some reason, on travis, the snapshot from the removal is being received later than the snapshot from the addition, i.e. out of order.

kvosper:
This is being postpone until required improvements are made separately.

kvosper:
Closing as this is dependent on other work, and the codebase has changed since it was opened. It will probably be done after Styx 1.0 is released.

PR #186, opened 2018-06-12
Cookies should not be case insensitive 

### LINKED ISSUE #185: Cookies should not be case insensitive
## The problem
Cookies names are treated in a case-insensitive manner inside Styx, but this doesn't seem to agree with RFC 6265 (https://tools.ietf.org/html/rfc6265) and most popular browsers are in fact case sensitive (https://stackoverflow.com/questions/11311893/is-the-name-of-a-cookie-case-sensitive?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)
## Acceptance criteria
* Cookie names should be compared in a case sensitive manner. 
* Other attributes of the cookie and headers should remain case insensitive.

dvlato:
Fixes #185

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
+++ b/components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
@@ -42,6 +42,7 @@ import static com.hotels.styx.api.HttpHeaderNames.CONNECTION;
 import static com.hotels.styx.api.HttpHeaderNames.CONTENT_LENGTH;
 import static com.hotels.styx.api.HttpHeaderNames.HOST;
 import static com.hotels.styx.api.HttpHeaderValues.KEEP_ALIVE;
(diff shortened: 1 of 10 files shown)


components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -335,7 +335,7 @@ private FormData toFormData(ByteBuf byteBuf) {
      */
     public Optional<HttpCookie> cookie(String name) {
         return cookies().stream()
-                .filter(cookie -> name.equalsIgnoreCase(cookie.name()))
+                .filter(cookie -> name.equals(cookie.name()))
                 .findFirst();
     }
```
mikkokar: I would add few tests to demonstrate that the cookie names indeed are case insensitive.

Issue #177, opened 2018-06-05
Document use of test-api for creating plugins


kvosper:
## The problem
Styx has a good `test-api` to use when writing plugins, but there is no documentation showing how to test plugins with this API.

## Acceptance criteria
* Write this documentation

mikkokar:
The `test-api` is not really used by anybody. The styx server can be started directly from the proxy project.

Therefore closing this.

PR #187, opened 2018-06-13
Allow consumer to configure styx test api server ports.

mikkokar:
Makes it possible for plugin tests to configure HTTP port numbers for the test server.

### DIFF
--- a/support/test-api/src/main/java/com/hotels/styx/testapi/StyxServer.java
+++ b/support/test-api/src/main/java/com/hotels/styx/testapi/StyxServer.java
@@ -190,6 +190,7 @@ public final class StyxServer {
 
         /**
          * Specifies the HTTP port for proxy server.
+         * Set port number to 0 (zero) to let Styx automatically allocate a free port number.
          *
          * @param proxyPort
          * @return this builder
@@ -201,6 +202,7 @@ public final class StyxServer {
 
         /**
          * Specifies the HTTPS port for proxy server.
+         * Set port number to 0 (zero) to let Styx automatically allocate a free port number.
          *
          * @param proxyPort
          * @return this builder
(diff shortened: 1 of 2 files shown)


support/test-api/src/main/java/com/hotels/styx/testapi/StyxServer.java
```diff
@@ -184,6 +184,42 @@ public HealthCheckRegistry healthCheckRegistry() {
     public static final class Builder {
         private final Map<String, com.hotels.styx.api.service.BackendService> routes = new HashMap<>();
         private final List<PluginFactoryConfig> pluginFactories = new ArrayList<>();
+        private int proxyHttpPort;
+        private int adminHttpPort;
+        private int proxyHttpsPort;
+
+        /**
+         * Specifies the HTTP port for proxy server.
```
kvosper: These methods should have javadoc comments explain that the endpoints will be assigned random free ports if not set explicitly.

kvosper: Side note: Speaking of documentation, we do not have actual documentation for using the test-api yet, but we have an issue #177 regarding this.

mikkokar: Ok, I will fix the documentation.

kvosper: Setting it to zero will do it, but it's probably a more likely use case for someone to simply not call the method at all. I would phrase the comment like:
`By default, Styx will automatically allocate a free port number. This happens when a port is not set a value on the builder, or it is set a value of 0 (zero).`

PR #192, opened 2018-06-25
Add configstore

kvosper:


support/testsupport/src/main/java/com/hotels/styx/support/Latch.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public Latch(int count) {
+        this.latch = new CountDownLatch(count);
+    }
+
+    public void await() {
+        try {
+            latch.await();
+        } catch (InterruptedException e) {
+            throw propagate(e);
```
dvlato: re-throwing an InterruptedException as a different Exception is not a good practice in general. I would call "Thread.interrupt()" or document it clearly when we consider it is not needed.

PR #193, opened 2018-06-25
Add config store

kvosper:
Cherry-picked from previous branch to styx-1.0-dev


components/common/src/test/java/com/hotels/styx/configstore/ConfigStoreTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        configStore.watch("foo", String.class)
+                .subscribe(value -> {
+                    unlockedByTestThread.await(1, SECONDS);
+                    unlockedBySubscribeThread.countDown();
+                });
+
+        configStore.set("foo", "bar");
+        unlockedByTestThread.countDown();
+        unlockedBySubscribeThread.await(2, SECONDS);
```
mikkokar: CyclicBarrier seems more appropriate for this use case.

support/testsupport/src/main/java/com/hotels/styx/support/Latch.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import static com.google.common.base.Throwables.propagate;
+import static java.lang.Thread.currentThread;
+
+/**
+ * Wrapper for CountDownLatch to make it more convenient to use.
+ */
+public class Latch {
```
mikkokar: As per our unwritten naming conventions, utility classes names like this would have their name pluralised.

That is, the class name should be `Latches`.

kvosper: It's not a utility class. Utility classes have only static methods. This is a normal class.

components/common/src/test/java/com/hotels/styx/configstore/ConfigStoreTest.java
kvosper: I don't know what difference that would make.

support/testsupport/src/main/java/com/hotels/styx/support/Latch.java
mikkokar: Right of course. Now I see that. I should have seen that straight away. 

But why to implement these await methods as a composed class? 

By making it a public test support class you are promoting its use in other tests. But then again the `Latch` class is not type compatible with `CountDownLatch`. Sooner or later some parts of the codebase will be using `Latch` while other parts stick to `CountDownLatch`. After that we will need to bridge the two classes at the interface level.

But all we really want is to re-propagate checked exceptions as runtime exceptions in `await` method.

For these reasons, please convert this to a utility class with `await` as a static utility method.

components/common/src/test/java/com/hotels/styx/configstore/ConfigStoreTest.java
mikkokar: It would be using only one synchronising primitive instead of two. That cuts the complexity by 50% ;-).

mikkokar: I think this is conceptually much simpler:

```
    public void listensOnSeparateThread2() {
        final CyclicBarrier barrier = new CyclicBarrier(2);

        configStore.watch("foo", String.class)
                .subscribe(value -> {
                        await(barrier);
                });

        configStore.set("foo", "bar");
        await(barrier);
    }
```

kvosper: I think this will work. I'll try it.

support/testsupport/src/main/java/com/hotels/styx/support/Latch.java
kvosper: I'll look into this.

components/common/src/test/java/com/hotels/styx/configstore/ConfigStoreTest.java
kvosper: This is not sufficient to test the functionality. The ordering matters. If we use `CyclicBarrier` then a `watch` method that uses the same thread as the unit test would still pass, since both awaits will get called.

To prove that a separate thread is used, the `watch` must be able to wait for the test thread. That said, it probably only needs the `unlockedByTestThread` latch, not both.

kvosper: OK.

support/testsupport/src/main/java/com/hotels/styx/support/LatchesAndBarriers.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public static void await(CountDownLatch latch, long timeout, TimeUnit unit) throws IllegalStateException {
+        try {
+            boolean completed = latch.await(timeout, unit);
+
+            if (!completed) {
+                throw new IllegalStateException(format("Latch timed out: max wait was %s %s", timeout, unit));
+            }
+        } catch (InterruptedException e) {
+            currentThread().interrupt();
```
mikkokar: Surely we need to re-throw  something from the InterruptedException case? Otherwise `await` just returns and the calling code carries on as if nothing happened.

support/testsupport/src/main/java/com/hotels/styx/support/LatchesAndBarriers.java
```diff
@@ -0,0 +1,57 @@
...
+import static java.lang.Thread.currentThread;
+
+/**
+ * Makes `CountDownLatch` more convenient for unit tests.
+ */
+public final class LatchesAndBarriers {
+    private LatchesAndBarriers() {
+    }
+
+    public static void await(CountDownLatch latch, long timeout, TimeUnit unit) throws IllegalStateException {
```
dvlato: To be honest, I don't understand why we think we need to capture InterruptedException here at all. The test can fail due to an InterruptedException and, in that case, it makes all the sense to have it in the signature of the Test itself.
If we still wanted to wrap the code of await() , we could have a method that receives a lambda instead of having to duplicate the code for CountdownLatch and CyclicBarrier.

support/testsupport/src/main/java/com/hotels/styx/support/LatchesAndBarriers.java
dvlato: Sure I failed to notice that we had stopped throwing an Exception here. My advice is that we just let InterruptedException propagate...

support/testsupport/src/main/java/com/hotels/styx/support/LatchesAndBarriers.java
mikkokar: I think the problem is with the lambda expressions. Without this utility you would be forced to try/catch inside lambda expressions.

kvosper: As Mikko says, the problem is that `InterruptedException` is a checked exception, and the interfaces our lambdas are implementing do not allow checked exceptions to be thrown.

support/testsupport/src/main/java/com/hotels/styx/support/LatchesAndBarriers.java
kvosper: Unfortunately, `Observable.subscribe` takes an `Action1` as its argument, which will not allow `InterruptedException` to be thrown.

support/testsupport/src/main/java/com/hotels/styx/support/LatchesAndBarriers.java
dvlato: Oh, sure. I think  some of the code inside the tests uses the Latch outside a lambda. 

 Then my suggestion of using a method that receives a lambda (with InterruptedException in the signature) and propagates InterruptedException as a RuntimeException (but keeping the Thread.interrupt() ) makes more sense  than wrapping specific classes.

mikkokar: Hi @dvlato. It no longer wraps a class. It is just a static helper method that accepts a `CountDownLatch` as an argument. IMO it should just set the interrupted flag and propagate an InterruptedException wrapped inside a RuntimeException.

Issue #196, opened 2018-06-28
Create unit tests for StyxCoreObservable in 1.0 branch


kvosper:
## The problem
`StyxCoreObservable` does not have associated unit tests.

## Detailed description
It is important to have unit tests for `StyxCoreObservable`. This class can be indirectly created through the API, using `StyxObservable.of`, so we cannot rely on our own system tests.

## Acceptance criteria
* Create unit tests for `StyxCoreObservable`

PR #198, opened 2018-06-28
Tests for styx core observable

kvosper:
For #196

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
+++ b/components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
@@ -36,7 +36,7 @@ class StyxCoreObservable<T> implements StyxObservable<T> {
         this.delegate = toObservable(future);
     }
 
-    public static <T> Observable<T> toObservable(CompletionStage<T> future) {
+    private static <T> Observable<T> toObservable(CompletionStage<T> future) {
         return Observable.create(subscriber ->
                 future.whenComplete((result, error) -> {
                     if (error != null) {
@@ -92,7 +92,7 @@ class StyxCoreObservable<T> implements StyxObservable<T> {
     }
 
     private static <T> CompletableFuture<T> fromSingleObservable(Observable<T> observable) {
-        final CompletableFuture<T> future = new CompletableFuture<>();
+        CompletableFuture<T> future = new CompletableFuture<>();
         observable.single().subscribe(future::complete, future::completeExceptionally);
         return future;
     }
--- a/components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
+++ b/components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
@@ -24,6 +24,7 @@ import java.util.function.Function;
 
 /**
  * An observable that underpins the StyxObservable interface.
+ *
  * @param <T>
  */
 class StyxCoreObservable<T> implements StyxObservable<T> {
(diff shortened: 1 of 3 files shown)


components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
```diff
@@ -92,7 +94,7 @@ public StyxCoreObservable(CompletionStage<T> future) {
     }
 
     private static <T> CompletableFuture<T> fromSingleObservable(Observable<T> observable) {
-        final CompletableFuture<T> future = new CompletableFuture<>();
```
dvlato: Why do you need to remove the final modifier?

components/api/src/test/java/com/hotels/styx/api/StyxCoreObservableTest.java
```diff
@@ -0,0 +1,107 @@
...
+            new StyxCoreObservable<>(just("foo", "bar"))
+                    .asCompletableFuture()
+                    .get(1, SECONDS);
+        } catch (ExecutionException e) {
+            throw e.getCause();
+        }
+    }
+
+    @Test(expectedExceptions = NoSuchElementException.class, expectedExceptionsMessageRegExp = "Sequence contains no elements")
+    public void throwsExceptionWhenTryingToConvertEmptyToFuture() throws Throwable {
```
mikkokar: It is worth noticing that `NoSuchElementException` and `IllegalArgumentException` are now going to be part of our API. Worth documenting this in the StyxObservable interface.

components/api/src/test/java/com/hotels/styx/api/StyxCoreObservableTest.java
```diff
@@ -0,0 +1,107 @@
...
+        String result = new StyxCoreObservable<>(just("foo", "bar", "baz"))
+                .reduce((a, b) -> b + a, "")
+                .asCompletableFuture()
+                .get(1, SECONDS);
+
+        assertThat(result, is("foobarbaz"));
+    }
+
+    @Test
+    public void flatMapsWorksWithCustomObservable() throws InterruptedException, ExecutionException, TimeoutException {
```
mikkokar: I don't understand this test.

components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
kvosper: Just keeping it in line with our standard coding practices. Excessive `final`s clutter the codebase, so we avoid them on local variables and method parameters (due to their limited scope). 
We only use them on fields because it could be very hard to tract down an unexpected mutation in a large class.
For locals and parameters, we instead aim to keep our methods small and simple enough that any mutations are easily visible (in addition to the other benefits of readable methods).

components/api/src/test/java/com/hotels/styx/api/StyxCoreObservableTest.java
kvosper: This test exists because of the problem you observed in your `TODO`:
> TODO: Mikko: Dangerous cast.
> Because StyxObservable is an interface, nothing prevents plugins from
> implementing it and returning a derived object from the `transformation`
> function. This would obviously break the cast.

It proves that the problem described has been fixed.

components/api/src/test/java/com/hotels/styx/api/StyxCoreObservableTest.java
kvosper: OK

Issue #182, opened 2018-06-11
Add admin endpoint for querying specific metrics


kvosper:
## The problem
We currently have an endpoint that returns the entire set of metrics, but it may be useful to have an endpoint that when queried can return an arbitrary metric specified in the HTTP request. This could allow more sophisticated health-checking options for those that need it.

## Detailed description
The URL pattern will be `/admin/metrics/<METRIC_NAME_OR_PREFIX>`, so if the URL was:
`/admin/metrics/foo.bar`
it would return a metric named `foo.bar` as well as metrics beginning with this prefix such as `foo.bar.baz`.

These can be returned as a JSON object by simply filtering the map of metrics for those that match the pattern and converting it to JSON.

If the metric does not exist it can return 404.

## Acceptance criteria
* Add an admin endpoint that allows the user to drill down to a specific metric
* Returns metric matching exact name if it exists
* Returns metric whose name has specified prefix
* If no metrics are found, returns 404

mikkokar:
Cherry picked to `styx-1.0-dev` by #218 .

PR #206, opened 2018-07-03
Add admin endpoint for querying specific metrics by name

kvosper:
Implements #182

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
@@ -115,7 +115,9 @@ public class AdminServerBuilder {
         httpRouter.add("/admin/status", new StatusHandler(healthCheckHandler));
         httpRouter.add("/admin/ping", new PingHandler());
         httpRouter.add("/admin/threads", new ThreadsHandler());
-        httpRouter.add("/admin/metrics", new MetricsHandler(environment.metricRegistry(), metricsCacheExpiration));
+        MetricsHandler metricsHandler = new MetricsHandler(environment.metricRegistry(), metricsCacheExpiration);
+        httpRouter.add("/admin/metrics", metricsHandler);
+        httpRouter.add("/admin/metrics/", metricsHandler);
         httpRouter.add("/admin/healthcheck", healthCheckHandler);
         httpRouter.add("/admin/configuration", new StyxConfigurationHandler(staticConfiguration()));
         httpRouter.add("/admin/configuration/origins", new OriginsHandler(backendServicesRegistry));
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
@@ -31,6 +31,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.function.BiPredicate;
 import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static com.hotels.styx.api.messages.FullHttpResponse.response;
 import static com.hotels.styx.api.messages.HttpResponseStatus.OK;
@@ -45,6 +47,8 @@ import static rx.Observable.just;
  * Handler for showing all registered metrics for styx server. Can cache page content.
  */
 public class MetricsHandler extends JsonHandler<MetricRegistry> {
+    private static final Pattern SPECIFIC_METRICS_PATH_PATTERN = Pattern.compile(".*/metrics/(.+)/?");
+
     private static final boolean DO_NOT_SHOW_SAMPLES = false;
     private final CodaHaleMetricRegistry metricRegistry;
 
@@ -65,13 +69,12 @@ public class MetricsHandler extends JsonHandler<MetricRegistry> {
 
     @Override
     public Observable<HttpResponse> handle(HttpRequest request) {
-        String path = removeFinalSlash(request.path());
+        return metricName(request.path())
+                .map(metricName -> specificMetricsResponse(request, metricName))
+                .orElseGet(() -> super.handle(request));
+    }
 
-        if (path.equals("/admin/metrics")) {
-            return super.handle(request);
-        }
-
-        String metricName = metricNameFromPath(path);
+    private Observable<HttpResponse> specificMetricsResponse(HttpRequest request, String metricName) {
         boolean pretty = request.queryParam("pretty").isPresent();
 
         String serialised = serialise(metrics(metricName), pretty);
@@ -80,6 +83,16 @@ public class MetricsHandler extends JsonHandler<MetricRegistry> {
                 .body(serialised, UTF_8)
                 .build()
                 .toStreamingResponse());
+    }
+
+    private static Optional<String> metricName(String path) {
+        Matcher matcher = SPECIFIC_METRICS_PATH_PATTERN.matcher(path);
+
+        if (matcher.matches()) {
+            return Optional.of(matcher.group(1));
+        }
+
+        return Optional.empty();
     }
 
     private String serialise(Object object, boolean pretty) {
@@ -110,15 +123,5 @@ public class MetricsHandler extends JsonHandler<MetricRegistry> {
 
     private static <K, V> Predicate<Map.Entry<K, V>> entryPredicate(BiPredicate<K, V> predicate) {
         return entry -> predicate.test(entry.getKey(), entry.getValue());
-    }
-
-    private static String metricNameFromPath(String path) {
-        return path.substring("/admin/metrics/".length());
-    }
-
-    private static String removeFinalSlash(String path) {
-        return path.endsWith("/")
-                ? path.substring(0, path.length() - 1)
-                : path;
     }
 }
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
@@ -23,21 +23,22 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectWriter;
 import com.hotels.styx.api.HttpRequest;
 import com.hotels.styx.api.HttpResponse;
-import com.hotels.styx.api.messages.FullHttpResponse;
 import com.hotels.styx.api.metrics.codahale.CodaHaleMetricRegistry;
 import rx.Observable;
 
 import java.time.Duration;
+import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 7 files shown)


components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -115,7 +115,9 @@ public HttpServer build() {
         httpRouter.add("/admin/status", new StatusHandler(healthCheckHandler));
         httpRouter.add("/admin/ping", new PingHandler());
         httpRouter.add("/admin/threads", new ThreadsHandler());
-        httpRouter.add("/admin/metrics", new MetricsHandler(environment.metricRegistry(), metricsCacheExpiration));
+        MetricsHandler metricsHandler = new MetricsHandler(environment.metricRegistry(), metricsCacheExpiration);
+        httpRouter.add("/admin/metrics", metricsHandler);
+        httpRouter.add("/admin/metrics/", metricsHandler);
         httpRouter.add("/admin/healthcheck", healthCheckHandler);
         httpRouter.add("/admin/configuration", new StyxConfigurationHandler(staticConfiguration()));
         httpRouter.add("/admin/configuration/origins", new OriginsHandler(backendServicesRegistry));
```
mikkokar: This feature needs to be described in the Styx user manual. Please add relevant information there.

mikkokar: Also, this feature would be more useful if it could return all metrics values starting with a prefix. For example:

`curl http://localhost:8080/admin/metrics/jvm.bufferpool`

Would return values for:

```
"jvm.bufferpool.direct.capacity": {
"value": 1
},
"jvm.bufferpool.direct.count": {
"value": 2
},
"jvm.bufferpool.direct.used": {
"value": 2
},
"jvm.bufferpool.mapped.capacity": {
"value": 0
},
"jvm.bufferpool.mapped.count": {
"value": 0
},
"jvm.bufferpool.mapped.used": {
"value": 0
},
```

components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
```diff
@@ -39,6 +57,58 @@
...
+    @Override
+    public Observable<HttpResponse> handle(HttpRequest request) {
+        String path = removeFinalSlash(request.path());
+
+        if (path.equals("/admin/metrics")) {
+            return super.handle(request);
+        }
+
+        String metricName = metricNameFromPath(path);
+        boolean pretty = request.queryParam("pretty").isPresent();
```
mikkokar: I wouldn't represent them as URL path components. This is because Coda Hale metric names are not really (counter-intuitively) scoped. 

For example `jvm.bufferpool.direct` is not really a scope in Coda Hale metrics. It can be simultaneously a prefix for other metrics like `jvm.bufferpool.direct.capacity` *and* a metric of its own.

Using path components to refer to a metric, as in `/metrics/jvm/bufferpool/direct`, gives an impression that they are scoped. Would this path refer to a metric of that name, or all metrics under that scope?

For that reason I propose to refer to the metrics names just as strings, so that `/metrics/jvm.bufferpool.direct` would be treated as all metrics that start with that name.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
```diff
@@ -39,6 +57,58 @@
...
+        super(requireNonNull(metricRegistry.getMetricRegistry()), cacheExpiration, new MetricsModule(SECONDS, MILLISECONDS, DO_NOT_SHOW_SAMPLES));
+
+        this.metricRegistry = metricRegistry;
+    }
+
+    @Override
+    public Observable<HttpResponse> handle(HttpRequest request) {
+        String path = removeFinalSlash(request.path());
+
+        if (path.equals("/admin/metrics")) {
```
dvlato: Can't we find a way to do this avoiding this string duplication (/admin/metrics) ?  I think in this case it might be justified to design bearing in mind that we might also want to expose also specific origins... I mean, do it slightly more generic.
  
Also, the metrics are documented and displayed in the metrics endpoint  with dots. E.g., "requests.received" . I don't see why we are using slashes in the endpoint for specific metrics,  i would say it's an unnecessary and unwanted change.

mikkokar: The idea was to use the URL path components to separate metric scopes. But not sure if that matters, really.

dvlato: I would say that in that case, we would also change the 'global metrics endpoint' and the docs to use those scope separators.

mikkokar: Hi. Yes I’ve already left a comment about docs.

Could you elaborate your comment about ‘global metrics endpoint’?

dvlato: The '/admin/metrics' endpoint which contains all metrics with values such as: 
    "connections.openssl.session.accept" ,
    "connections.openssl.session.acceptGood"

In my opinion, if we supported metric scopes we wouldn't display the metrics in this endpoint as connections.openssl but  as 'connections/openssl' (if we had a scope named 'connections').   Why would the displayed metric name have to be transformed in order to query individual parameters?

docs/user-guide/metrics.md
```diff
@@ -5,6 +5,23 @@ can be viewed via admin interface at `http://<styx-host>/admin/metrics`.
...
+This will return the metric that matches the specified name (if it exists) as well as any metrics
+that use the specified name as a prefix.
+
+For example `http://<styx-host>/admin/metrics/requests.response` would return
+
+    requests.response.sent
+    requests.response.status.1xx
+    requests.response.status.2xx
+    ...
+
```
mikkokar: What does it return for an unrecognised metric name?

mikkokar: Now thinking about it, if an end user queries for `admin/metrics/requests.respo`, styx should still return all metric names starting with `request.respo`. 

After playing with this it now feels bit arbitrary and even confusing to require users to stick with strict scopes.

mikkokar: Please change so that `admin/metrics/requests.respo` returns all metrics starting with `requests.respo`.

mikkokar: We also need e2e tests for this feature.

kvosper: It will be an empty JSON object. I'll add a test for it.

kvosper: OK

mikkokar: Should it not return a 404 Not found?

kvosper: Depends on how we look at it. I see two possible interpretations:
1. We failed to find a page matching that metric prefix
2. We succeeded in finding a page that tells us what metrics exist with that prefix

components/proxy/src/test/java/com/hotels/styx/admin/handlers/MetricsHandlerTest.java
```diff
@@ -17,33 +17,57 @@
...
-        FullHttpResponse response = waitForResponse(handler.handle(get("/metrics").build()));
+        FullHttpResponse response = waitForResponse(handler.handle(get("/admin/metrics").build()));
         assertThat(response.bodyAs(UTF_8), is("{\"version\":\"3.1.3\",\"gauges\":{},\"counters\":{\"foo\":{\"count\":1}},\"histograms\":{},\"meters\":{},\"timers\":{}}"));
     }
+
+    @Test
+    public void canRequestMetricsBeginningWithPrefix() {
+        metricRegistry.counter("foo.bar").inc(1);
+        metricRegistry.counter("foo.bar.baz").inc(1);
+        metricRegistry.counter("foo.barx").inc(1); // should not be included
```
mikkokar: I would move the "foo.barx" case into a separate test.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
```diff
@@ -39,6 +61,61 @@
...
+    }
+
+    private Observable<HttpResponse> specificMetricsResponse(HttpRequest request, String metricName) {
+        Map<String, Metric> metrics = metrics(metricName);
+
+        if (metrics.isEmpty()) {
+            return just(response(NOT_FOUND).build().toStreamingResponse());
+        }
+
+        boolean pretty = request.queryParam("pretty").isPresent();
```
mikkokar: I am starting to think this "pretty" query param should be moved out into a separate class, and somehow composed with other admin handlers. But that would be a separate improvement.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
```diff
@@ -15,22 +15,44 @@
...
+import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
+import static rx.Observable.just;
 
 /**
  * Handler for showing all registered metrics for styx server. Can cache page content.
  */
 public class MetricsHandler extends JsonHandler<MetricRegistry> {
+    private static final Pattern SPECIFIC_METRICS_PATH_PATTERN = Pattern.compile(".*/metrics/(.+)/?");
```
mikkokar: Why does SPECIFIC_METRICS_PATH_PATTERN accept random characters at the beginning of `/metrics` path? Should the path always start strictly with `/metrics/`?

My comment refers to the `.*` part of the regex.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
```diff
@@ -39,6 +61,61 @@
...
+        ObjectWriter writer = pretty ? metricSerialiser.writerWithDefaultPrettyPrinter() : metricSerialiser.writer();
+
+        try {
+            return writer.writeValueAsString(object);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private Map<String, Metric> metrics(String metricNameStart) {
```
mikkokar: Could we rename `metricNameStart` to `metricNamePrefix` or something?

In my mind `metricNameStart` sounds like an index offset or a boolean variable.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
kvosper: The `.*` allows us to have `/admin/` (or any other parent path) without needing to couple the `MetricsHandler` to it.

PR #217, opened 2018-07-23
Cookies 1.0 api

kvosper:
A PR to make improvements to the way that Styx handles cookies.
The main changes are:
* Use the `Cookie` and `Set-Cookie` headers as the sole source-of-truth instead of trying to keep a separate collection of cookies in the message object.
* Introduce a clear distinction between request cookies (key-value pairs set in `Cookie` header) and response cookies (individual `Set-Cookie` headers that have attributes like `Domain` and `Path`).


components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -309,14 +311,18 @@ private void recordErrorStatusMetrics(HttpResponse response) {
...
+            return addCookie(httpResponse, newStickySessionCookie(id, origin.id(), maxAge));
         } else {
             return httpResponse;
         }
     }
 
+    private static HttpResponse addCookie(HttpResponse response, ResponseCookie cookie) {
+        return response.newBuilder()
+                .addHeader(SET_COOKIE, ResponseCookie.encode(cookie))
+                .build();
```
mikkokar: This should be using a yet-to-be-added `addCookie` method.

components/client/src/test/integration/scala/com/hotels/styx/client/StickySessionSpec.scala
```diff
@@ -145,9 +151,10 @@ class StickySessionSpec extends FunSuite with BeforeAndAfter with ShouldMatchers
       .build
 
     val request: HttpRequest = get("/")
-      .addCookie("other_cookie1", "foo")
-      .addCookie("styx_origin_app", "app-02")
-      .addCookie("other_cookie2", "bar")
+      .cookies(
+        RequestCookie.requestCookie("other_cookie1", "foo"),
+        RequestCookie.requestCookie("styx_origin_app", "app-02"),
+        RequestCookie.requestCookie("other_cookie2", "bar"))
```
mikkokar: static import?

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -614,13 +619,78 @@ public Builder cookies(RequestCookie... cookies) {
...
+         *
+         * @param cookies new cookies
+         * @return this builder
+         */
+        public Builder addCookies(RequestCookie... cookies) {
+            return addCookies(asList(cookies));
+        }
+
+        /**
+         * Adds cookies into this request by overwriting the value of the "Cookie" header.
```
mikkokar: I cannot see any unit tests for `addCookies`. Please add tests for `addCookies()`, or am I just missing something?

mikkokar: Please add unit tests for the following `addTests` scenarios:
1) add a new cookie when Cookie header is not present.
2) add a new cookie when Cookie header is already present.
3) modify a cookie value using `addCookie` when a Cookie header is present and the cookie is already set.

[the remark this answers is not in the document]
mikkokar: I would also elaborate that:

> Adds cookies into Cookie header, or modified existing cookie values.

kvosper: Yes, there are missing tests. I will add tests for the new methods.

components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
```diff
@@ -549,38 +570,91 @@ public Builder method(HttpMethod method) {
...
+
+        private Builder removeCookiesIf(Predicate<String> removeIfName) {
+            Predicate<RequestCookie> keepIf = cookie -> !removeIfName.test(cookie.name());
+
+            List<RequestCookie> newCookies = decode(headers.get(COOKIE)).stream()
+                    .filter(keepIf)
+                    .collect(toList());
+
+            return cookies(newCookies);
+        }
```
mikkokar: This is perhaps bit contrived example, but perhaps worth looking into @kvosper:

```
    @Test
    public void removesCookies2() {
        HttpRequest r1 = HttpRequest.get("/")
                .addCookies(requestCookie("x", "x1"))
                .removeCookies("x")
                .build();

        assertThat(r1.cookie("x"), is(Optional.empty()));
    }
```

This test would fail as follows:

```
java.lang.AssertionError: 
Expected: is <Optional.empty>
     but: was <Optional[x=x1]>
```

I'll find this counter-intuitive.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -362,14 +350,40 @@ public Builder newBuilder() {
...
     }
 
+    /**
+     * Decodes the "Cookie" header in this request and returns the cookies.
+     *
+     * @return cookies
+     */
+    public Set<RequestCookie> cookies() {
+        // Note: there should only be one "Cookie" header, but we check for multiples just in case
+        // the alternative would be to respond with a 400 Bad Request status if multiple "Cookie" headers were detected
```
dvlato: The RFC clearly states that there should be a single Cookie header. I think we can return an error in that case.

components/api/src/main/java/com/hotels/styx/api/cookies/RequestCookie.java
```diff
@@ -0,0 +1,149 @@
...
+                .collect(toSet());
+    }
+
+    /**
+     * Encodes a collection of {@link RequestCookie} objects into a "Cookie" header value.
+     *
+     * @param cookies cookies
+     * @return "Cookie" header value
+     */
+    public static String encode(Collection<RequestCookie> cookies) {
```
dvlato: According to the RFC, servers should not discard duplicate cookie values based on the order of appearance: https://tools.ietf.org/html/rfc6265#section-4.2.2  But it seems difficult to understand what´s expected (other than returning a list) to do with these duplicate cookies, have we verified what´s the most common approach to handle this_?

Also, I don´t know this method is semantically correct if we intend to discard all duplicate cookies and keep the first one. A Collection could be traversed in LIFO order... We may consider that it still means that the first cookie returned will be kept and all the following cookies with the same name removed.

kvosper: The treatment of duplicate cookies is part of the netty encoding logic - it is not something introduced by Styx.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
kvosper: We would need to return the error in the builder. The `cookies` method could be changed to assume only one `Cookie` header exists though.

PR #219, opened 2018-07-24
Fix StyxObservable type parameters

kvosper:
The type parameters were overly strict.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
+++ b/components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
@@ -25,7 +25,7 @@ import java.util.function.Function;
 /**
  * An observable that underpins the StyxObservable interface.
  *
- * @param <T>
+ * @param <T> type of object published
  */
 class StyxCoreObservable<T> implements StyxObservable<T> {
     private final Observable<T> delegate;
@@ -50,16 +50,16 @@ class StyxCoreObservable<T> implements StyxObservable<T> {
         return new StyxCoreObservable<T>(Observable.error(cause));
     }
 
-    public <U> StyxObservable<U> map(Function<? super T, ? extends U> transformation) {
+    public <R> StyxObservable<R> map(Function<? super T, ? extends R> transformation) {
         return new StyxCoreObservable<>(delegate.map(transformation::apply));
     }
 
-    public <U> StyxObservable<U> flatMap(Function<? super T, ? extends StyxObservable<? extends U>> transformation) {
+    public <R> StyxObservable<R> flatMap(Function<? super T, ? extends StyxObservable<? extends R>> transformation) {
         return new StyxCoreObservable<>(delegate.flatMap(response ->
                 toObservable(transformation.apply(response))));
     }
 
-    public <U> StyxObservable<U> reduce(BiFunction<? super T, U, U> accumulator, U seed) {
+    public <R> StyxObservable<R> reduce(BiFunction<? super T, R, R> accumulator, R seed) {
         return new StyxCoreObservable<>(delegate.reduce(seed, (result, element) -> accumulator.apply(element, result)));
     }
 
@@ -83,9 +83,9 @@ class StyxCoreObservable<T> implements StyxObservable<T> {
         return future;
     }
 
-    private static <U> Observable<? extends U> toObservable(StyxObservable<U> styxObservable) {
+    private static <T> Observable<T> toObservable(StyxObservable<T> styxObservable) {
         return styxObservable instanceof StyxCoreObservable
-                ? ((StyxCoreObservable<U>) styxObservable).delegate
+                ? ((StyxCoreObservable<T>) styxObservable).delegate
                 : toObservable(styxObservable.asCompletableFuture());
     }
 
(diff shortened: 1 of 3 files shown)


components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
```diff
@@ -103,4 +83,21 @@ public StyxCoreObservable(CompletionStage<T> future) {
         return future;
     }
 
+    private static <U> Observable<? extends U> toObservable(StyxObservable<U> styxObservable) {
```
dvlato: Do you see any advantage to the return type? It's usually considered a bad practice to include wildcard types in return types as, rather than providing more flexibility to the caller, it's just makes the code more bloated.

If we pass a StyxObservable<U> we are always going to return an Observable<U> and not any subclass of U.

 BTW,  T seems better than U if we want to be consistent.

components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
```diff
@@ -62,31 +50,23 @@ public StyxCoreObservable(CompletionStage<T> future) {
         return new StyxCoreObservable<T>(Observable.error(cause));
     }
 
-    public <U> StyxObservable<U> map(Function<T, U> transformation) {
+    public <U> StyxObservable<U> map(Function<? super T, ? extends U> transformation) {
         return new StyxCoreObservable<>(delegate.map(transformation::apply));
     }
 
-    public <U> StyxObservable<U> flatMap(Function<T, StyxObservable<U>> transformation) {
+    public <U> StyxObservable<U> flatMap(Function<? super T, ? extends StyxObservable<? extends U>> transformation) {
```
mikkokar: I am skeptical about `? extends StyxObservable<? extends U>` bit.

Due to some technical limitations, we don't intend `StyxObservable` to be extendable (although it is implemented by a package private subclass).

But the above annotation suggests that it is OK to provide a mapping function that provides `StyxObservable` subclasses.

Perhaps the function signature needs to be just: 

`Function<? super T, StyxObservable<? extends U>> transformation`

In fact this remark probably applies to all instances where we see this pattern repeating.

components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
```diff
@@ -62,31 +50,23 @@ public StyxCoreObservable(CompletionStage<T> future) {
         return new StyxCoreObservable<T>(Observable.error(cause));
     }
 
-    public <U> StyxObservable<U> map(Function<T, U> transformation) {
+    public <U> StyxObservable<U> map(Function<? super T, ? extends U> transformation) {
```
dvlato: I'd rather change the return type name to R (instead of U) following Java 8 (e.g,  https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html) and Observable's convention.  
Maybe just copy the generic names from the Observable interface for the same methods?

components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
dvlato: I think StyxObservable is an interface, isn't it? :)  In this case, both signatures would accept any implementation of 'StyxObservable' (mmm, provided the declared type was for StyxObservable?) , but I think the one used by Kyle/rxJava would allow to pass  Function<StyxCoreObservable, SecondParam> whereas   your version only allows Function<StyxObservable, SecondParam>     .

components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
dvlato: I didn't know that was in the Java tutorials! I  thought it was usually quoted by the 'Effective Java' book.

components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
kvosper: Upon making the suggested change to `flatMap`, a lot of the scala Specs stop compiling. It's not completely clear to me why, but I think there may be some significance to the new signature style (as seen in Rx) beyond what we currently realise.

components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
kvosper: This one is easily fixed as it is private and doesn't break anything. As for which letters to use for type parameters, both Rx and java streams use `T` for input and `R` for output.

components/api/src/main/java/com/hotels/styx/api/StyxCoreObservable.java
kvosper: I have investigated a little bit an the problem with `Function<? super T, StyxObservable<? extends U>>` is that it will only accept a function whose return type is **verbatim** `StyxObservable<? extends U>`.

This means that a function that returns `StyxObservable<B>` where `B` is a subtype of `U` will not be allowed.

For this reason we need the `Function<? super T, ? extends StyxObservable<? extends U>>`.

dvlato: Ah, of course,   `<StyxObservable<Integer>>` is not compatible with `<StyxObservable<? extends Number>>` due to the type invariance of the first parametrized type. That´s what happens (to me) when reviewing code outside the IDE...

Issue #208, opened 2018-07-06
Add a metrics endpoint that allows all metrics starting with a given string to be found.


kvosper:
## The problem
Note: this is a follow-up to #182 which must be completed first.

Add a metrics endpoint that allows all metrics starting with a given string to be found. This will differ from the existing behaviour by treating the metric name as a simple string, i.e. treating the `.` delimiters the same as any other character.

## Detailed description
For example:
`/admin/metrics/jvm?prefix=mem` would return everything under `jvm.memory`.

## Possible enhancement to this issue
Do we want more thorough searching, e.g. using regex?
Or at least to find the specified string anywhere in the metric?
This could allow us to search for metrics when we don't know the exact name, or when we want anything related to a particular aspect of the system, without caring where the word appears in the metric name.

## Acceptance criteria
* If `prefix` query parameter is supplied to `/metrics/` endpoint, return all metrics starting with the provided string
* If there are no such metrics, just return an empty JSON object

PR #221, opened 2018-07-31
Metrics endpoint enhancement

kvosper:
Implementing issue #208 

I've extended the scope a little - rather than only looking at the start of the metric name for a search term, we're looking anywhere in the string.

This is more useful as someone looking for general metrics may know what word they need to find in the metric name, but not the root location.

Note: This does not remove the ability to restrict metrics to a certain root name.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
@@ -74,9 +74,7 @@ public class MetricsHandler extends JsonHandler<MetricRegistry> {
 
         return metricRequest.fullMetrics()
                 ? super.handle(request)
-                : just(restrictedMetricsResponse(metricRequest)
-                .build()
-                .toStreamingResponse());
+                : just(restrictedMetricsResponse(metricRequest).build().toStreamingResponse());
     }
 
     private FullHttpResponse.Builder restrictedMetricsResponse(MetricRequest request) {
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectWriter;
 import com.hotels.styx.api.HttpRequest;
 import com.hotels.styx.api.HttpResponse;
+import com.hotels.styx.api.messages.FullHttpResponse;
 import com.hotels.styx.api.metrics.codahale.CodaHaleMetricRegistry;
 import com.hotels.styx.common.MapStream;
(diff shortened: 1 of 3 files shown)


components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
```diff
@@ -23,6 +23,7 @@
 import com.fasterxml.jackson.databind.ObjectWriter;
```
mikkokar: Does this not need any additions to our end user documentation?

kvosper: I will add some.

docs/user-guide/metrics.md
```diff
@@ -21,6 +21,15 @@ For example `http://<styx-host>/admin/metrics/requests.response` would return
...
     
+## Searching Metrics for string
+
+To see all metrics containing a given string, a query parameter can be provided.
+This works with both specific metric roots and the set of all metrics:
+
+`http://<styx-host>/admin/metrics?search=<term>`
+
+`http://<styx-host>/admin/metrics/<metric-name>?search=<term>`
+
```
mikkokar: My immediate question is "what is `<term>`"? Please elaborate this. 

Also, I would use more straightforward language:

> Use a `search` query parameter to filter for metrics names matching a given string. For example `search=count` only shows metrics whose name contains `count`. The filtering is applied to the results of the metrics query.

mikkokar: Also I am being very pedantic now but `filter` would be more accurate name for the query parameter, instead of `search`. WDYT?

components/proxy/src/main/java/com/hotels/styx/admin/handlers/MetricsHandler.java
```diff
@@ -68,37 +70,35 @@ public MetricsHandler(CodaHaleMetricRegistry metricRegistry, Optional<Duration>
...
-        }
+        String body = serialise(searched, request.prettyPrint);
 
-        return Optional.empty();
+        return response(OK)
+                .body(body, UTF_8)
+                .disableCaching();
+    }
+
+    private static <K, V> Map<K, V> filter(Map<K, V> map, BiPredicate<K, V> predicate) {
```
mikkokar: We need to refactor this code.

kvosper: In what way?

mikkokar: Well, this handler should perform three logical steps:

1. Query for the metrics
2. Filter results
3. Serialise and respond

But these logical steps disappear within the layers of abstraction. I'll find this class rather confusing considering its really simple responsibility.

mikkokar: Anyway, I will approve. Let's merge and get back to this another time.

PR #224, opened 2018-08-03
Fixes #222: A memory leak in DashboardData.

### LINKED ISSUE #222: Memory leak in styx V 0.14.0
## The problem
We are facing a memory leak in Styx V-0.14.0.

## Detailed description
After investigation we found that the root cause of this issue is related to the dashboard each time we trigger the dashborad update event the code register new Origins objects to the google guava eventbus and it dose not unregister old objects see
com.hotels.styx.admin.dashboard.DashboardData$Backend constructer
with time this behavior fill the memory with unused objects that not eligible for the GC 

## Acceptance criteria
When trigger the dashboard update event before registering new origins to the eventbus, old Origins MUST unregistered.

mikkokar:
Fixes #222, a memory leak in dashboard data supplier.

The leak was caused by `DashobardData.Origin` objects being continuously registered to the message bus, but never unregistered.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/admin/dashboard/DashboardData.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/dashboard/DashboardData.java
@@ -25,12 +25,13 @@ import com.google.common.collect.Lists;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import com.hotels.styx.Version;
-import com.hotels.styx.api.client.OriginsSnapshot;
 import com.hotels.styx.api.client.OriginsChangeListener;
+import com.hotels.styx.api.client.OriginsSnapshot;
 import com.hotels.styx.api.metrics.MetricRegistry;
 import com.hotels.styx.api.service.BackendService;
 import com.hotels.styx.api.service.spi.Registry;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -89,6 +90,10 @@ public class DashboardData {
         return System.currentTimeMillis();
     }
 
+    void unregister() {
+        this.downstream.unregister();
+    }
+
     /**
      * Styx-related data.
      */
@@ -112,7 +117,7 @@ public class DashboardData {
         }
 
         @JsonProperty("uptime")
-        public String uptime() {
+        String uptime() {
             return uptimeGauge == null ? null : uptimeGauge.getValue();
         }
 
@@ -130,7 +135,7 @@ public class DashboardData {
         private final Supplier<Map<String, Integer>> responsesSupplier;
 
         private Downstream() {
-            updateBackendsFromRegistry();
+            this.backends = updateBackendsFromRegistry();
 
             this.responsesSupplier = new ResponseCodeSupplier(metrics, COUNTER, "origins.response.status", false);
             backendServicesRegistry.addListener(this);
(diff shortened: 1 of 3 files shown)


components/proxy/src/main/java/com/hotels/styx/admin/dashboard/DashboardData.java
```diff
@@ -166,14 +171,22 @@ Backend backend(String backendId) {
...
         public void onChange(Registry.Changes<BackendService> changes) {
-            updateBackendsFromRegistry();
+            this.backends = updateBackendsFromRegistry();
         }
 
-        private void updateBackendsFromRegistry() {
-            this.backends = stream(backendServicesRegistry.get().spliterator(), false)
+        private List<Backend> updateBackendsFromRegistry() {
+            if (this.backends != null) {
+                this.backends.forEach(Backend::unregister);
```
kvosper: this could just be
```
if (this.backends != null) {
  unregister();
}
```

components/proxy/src/main/java/com/hotels/styx/admin/dashboard/DashboardData.java
```diff
@@ -206,6 +225,13 @@ private Backend(BackendService application) {
             this.responsesSupplier = new ResponseCodeSupplier(metrics, METER, prefix, true);
         }
 
+        void unregister() {
+            registeredOrigins.forEach(origin -> {
```
kvosper: Why do we have a new field `registeredOrigins` that contains the same list as the existing field `List<Origin> origin`?

components/proxy/src/test/java/com/hotels/styx/admin/dashboard/DashboardDataTest.java
```diff
@@ -292,6 +295,25 @@ public void providesOriginConnectionPoolData() {
...
+        backendServicesRegistry.add(application("test", origin("test-01", "localhost", 9090)));
+
+        DashboardData dashbaord = new DashboardData(metricRegistry, backendServicesRegistry, "styx-prod1-presentation-01", new Version("releaseTag"), eventBus);
+
+        // Twice for each backend. One during backend construction, another from BackendServicesRegistry listener callback.
+        verify(eventBus, times(4)).register(any(DashboardData.Origin.class));
+
+        dashbaord.unregister();
+
+        verify(eventBus, times(4)).register(any(DashboardData.Origin.class));
```
kvosper: Shouldn't this be verifying that the origins were unregistered, not registered again?

mikkokar: Good point.

components/proxy/src/main/java/com/hotels/styx/admin/dashboard/DashboardData.java
mikkokar: This is so that we unregister only those origins that actually have been registered.
I noticed that in some test cases (DashboardDataSupplier) the `origins` list contained different origins to those from `registeredOrigins`. But this could have been just a unusual or unrealistic dependency injection in the test itself.

mikkokar:
Note, this is required as a bug fix for styx 0.7.n branches only. We will not cherry pick to styx 1.0 branch as it will have a brand new implementation for dashboard data supplier.

mikkokar:
Will cherry pick to 1.0 as the refactoring is likely to be delayed.

PR #230, opened 2018-08-06
Tidy up cookies API

mikkokar:
- Use `java.util.Objects` instead of Guava.
- Replace Guava-generated toString method.
- Restrict visibility of `ServerCookieEncoder` class

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/cookies/ResponseCookie.java
+++ b/components/api/src/main/java/com/hotels/styx/api/cookies/ResponseCookie.java
@@ -26,7 +26,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import static com.google.common.base.Objects.toStringHelper;
 import static io.netty.handler.codec.http.cookie.Cookie.UNDEFINED_MAX_AGE;
 import static java.util.Objects.hash;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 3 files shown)


components/api/src/main/java/com/hotels/styx/api/cookies/ResponseCookie.java
```diff
@@ -224,20 +223,21 @@ public boolean equals(Object o) {
...
-        return toStringHelper(this)
-                .add("name", name)
-                .add("value", value)
-                .add("domain", domain)
-                .add("maxAge", maxAge)
-                .add("path", path)
-                .add("httpOnly", httpOnly)
-                .add("secure", secure)
-                .toString();
+        final StringBuilder sb = new StringBuilder("ResponseCookie{");
```
kvosper: This could be simplified to:
```
return new StringBuilder("ResponseCookie{")
                .append("name='").append(name).append('\'')
                .append(", value='").append(value).append('\'')
                .append(", domain='").append(domain).append('\'')
                .append(", maxAge=").append(maxAge)
                .append(", path='").append(path).append('\'')
                .append(", httpOnly=").append(httpOnly)
                .append(", secure=").append(secure)
                .append('}').toString();
```
or simply
```
return "ResponseCookie{" +
                "name='" + name + '\'' +
                ", value='" + value + '\'' +
                ", domain='" + domain + '\'' +
                ", maxAge=" + maxAge +
                ", path='" + path + '\'' +
                ", httpOnly=" + httpOnly +
                ", secure=" + secure +
                '}';
```

mikkokar: Addressed & pushed changes.

PR #232, opened 2018-08-06
Remove MediaType from public methods in API module

kvosper:


components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -445,8 +444,8 @@ public Builder addHeader(CharSequence name, Object value) {
             return this;
         }
 
-        public Builder contentType(MediaType mediaType) {
-            return addHeader(CONTENT_TYPE, mediaType);
+        public Builder contentType(String contentType) {
+            return addHeader(CONTENT_TYPE, contentType);
```
mikkokar: I think we should remove `contentType` methods for now. Seems bit pointless if all it does is to set the header with an arbitrary string. The consumers can just set the Content-Type header directly.

We can add them in if or when we'll add a Styx MediaType abstraction.

components/api/src/main/java/com/hotels/styx/api/FullHttpResponse.java
```diff
@@ -458,8 +457,8 @@ public Builder addHeader(CharSequence name, Object value) {
             return this;
         }
 
-        public Builder contentType(MediaType mediaType) {
-            return addHeader(CONTENT_TYPE, mediaType);
+        public Builder contentType(String contentType) {
+            return addHeader(CONTENT_TYPE, contentType);
```
mikkokar: See my comment to the `HttpResponse` message.

PR #240, opened 2018-08-23
Simplify package structure for 1.0 styx-api module

mikkokar:
Reduces number of java packages in styx-api module. Results in a more consistent and less cluttered API structure.

* All HTTP message & interceptor related definitions are now in the `com.hotels.styx.api` package.

* Any SPI extensions go under `com.hotels.styx.api.extension.<name>.spi`. However to minimise impact on existing plugins the plugin definitions still live in `com.hotels.styx.api.plugins`.

This PR also moves some unnecessary classes away from the API module.


components/api/src/main/java/com/hotels/styx/api/extension/service/spi/AbstractStyxService.java
```diff
@@ -46,7 +40,7 @@
  */
 public abstract class AbstractStyxService implements StyxService {
     private final String name;
-    private final AtomicReference<StyxServiceStatus> status = new AtomicReference<>(CREATED);
+    private final AtomicReference<StyxServiceStatus> status = new AtomicReference<>(StyxServiceStatus.CREATED);
```
kvosper: Please reintroduce the static imports.

mikkokar: Fixed.

Issue #235, opened 2018-08-21
Unused method `isHttps` in StyxHttpClient.


mikkokar:
## The problem

StyxHttpClient has a unused public `isHttps` method. Removing the method allows us also to drop a dependency to `BackendService` class.

## Acceptance criteria

* There is no `isHttps` method in `StyxHttpClient`.
* `StyxHttpClient.Builder` no longer takes a `BackendService` as a constructor argument. It shall take an application ID instead.

VivianLopes:
We cannot eliminate `BackendService` from `StyxHttpClient` just by deleting `StyxHttpClient.isHttps` as it is also used to get `stickySessionConfig()`.

mikkokar:
Hi @VivianLopes, just pass in the SticySessionConfig instead of BackendService ;-).

PR #241, opened 2018-08-23
#235 - Removed unused method

VivianLopes:
Removed unused `isHttps()` and removed `BackendService` constructor argument from `StyxHttpClient.Builder`

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
+++ b/components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
@@ -77,7 +77,7 @@ public final class StyxHttpClient implements HttpClient {
     private final StickySessionConfig stickySessionConfig;
 
     private StyxHttpClient(Builder builder) {
-        this.id = requireNonNull(builder.applicationId);
+        this.id = requireNonNull(builder.backendServiceId);
 
         this.stickySessionConfig = requireNonNull(builder.stickySessionConfig);
 
@@ -107,8 +107,8 @@ public final class StyxHttpClient implements HttpClient {
      *
      * @return a new builder
      */
-    public static Builder newHttpClientBuilder(Id applicationId) {
-        return new Builder(applicationId);
+    public static Builder newHttpClientBuilder(Id backendServiceId) {
+        return new Builder(backendServiceId);
     }
 
     private static boolean isError(HttpResponseStatus status) {
@@ -335,7 +335,7 @@ public final class StyxHttpClient implements HttpClient {
      */
     public static class Builder {
 
-        private final Id applicationId;
+        private final Id backendServiceId;
         private MetricRegistry metricsRegistry = new CodaHaleMetricRegistry();
         private List<RewriteRule> rewriteRules = emptyList();
         private RetryPolicy retryPolicy = new RetryNTimes(3);
@@ -345,8 +345,8 @@ public final class StyxHttpClient implements HttpClient {
         private String originsRestrictionCookieName;
         private StickySessionConfig stickySessionConfig = stickySessionDisabled();
 
-        public Builder(Id applicationId) {
-            this.applicationId = requireNonNull(applicationId);
+        public Builder(Id backendServiceId) {
+            this.backendServiceId = requireNonNull(backendServiceId);
         }
 
         public Builder stickySessionConfig(StickySessionConfig stickySessionConfig) {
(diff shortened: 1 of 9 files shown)


components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -95,22 +97,19 @@ private StyxHttpClient(Builder builder) {
...
-    }
-
     /**
      * Create a new builder.
      *
      * @return a new builder
      */
-    public static Builder newHttpClientBuilder(BackendService backendService) {
-        return new Builder(backendService);
+    public static Builder newHttpClientBuilder(Id applicationId) {
```
kvosper: I think we should call the parameter `backendServiceId` for consistency.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -336,19 +335,27 @@ public String toString() {
...
         private RetryPolicy retryPolicy = new RetryNTimes(3);
         private LoadBalancer loadBalancer;
         private boolean contentValidation;
         private OriginStatsFactory originStatsFactory;
         private String originsRestrictionCookieName;
+        private StickySessionConfig stickySessionConfig = stickySessionDisabled();
 
-        public Builder(BackendService backendService) {
-            this.backendService = checkNotNull(backendService);
+        public Builder(Id applicationId) {
```
kvosper: I think we should call the parameter and field `backendServiceId` for consistency.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -336,19 +335,27 @@ public String toString() {
...
         private LoadBalancer loadBalancer;
         private boolean contentValidation;
         private OriginStatsFactory originStatsFactory;
         private String originsRestrictionCookieName;
+        private StickySessionConfig stickySessionConfig = stickySessionDisabled();
 
-        public Builder(BackendService backendService) {
-            this.backendService = checkNotNull(backendService);
+        public Builder(Id backendServiceId) {
+            this.backendServiceId = checkNotNull(backendServiceId);
```
kvosper: Please use Java 8's `requireNonNull` instead of Guava's `checkNotNull`. Same comment for `stickySessionConfig(...)`

dvlato: See ticket https://github.com/HotelsDotCom/styx/issues/243

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -336,17 +333,24 @@ public String toString() {
...
         private String originsRestrictionCookieName;
+        private StickySessionConfig stickySessionConfig = stickySessionDisabled();
 
-        public Builder(BackendService backendService) {
-            this.backendService = checkNotNull(backendService);
+        public Builder(Id backendServiceId) {
+            this.backendServiceId = requireNonNull(backendServiceId);
+        }
+
+        public Builder stickySessionConfig(StickySessionConfig stickySessionConfig) {
```
mikkokar: Sticky session config is not called from the client factory.

That needs to be added in, and ideally testing hardened to catch this issue.

components/client/src/test/integration/scala/com/hotels/styx/client/StickySessionSpec.scala
```diff
@@ -97,9 +98,13 @@ class StickySessionSpec extends FunSuite with BeforeAndAfter with ShouldMatchers
 
   def stickySessionStrategy(activeOrigins: ActiveOrigins) = new StickySessionLoadBalancingStrategy(activeOrigins, roundRobinStrategy(activeOrigins))
 
+
   test("Responds with sticky session cookie when STICKY_SESSION_ENABLED=true") {
+    val stickySessionConfig = StickySessionConfig.newStickySessionConfigBuilder().timeout(100, TimeUnit.SECONDS).build()
+
     val client = newHttpClientBuilder(backendService.id)
       .loadBalancer(stickySessionStrategy(activeOrigins(backendService)))
+      .stickySessionConfig(stickySessionConfig)
```
dvlato: It's not clear how to do this properly/generally in our case,  but we might want to test that the required 'Builder' methods (in this case, stickySessionConfig() ) have been invoked in our UnitTests instead of only during the EndToEndTests .  Ideally, we would guarantee this at compile time or at the very least at runtime.
 This is a clear short-coming of our usage of the builder pattern, we cannot guarantee that the object has been constructed with all the required parameters...

PR #256, opened 2018-08-30
Fix intermittently failing ChunkedDownloadSpec

mikkokar:
Fixes an intermittently failing end-to-end test.

Root cause: Failing assertion was based on origin metrics. However the origin was shared with two other tests which influenced the asserted metric values.

The fix is to add a new backend service that is only used in the failing test, preventing other tests to influence the origin metrics, thus affecting the assertions.

### DIFF
--- a/system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/ChunkedDownloadSpec.scala
+++ b/system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/ChunkedDownloadSpec.scala
@@ -54,6 +54,7 @@ class ChunkedDownloadSpec extends FunSpec
   }
 
   override protected def afterAll(): Unit = {
+    println("metrics: " + styxServer.metricsSnapshot)
     originOneServer.stopAsync().awaitTerminated()
     super.afterAll()
   }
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -92,7 +92,7 @@
     <netty-tcnative.classifier>linux-x86_64</netty-tcnative.classifier>
 
     <!-- General Versions -->
-    <guava.version>16.0.1</guava.version>
+    <guava.version>18.0</guava.version>
```
dvlato: Should this change be included in this PR?

mikkokar: Nope

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/ChunkedDownloadSpec.scala
```diff
@@ -44,12 +46,14 @@ class ChunkedDownloadSpec extends FunSpec
   with TestClientSupport
   with Eventually {
 
-  val (originOne, originOneServer) = originAndCustomResponseWebServer("NettyOrigin")
+  val (originOne, originOneServer) = originAndCustomResponseWebServer("NettyOrigin-01")
```
dvlato: The solution is fine but I would expect that we just create the origin in  each test / setup method if they can't be reused?

mikkokar: It is the styx server that would have to be created in each test/setup cycle.

dvlato: That's also possible, but I was just thinking of limiting the scope of the origins ( they can be created with ,e.g., a random name in each test). But since , as you say, there's only one Styx instance it makes sense to set all of them up in the fixture.

PR #260, opened 2018-08-31
Changelog for releases up to styx 0.7.9

dvlato:


mikkokar:
The Travis failure was caused by https://github.com/HotelsDotCom/styx/pull/256, which is already fixed in the 1.0 release. Therefore merging.

PR #265, opened 2018-09-03
Issue #264: Remove unnecessary calls to `freePort`.

### LINKED ISSUE #264: Eliminate unnecessary `freePort` calls.
## The problem

There are some unit tests and e2e that are unnecessarily allocating a free TCP port using a `freePort` method.

Remove these invocations and replace them with constant port numbers.

mikkokar:
Fixes #264.


system-tests/e2e-suite/src/test/java/com/hotels/styx/metrics/ProtocolMetricsTest.java
```diff
@@ -85,19 +71,18 @@ public void stopStyx() {
...
     public void recordsServerSideHttp() {
         styxServer = new StyxServer.Builder()
-                .addRoute("/", nonSslOrigin.port())
+                .addRoute("/", origin.port())
                 .start();
 
         FullHttpResponse response = doGet("/");
 
-        assertThat(response.status(), is(OK));
+        assertThat(response.status().code(), is(OK.code()));
```
kvosper: Why was this necessary? If they don't match, it may indicate that the classes are not the same, e.g. Styx `HttpResponseStatus` vs Netty `HttpResponseStatus`.

components/proxy/src/test/java/com/hotels/styx/proxy/StyxProxyTest.java
```diff
@@ -77,15 +76,14 @@ public void startsAndStopsAServer() {
     public void startsServerWithHttpConnector() {
         HttpInterceptor echoInterceptor = (request, chain) -> textResponse("Response from http connector");
 
-        int port = freePort();
         HttpServer server = NettyServerBuilder.newBuilder()
-                .setHttpConnector(connector(port))
+                .setHttpConnector(connector(0))
```
dvlato: I would say the code is a bit easier to write if we use freeport for the cases with  '0' and WireMockServer.dynamic*Port()  instead of having to find what's the correct way to get a random port in each case.  That doesn't mean we have to keep the 'port' variable used later...

mikkokar: Hi @dvlato. `freePort` is problematic because it can easily cause port conflicts in parallel environments. Because it is susceptible for race conditions it should be avoided.

system-tests/e2e-suite/src/test/java/com/hotels/styx/metrics/ProtocolMetricsTest.java
mikkokar: Ok. Importing now Styx `HttpResponseStatus` instead of Netty.

PR #271, opened 2018-09-06
Improve Javadocs for styx-api module

mikkokar:
This PR also adds package level API docs for `com.hotels.styx.api` and improves existing Javadocs for HTTP message objects.

BTW, the Javadocs can be generated locally with:

```
mvn javadoc:javadoc
```

Or specifically for the *styx-api* module:

```
mvn -f components/api/pom.xml javadoc:javadoc
```

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/ContentOverflowException.java
+++ b/components/api/src/main/java/com/hotels/styx/api/ContentOverflowException.java
@@ -16,7 +16,12 @@
 package com.hotels.styx.api;
 
 /**
- * An exception that indicates that a content-aggregation operation exceeded the maximum permitted amount of content.
+ * Thrown when HTTP content aggregation fails because received more data than excepted.
+ * <p>
+ * A streaming HTTP message can  be aggregated and converted (aggregated) to a
+ * full message. Normally the aggregator methods ({@code HttpRequest.toFullRequest}, {@code HttpResponse.toFullResponse})
+ * receive data from the network. The {@code ContentOverflowException}
+ * is thrown when more data is received than the aggregator expected.
  */
 public class ContentOverflowException extends RuntimeException {
     public ContentOverflowException(String message) {
--- a/components/api/src/main/java/com/hotels/styx/api/ContentOverflowException.java
+++ b/components/api/src/main/java/com/hotels/styx/api/ContentOverflowException.java
@@ -16,12 +16,14 @@
 package com.hotels.styx.api;
 
 /**
- * Thrown when HTTP content aggregation fails because received more data than excepted.
+ * Indicates a content aggregation failure because an aggregator function received
+ * more data than expected.
  * <p>
- * A streaming HTTP message can  be aggregated and converted (aggregated) to a
- * full message. Normally the aggregator methods ({@code HttpRequest.toFullRequest}, {@code HttpResponse.toFullResponse})
- * receive data from the network. The {@code ContentOverflowException}
- * is thrown when more data is received than the aggregator expected.
+ * A streaming HTTP message can be aggregated to a full message.
+ * The aggregator methods {@link HttpRequest#toFullRequest}, and
+ * {@link HttpResponse#toFullResponse} consume data from a network socket or
+ * some other unpredictable source and emit this exception when more data is
+ * received than expected.
  */
 public class ContentOverflowException extends RuntimeException {
     public ContentOverflowException(String message) {
--- a/components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
+++ b/components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
@@ -55,7 +55,34 @@ import static java.util.stream.Collectors.toList;
 import static java.util.stream.Stream.concat;
 
 /**
- * HTTP request with a fully aggregated/decoded body.
+ * An immutable HTTP request object containing a finite size content body in full.
+ * <p>
+ * A {@code FullHttpRequest} is useful for HTTP request messages with
+ * finite body content such as PUT or POST for creating or modifying a RESTful
+ * object.
+ * <p>
+ * FullHttpRequest is immutable. Once created, it is impossible to modify any
+ * request attributes, including body content. However the request can
+ * be transformed to another via {@code FullHttpRequest.Builder}.
+ * <p>
+ * A FullHttpRequest is created via {@code FullHttpRequest.Builder}. A new builder
+ * can be obtained by a call to following static methods:
+ *
+ * <ul>
+ *     <li>{@code get}</li>
+ *     <li>{@code head}</li>
+ *     <li>{@code post}</li>
+ *     <li>{@code put}</li>
+ *     <li>{@code delete}</li>
+ *     <li>{@code patch}</li>
+ * </ul>
+ *
+ * A builder can also be created with one of the {@code Builder} constructors.
+ *
+ * A special method {@code newBuilder} creates a prepopulated {@code Builder}
+ * from the current request object. It is useful for transforming a request
+ * to another one my modifying one or more of its attributes.
+ *
  */
 public class FullHttpRequest implements FullHttpMessage {
     private final Object id;
--- a/components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
+++ b/components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
@@ -57,7 +57,7 @@ import static java.util.stream.Stream.concat;
 /**
  * An immutable HTTP request object including full body content.
  * <p>
- * A {@link FullHttpRequest} is useful for request messages with a
+ * A {@link FullHttpRequest} is useful for requests with a
  * finite body content, such as when PUT or POST are used to create or
  * modify a RESTful object.
  * <p>
--- a/components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
+++ b/components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
@@ -55,17 +55,13 @@ import static java.util.stream.Collectors.toList;
 import static java.util.stream.Stream.concat;
 
 /**
- * An immutable HTTP request object containing a finite size content body in full.
+ * An immutable HTTP request object including full body content.
  * <p>
- * A {@code FullHttpRequest} is useful for HTTP request messages with
- * finite body content such as PUT or POST for creating or modifying a RESTful
- * object.
+ * A {@link FullHttpRequest} is useful for request messages with a
+ * finite body content, such as when PUT or POST are used to create or
+ * modify a RESTful object.
  * <p>
- * FullHttpRequest is immutable. Once created, it is impossible to modify any
- * request attributes, including body content. However the request can
- * be transformed to another via {@code FullHttpRequest.Builder}.
- * <p>
- * A FullHttpRequest is created via {@code FullHttpRequest.Builder}. A new builder
+ * A FullHttpRequest is created via {@link FullHttpRequest.Builder}. A new builder
  * can be obtained by a call to following static methods:
  *
  * <ul>
@@ -79,10 +75,10 @@ import static java.util.stream.Stream.concat;
  *
  * A builder can also be created with one of the {@code Builder} constructors.
  *
- * A special method {@code newBuilder} creates a prepopulated {@code Builder}
- * from the current request object. It is useful for transforming a request
- * to another one my modifying one or more of its attributes.
- *
+ * FullHttpRequest is immutable. Once created it cannot be modified.
+ * However a request can be transformed to another using the {@link this#newBuilder}
+ * method. It creates a new {@link Builder} with all message properties and
+ * body content cloned in.
  */
 public class FullHttpRequest implements FullHttpMessage {
     private final Object id;
@@ -328,7 +324,7 @@ public class FullHttpRequest implements FullHttpMessage {
     /**
      * Converts this request into a streaming form (HttpRequest).
      * <p>
-     * Converts this request into a HttpRequest object which represents the HTTP request as a
+     * Converts this request into an HttpRequest object which represents the HTTP request as a
      * stream of bytes.
      *
      * @return A streaming HttpRequest object.
--- a/components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
+++ b/components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
@@ -106,7 +106,7 @@ public class FullHttpRequest implements FullHttpMessage {
      * Creates a request with the GET method.
      *
      * @param uri URI
-     * @return {@code this}
+     * @return {@link FullHttpRequest.Builder}
      */
     public static Builder get(String uri) {
         return new Builder(GET, uri);
@@ -116,7 +116,7 @@ public class FullHttpRequest implements FullHttpMessage {
      * Creates a request with the HEAD method.
      *
      * @param uri URI
-     * @return {@code this}
+     * @return {@link FullHttpRequest.Builder}
      */
     public static Builder head(String uri) {
         return new Builder(HEAD, uri);
@@ -126,7 +126,7 @@ public class FullHttpRequest implements FullHttpMessage {
      * Creates a request with the POST method.
      *
      * @param uri URI
-     * @return {@code this}
+     * @return {@link FullHttpRequest.Builder}
      */
     public static Builder post(String uri) {
         return new Builder(POST, uri);
@@ -136,7 +136,7 @@ public class FullHttpRequest implements FullHttpMessage {
      * Creates a request with the DELETE method.
      *
      * @param uri URI
-     * @return {@code this}
+     * @return {@link FullHttpRequest.Builder}
      */
     public static Builder delete(String uri) {
         return new Builder(DELETE, uri);
@@ -146,7 +146,7 @@ public class FullHttpRequest implements FullHttpMessage {
      * Creates a request with the PUT method.
      *
      * @param uri URI
-     * @return {@code this}
+     * @return {@link FullHttpRequest.Builder}
      */
     public static Builder put(String uri) {
         return new Builder(PUT, uri);
@@ -156,22 +156,32 @@ public class FullHttpRequest implements FullHttpMessage {
      * Creates a request with the PATCH method.
      *
      * @param uri URI
-     * @return {@code this}
+     * @return {@link FullHttpRequest.Builder}
      */
     public static Builder patch(String uri) {
         return new Builder(PATCH, uri);
     }
 
+    /**
+     * @return HTTP protocol version
+     */
     @Override
     public HttpVersion version() {
         return this.version;
     }
 
+    /**
+     * @return all HTTP headers as an {@link HttpHeaders} instance
+     */
     @Override
     public HttpHeaders headers() {
         return headers;
     }
 
+    /**
+     * @param name header name
+     * @return all values for a given HTTP header name or an empty list if the header is not present
+     */
     @Override
     public List<String> headers(CharSequence name) {
         return headers.getAll(name);
@@ -185,7 +195,7 @@ public class FullHttpRequest implements FullHttpMessage {
      * reference cannot be used to modify the message content.
      * <p>
      *
-     * @return Message body content
+     * @return message body content
      */
     @Override
     public byte[] body() {
@@ -199,8 +209,8 @@ public class FullHttpRequest implements FullHttpMessage {
      * The caller must ensure the provided charset is compatible with message content
      * type and encoding.
      *
-     * @param charset Charset used to decode message body.
-     * @return Message body as a String.
+     * @param charset Charset used to decode message body
+     * @return message body as a String
      */
     @Override
     public String bodyAs(Charset charset) {
@@ -210,17 +220,13 @@ public class FullHttpRequest implements FullHttpMessage {
     }
 
     /**
-     * Gets the unique ID for this request.
-     *
-     * @return request ID
+     * @return an unique request ID
      */
     public Object id() {
         return id;
     }
 
     /**
-     * Returns the HTTP method of this request.
-     *
      * @return the HTTP method
      */
     public HttpMethod method() {
@@ -228,18 +234,14 @@ public class FullHttpRequest implements FullHttpMessage {
     }
 
     /**
-     * Returns the requested URI (or alternatively, path).
-     *
-     * @return The URI being requested
+     * @return the request URL
      */
     public Url url() {
         return url;
     }
 
     /**
-     * Returns the requested path.
-     *
-     * @return the path being requested
+     * @return the request URL path component
      */
     public String path() {
         return url.path();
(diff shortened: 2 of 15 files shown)


components/api/src/main/java/com/hotels/styx/api/ContentOverflowException.java
```diff
@@ -16,7 +16,12 @@
 package com.hotels.styx.api;
 
 /**
- * An exception that indicates that a content-aggregation operation exceeded the maximum permitted amount of content.
+ * Thrown when HTTP content aggregation fails because received more data than excepted.
```
kvosper: 1. expected, not excepted.
2. What received more data than expected?

components/api/src/main/java/com/hotels/styx/api/ContentOverflowException.java
```diff
@@ -16,7 +16,12 @@
...
+ * <p>
+ * A streaming HTTP message can  be aggregated and converted (aggregated) to a
```
kvosper: 1. Double space between "can" and "be".
2. The word "aggregated" appears twice.

components/api/src/main/java/com/hotels/styx/api/ContentOverflowException.java
```diff
@@ -16,7 +16,12 @@
...
+ * full message. Normally the aggregator methods ({@code HttpRequest.toFullRequest}, {@code HttpResponse.toFullResponse})
```
kvosper: I don't think the `@code` syntax is correct here. Doesn't it have to be a `#` separating the method name from the class instead of a `.`?

components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
```diff
@@ -55,7 +55,34 @@
 import static java.util.stream.Stream.concat;
 
 /**
- * HTTP request with a fully aggregated/decoded body.
+ * An immutable HTTP request object containing a finite size content body in full.
+ * <p>
+ * A {@code FullHttpRequest} is useful for HTTP request messages with
```
kvosper: This seems like a bit of a run-on sentence and not completely clear. Maybe it would be better to write something like "A {@code FullHttpRequest} is useful for HTTP request messages with finite body content, such as when PUT or POST are used to create or modify a RESTful object".

components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
```diff
@@ -55,7 +55,34 @@
...
+ * finite body content such as PUT or POST for creating or modifying a RESTful
+ * object.
+ * <p>
+ * FullHttpRequest is immutable. Once created, it is impossible to modify any
+ * request attributes, including body content. However the request can
+ * be transformed to another via {@code FullHttpRequest.Builder}.
```
kvosper: This seems a little disconnected from the text that explains the `newBuilder` method.

components/api/src/main/java/com/hotels/styx/api/FullHttpResponse.java
```diff
@@ -43,7 +43,29 @@
 import static java.util.stream.Collectors.toList;
 
 /**
- * HTTP response with a fully aggregated/decoded body.
+ * An immutable HTTP response object containing a finite size content body in full.
+ * <p>
+ * A {@code FullHttpResponse} is useful for HTTP response messages with
+ * finite body content such as returning a REST object.
```
kvosper: "A {@code FullHttpResponse} is useful for HTTP response messages with finite body content such as a REST object." perhaps?

The act of returning a message is not a part of a message.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -63,7 +63,41 @@
 import static java.util.stream.Stream.concat;
 
 /**
- * HTTP request with a fully aggregated/decoded body.
+ * A HTTP request object with a byte stream body.
+ * <p>
+ * A {@code HttpRequest} is used in {@link HttpInterceptor} where each content
+ * chunk must be processed as they arrive. It is also useful for dealing with
```
kvosper: Singular/plural mismatch. Should be "each content chunk must be processed as **it** arrive**s**".

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -183,21 +217,45 @@ public static Builder patch(String uri, StyxObservable<ByteBuf> body) {
         return new Builder(PATCH, uri).body(body);
     }
 
+    /**
+     * Returns the HTTP protocol version.
+     * @return
```
kvosper: The `@return` needs some text after for it to be correct javadoc.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -183,21 +217,45 @@ public static Builder patch(String uri, StyxObservable<ByteBuf> body) {
...
+     */
     @Override
     public HttpVersion version() {
         return this.version;
     }
 
+    /**
+     * Returns all HTTP headers as {@link HttpHeaders} instance.
+     *
+     * @return {@link HttpHeaders} object.
```
kvosper: In Javadoc convention, the text after a tag does not end with a full stop (`.`). Presumably, this is because the text is not a full sentence.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -183,21 +217,45 @@ public static Builder patch(String uri, StyxObservable<ByteBuf> body) {
...
         return headers;
     }
 
+    /**
+     * Returns all values for a given HTTP header name.
+     *
+     * Returns an empty list if header is not present.
+     *
+     * @param name header name
+     * @return A list of all header names.
```
kvosper: In Javadoc convention, the text after a tag does not start with a capital letter. Presumably, this is because the text is not a full sentence.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
kvosper: Same comment applies to the rest.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
kvosper: Same comment applies to the rest.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
kvosper: Same comment applies to the rest.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -252,16 +310,22 @@ public boolean keepAlive() {
...
+     * If the protocol is HTTPS then the request is considered to be secure.
      *
-     * @return true if the request is transferred securely
+     * @return true if protocol is set to "https".
      */
     public boolean isSecure() {
         return secure;
     }
 
+    /**
```
kvosper: There should be an `@deprecated` tag in the Javadoc. "The @deprecated description in the first sentence should at least tell the user when the API was deprecated and what to use as a replacement." [Source](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@deprecated)

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -306,15 +370,36 @@ public InetSocketAddress clientAddress() {
...
+     * request line, headers, and content as this request.
+     *
+     * The content stream is aggregated asynchronously. The stream may be connected
+     * to a network socket or some other content producer. Once aggregated, a
+     * FullHttpRequest object is emitted on the returned {@link StyxObservable}.
+     *
+     * A sole {@code maxContentBytes} argument is a backstop defence against excessively
+     * long content streams. The {@code maxContentBytes} should be set to a sensible
+     * value according to your application requirements and heap size. When the content
+     * size stream exceeds the {@code maxContentBytes}, a @{link ContentOverflowException}
```
kvosper: It would be good to transform this sentence into an `@throws ContentOverflowException` tag.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -384,7 +469,7 @@ public String toString() {
     }
 
     /**
-     * Builder.
+     * A HTTP request builder.
```
kvosper: **An** HTTP...

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -399,19 +484,28 @@ public String toString() {
...
+        /**
+         * Creates a new {@link Builder} object with default attributes.
+         */
         public Builder() {
             this.url = Url.Builder.url("/").build();
             this.headers = new HttpHeaders.Builder();
             this.body = new StyxCoreObservable<>(Observable.empty());
         }
 
+        /**
```
kvosper: Missing parameter tags.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -399,19 +484,28 @@ public String toString() {
...
+         * Creates a new {@link Builder} with specified HTTP method and URI.
+         */
         public Builder(HttpMethod method, String uri) {
             this();
             this.method = requireNonNull(method);
             this.url = Url.Builder.url(uri).build();
             this.secure = url.isSecure();
         }
 
+        /**
```
kvosper: Missing parameter tags.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -561,17 +655,25 @@ public Builder method(HttpMethod method) {
             return this;
         }
 
+        /**
+         * Deprecated. Do not use in any new code.
```
kvosper: See comments above regarding `@deprecated`.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -63,7 +63,41 @@
...
- * HTTP request with a fully aggregated/decoded body.
+ * A HTTP request object with a byte stream body.
+ * <p>
+ * A {@code HttpRequest} is used in {@link HttpInterceptor} where each content
+ * chunk must be processed as they arrive. It is also useful for dealing with
+ * very large content sizes, and in situations where content size is not known
+ * upfront.
+ * <p>
+ * A {@code HttpRequest} object is immutable with respect to the request line
+ * attributes and HTTP headers. Once an instance is created, they cannot change.
```
kvosper: Should be "Once an instance is created, **it** cannot change."

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -53,7 +53,38 @@
 import static java.util.stream.Collectors.toList;
 
 /**
- * HTTP response with a fully aggregated/decoded body.
```
kvosper: See comments on `HttpRequest` regarding singular/plural mismatch.

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -98,43 +129,115 @@ public static Builder response(HttpResponseStatus status, StyxObservable<ByteBuf
         return new Builder(status).body(body);
     }
 
+    /**
+     * Returns a HTTP header value.
```
kvosper: "**An** HTTP"

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -98,43 +129,115 @@ public static Builder response(HttpResponseStatus status, StyxObservable<ByteBuf
...
+     * @param name header name
+     * @return An optional of value, or empty if not present.
+     */
     @Override
     public Optional<String> header(CharSequence name) {
         return headers.get(name);
     }
 
+    /**
+     * Returns all HTTP headers as {@link HttpHeaders} instance.
```
kvosper: Should be "Returns all HTTP headers as {@link HttpHeaders} instance**s**".

components/api/src/main/java/com/hotels/styx/api/package-info.java
```diff
@@ -0,0 +1,43 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+
+/**
+ * A programming interface for implementing Styx extensions. This includes
+ * {@code HttpInterceptor} interface for intercepting HTTP traffic, and classes
```
kvosper: I'd probably use "the interface {@code HttpInterceptor}" as a phrasing.

components/api/src/main/java/com/hotels/styx/api/package-info.java
```diff
@@ -0,0 +1,43 @@
...
+ * to represent HTTP Request and Response messages.
+ * <p></p>
+ *
+ * Styx exposes intercepted traffic to custom extensions by {@code HttpInterceptor}
+ * interface. A Styx plugin extends this interface to provide additional features
+ * for shaping or modifying proxied HTTP messages, including requests and responses.
+ * <p></p>
+ * Styx represents proxied life traffic as instances of {@code HttpRequest} and
```
kvosper: "life traffic" should say "live traffic".

pom.xml
```diff
@@ -505,6 +505,11 @@
   <build>
     <pluginManagement>
       <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-javadoc-plugin</artifactId>
```
dvlato: I haven't checked anyhting this, but I don't know why we need this change. We are already generating javadoc, aren't we? https://javadoc.io/doc/com.hotels.styx/styx-api/0.7.8 , https://javadoc.io/doc/com.hotels.styx/styx-api/0.7.9

components/api/src/main/java/com/hotels/styx/api/ContentOverflowException.java
mikkokar: I'm thinking to reword this description as:

```
/**
 * Indicates a content aggregation failure because more data was received than expected.
 * <p>
 * A streaming HTTP message can be aggregated to a full message. 
 * The aggregator methods {@link HttpRequest#toFullRequest}, and
 * {@link HttpResponse#toFullResponse} consume data from a network socket or
 * some other unpredictable source and emit this exception when more data is 
 * received than expected.
 */
```

components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
mikkokar: Okay. Sounds good.

It now reads:

```
/**
 * An immutable HTTP request object including full body content.
 * <p>
 * A {@link FullHttpRequest} is useful for HTTP request messages with
 * finite body content, such as when PUT or POST are used to create or
 * modify a RESTful object.
 * <p>
 * .... 
```

components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
mikkokar: @kvosper @dvlato @VivianLopes WDYT?

```
/**
 * An immutable HTTP request object including full body content.
 * <p>
 * A {@link FullHttpRequest} is useful for request messages with a
 * finite body content, such as when PUT or POST are used to create or
 * modify a RESTful object.
 * <p>
 * A FullHttpRequest is created via {@link FullHttpRequest.Builder}. A new builder
 * can be obtained by a call to following static methods:
 *
 * <ul>
 *     <li>{@code get}</li>
 *     <li>{@code head}</li>
 *     <li>{@code post}</li>
 *     <li>{@code put}</li>
 *     <li>{@code delete}</li>
 *     <li>{@code patch}</li>
 * </ul>
 *
 * A builder can also be created with one of the {@code Builder} constructors.
 *
 * FullHttpRequest is immutable. Once created it cannot be modified.
 * However a request can be transformed to another using the {@link this#newBuilder}
 * method. It creates a new {@link Builder} with all message properties and 
 * body content cloned in.
 */
```

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
mikkokar: Why? It returns just a single `HttpHeaders` instance.

components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
```diff
@@ -55,7 +55,30 @@
 import static java.util.stream.Stream.concat;
 
 /**
- * HTTP request with a fully aggregated/decoded body.
+ * An immutable HTTP request object including full body content.
+ * <p>
+ * A {@link FullHttpRequest} is useful for request messages with a
```
kvosper: It might be better to just say "requests" instead of "request messages", even though they both make sense. I think HTTP requests are typically just referred to that way.

Perhaps we could go with "A {@link FullHttpRequest} is useful for representing requests with ..."

components/api/src/main/java/com/hotels/styx/api/FullHttpResponse.java
```diff
@@ -43,7 +43,26 @@
 import static java.util.stream.Collectors.toList;
 
 /**
- * HTTP response with a fully aggregated/decoded body.
+ * An immutable HTTP response object including full body content.
+ * <p>
+ * A {@link FullHttpResponse} is useful for response messages with a
```
kvosper: If you agree with the comment for `FullHttpRequest`, the same applies here.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -63,7 +63,41 @@
 import static java.util.stream.Stream.concat;
 
 /**
- * HTTP request with a fully aggregated/decoded body.
+ * An HTTP request object with a byte stream body.
+ * <p>
+ * A {@code HttpRequest} is used in {@link HttpInterceptor} where each content
```
kvosper: > A**n** {@code HttpRequest}

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -63,7 +63,41 @@
...
+ * chunk must be processed as they arrive. It is also useful for dealing with
+ * very large content sizes, and in situations where content size is not known
+ * upfront.
+ * <p>
+ * A {@code HttpRequest} object is immutable with respect to the request line
```
kvosper: > A**n** {@code HttpRequest}

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -63,7 +63,41 @@
...
+ * attributes and HTTP headers. Once an instance is created, they cannot change.
+ *
+ * A {@code HttpRequest} body is a byte buffer stream that can be consumed
+ * as sequence of asynchronous events. Once consumed, stream is exhausted and
```
kvosper: > Once consumed, **the** stream is exhausted and

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -53,7 +53,38 @@
...
+ * A {@code HttpResponse} is used in {@link HttpInterceptor} where each content
+ * chunk must be processed as they arrive. It is also useful for dealing with
+ * very large content sizes, and in situations where content size is not known
+ * upfront.
+ * <p>
+ * A {@code HttpResponse} object is immutable with respect to the response
+ * attributes and HTTP headers. Once an instance is created, they cannot change.
+ *
+ * A {@code HttpResponse} body is a byte buffer stream that can be consumed
+ * as sequence of asynchronous events. Once consumed, stream is exhausted and
```
kvosper: > Once consumed, **the** stream is exhausted and

components/api/src/test/java/com/hotels/styx/api/HttpResponseTest.java
```diff
@@ -286,6 +286,16 @@ public void rejectsMultipleContentLength() {
...
     }
 
+    @Test
+    public void headerValues() {
+        HttpResponse r = response()
+                .addHeader("X-Styx", "14, 16")
+                .addHeader("X-Styx", "15")
+                .build();
+
+        System.out.println("headers: " + r.header("X-Styx"));
```
kvosper: Was this test intended to be committed? The name is bad, there are no assertions and it uses `System.out.println`.

components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
mikkokar: It would have to say:

> A {@link FullHttpRequest} represents requests with ...

mikkokar: I'd leave it like this:
``` 
 * A {@link FullHttpRequest} is useful for requests with a
 * finite body content, such as when PUT or POST are used to create or
 * modify a RESTful object.
```

components/api/src/main/java/com/hotels/styx/api/FullHttpResponse.java
mikkokar: Okay I have replaced "response messages" with just "responses".

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
mikkokar: FYI:

From: http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@param.

> The description begins with a lowercase letter if it is a phrase (contains no verb), or an uppercase letter if it is a sentence. End the phrase with a period only if another phrase or sentence follows it.

Clearly in this case it is a phrase instead of a sentence.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
mikkokar: Hmm... but it doesn't really *throw* that exception. It is just emitted thought the observable. In fact we never `toFullRequest` to throw anything except an `IllegalArgumentException`.

components/api/src/main/java/com/hotels/styx/api/FullHttpRequest.java
```diff
@@ -106,7 +106,7 @@
      * Creates a request with the GET method.
      *
      * @param uri URI
-     * @return {@code this}
+     * @return {@link FullHttpRequest.Builder}
```
kvosper: The previous text was intended to show that it was the same builder being returned, so that chained calls can be used to set the values. The new text only tells us the type, which we already know.

Is there some builder convention we should be following?

mikkokar: Even if the builder returned a new instance the method calls could still be chained.

We haven't agreed on any conventions. Feel free to suggest one.

Is this the only remaining issue that is blocking this PR?

Issue #274, opened 2018-09-07
Remove isSecure method from HttpRequest class


mikkokar:
## The problem

An HTTP protocol request doesn't have a concept of being "secure" or not. Anybody can just instantiate an HttpRequest instance with `secure` flag set without any practical effect or relation with reality.

Therefore `isSecure` method should be removed from the `HttpRequest` class. 

In case Styx core needs `isSecure` for something, this property should be exposed by an `HttpInterceptor.Context`.

## Acceptance criteria
* `isSecure` method is removed from `HttpRequest` (and `Response`?) class.
* If necessary, it is exposed in `HttpInterceptor.Context`.
* A request consumer can determine the HTTP protocol used ("http" or "https").

PR #277, opened 2018-09-10
Move isSecure flag to HttpInterceptor.Context

mikkokar:
Concludes the work for #274.
* Move `isSecure` attribute from HTTP request objects into `HttpInterceptor.Context`.

This has an implication to a request logging feature. The isSecure is no longer a request attribute, and therefore it no longer appears in request's textual representation. It is logged separately from the request, but only when Interceptor.Context is available. Therefore `isSecure` is logged for the inbound requests because the interceptor context is visible for the inbound request logger. But it is not logged for outbound requests because the context is not visible to the StyxHttpClient that logs the outbound requests.

### DIFF
--- a/components/api/src/test/java/com/hotels/styx/api/HttpRequestTest.java
+++ b/components/api/src/test/java/com/hotels/styx/api/HttpRequestTest.java
@@ -59,7 +59,6 @@ public class HttpRequestTest {
     @Test
     public void decodesToFullHttpRequest() throws Exception {
         HttpRequest streamingRequest = post("/foo/bar", body("foo", "bar"))
-                .secure(true)
                 .version(HTTP_1_0)
                 .header("HeaderName", "HeaderValue")
                 .cookies(requestCookie("CookieName", "CookieValue"))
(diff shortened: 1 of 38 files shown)


components/api/src/test/java/com/hotels/styx/api/HttpRequestTest.java
```diff
@@ -169,12 +166,6 @@ public void canModifyPreviouslyCreatedRequest() {
         assertThat(newRequest.headers(), hasItem(header("remove", "notanymore")));
     }
 
-    @Test
-    public void shouldDetermineIfRequestIsSecure() {
-        assertThat(get("https://hotels.com").build().isSecure(), is(true));
-        assertThat(get("http://hotels.com").build().isSecure(), is(false));
```
kvosper: I see an issue here. If we can still include the protocol when constructing the request, but then ignoring it, we are creating a misleading interface.

mikkokar: These URLS are in absolute form. But the URL is usually specified in an "origin form" in which case the URL protocol scheme isn't present. See RFC 7320, chapter *5.3.  Request Target* for more details. Also see chapter *5.5.  Effective Request URI* for related information about http schemes.

Issue #275, opened 2018-09-07
Remove the clientAddress field from HttpRequest


mikkokar:
## The problem

Client address doesn't belong to the HttpRequest (or Response) objects. Therefore the client address should be removed.

The styx core however logs the client address (in a pre-built interceptor). I would propose to add the client address inside `HttpInterceptor.Context` instead.

## Acceptance criteria
* HttpRequest no longer has a client address attribute.
* Plugins can obtain the client address from `HttpInterceptor.Context`.
* Styx still logs the client address as before.

PR #280, opened 2018-09-12
Remove client address

kvosper:
Implements #275


components/api/src/test/java/com/hotels/styx/api/extension/service/spi/MockContext.java
```diff
@@ -33,4 +37,9 @@ public void add(String key, Object value) {
     public boolean isSecure() {
         return false;
     }
+
+    @Override
+    public Optional<InetSocketAddress> clientAddress() {
+        return Optional.empty();
+    }
```
mikkokar: Hi @kvosper I notice `InetSocketAddress` is an Optional. Under what circumstances it is absent? Can you remind me? Apart from unit tests I cannot think of any situation where client address wouldn't be available in the Context.

components/common/src/main/java/com/hotels/styx/server/HttpInterceptorContext.java
```diff
@@ -17,18 +17,54 @@
...
+    private final boolean secure;
+    private final InetSocketAddress clientAddress;
+
+    /**
+     * Construct a new instance.
+     *
+     * @param secure true if the request was received via SSL
+     * @param clientAddress address that request came from, or null if not-applicable
+     */
+    public HttpInterceptorContext(boolean secure, InetSocketAddress clientAddress) {
```
mikkokar: Please back off this change to the constructor and the static factory method. This PR is now changing 53 files. Majority of them simply because of this unrelated constructor change. Let's keep this PR strictly to removing the client address. This way it is easier to justify and manage the changes, and to review the PRs.

Whether or not it was good idea back in the days to hide the constructor in favour of static factory method is another matter. If you feel strongly about it then please propose that in a separate pull request.

This change will also invalidate any plugins that use `HttpInterceptorContext.createa in unit tests.

components/api/src/test/java/com/hotels/styx/api/extension/service/spi/MockContext.java
kvosper: When we use the Styx client, there is a line `host.hostClient().handle(request, new HttpInterceptorContext())`. This requires a context because `hostClient` returns an `HttpHandler`, however `clientAddress` is not available in `StyxHttpClient` (nor it is relevant).

mikkokar: I see.

components/common/src/main/java/com/hotels/styx/server/HttpInterceptorContext.java
kvosper: I have added back the method `HttpInterceptorContext.create()` and reverted tests that use the new constructor.

components/common/src/main/java/com/hotels/styx/server/HttpInterceptorContext.java
```diff
@@ -17,18 +17,59 @@
...
+
+    /**
+     * Construct a new instance.
+     */
+    public HttpInterceptorContext() {
+        this(false, null);
+    }
+
+    // TODO deprecate
+    public static HttpInterceptor.Context create() {
```
mikkokar: BTW why are you so keen to deprecate this?

kvosper: The static-factory-method isn't really adding anything. It seems to just be a totally redundant wrapper for the constructor.

Sometimes when we add static-factory-methods to our classes it's to make it read more fluently when statically-imported, but this is just called `create` so it doesn't have that benefit.

It just changes `new HttpInterceptorContext()` to `HttpInterceptorContext.create()` which is actually longer and less convenient.

components/proxy/src/test/scala/com/hotels/styx/routing/handlers/ConditionRouterConfigSpec.scala
```diff
@@ -159,7 +159,8 @@ class ConditionRouterConfigSpec extends FunSpec with ShouldMatchers with Mockito
         |""".stripMargin)
 
     val router = new ConditionRouter.ConfigFactory().build(List(), routeHandlerFactory, config)
-    val resp = StyxFutures.await(router.handle(request, HttpInterceptorContext.create(false)).asCompletableFuture())
+
+    val resp = StyxFutures.await(router.handle(request, new HttpInterceptorContext()).asCompletableFuture())
 
     resp.status() should be(BAD_GATEWAY)
   }
```
mikkokar: Are the changes in this file still necessary?

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpPipelineHandler.java
```diff
@@ -706,27 +711,11 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpRequest request) thro
...
     }
 
-    private static final class HttpInterceptorContext implements HttpInterceptor.Context {
-        private final Map<String, Object> context = new ConcurrentHashMap<>();
-        private final boolean secure;
-
-        public HttpInterceptorContext(boolean secure) {
-            this.secure = secure;
+    private static InetSocketAddress remoteAddress(ChannelHandlerContext ctx) {
+        if (ctx.channel() instanceof EmbeddedChannel) {
```
mikkokar: Might be worth adding a comment that this is just for unit testing?

components/server/src/test/java/com/hotels/styx/server/routing/AntlrMatcherTest.java
```diff
@@ -24,11 +24,11 @@
...
 
-    private final HttpInterceptorContext contextHttps = HttpInterceptorContext.create(true);
-    private final HttpInterceptorContext contextHttp = HttpInterceptorContext.create();
+    private final HttpInterceptorContext contextHttps = new HttpInterceptorContext(true);
+    private final HttpInterceptorContext contextHttp = new HttpInterceptorContext();
 
     @Test
-    public void matchesHttpProtocol() throws Exception {
+    public void matchesHttpProtocol() {
         AntlrMatcher matcher = AntlrMatcher.antlrMatcher("protocol() == 'https'");
```
mikkokar: These changes can also be backed off?

support/api-testsupport/pom.xml
```diff
@@ -44,10 +49,6 @@
       <artifactId>wiremock</artifactId>
       <scope>compile</scope>
     </dependency>
-    <dependency>
-      <groupId>com.hotels.styx</groupId>
-      <artifactId>styx-common</artifactId>
-    </dependency>
-  </dependencies>
 
+  </dependencies>
```
mikkokar: Is this change relevant? Looks like it could be backed off?

Let's try to minimise unnecessary changes. Later on if or when we need to consult the Git change log for the relevant changes, perhaps to find out what merge caused a bug, perhaps for another reason, irrelevant or stylistic changes are just a distraction. Therefore they should go into separate pull requests if possible.

components/proxy/src/main/java/com/hotels/styx/proxy/interceptors/RequestEnrichingInterceptor.java
```diff
@@ -37,28 +44,42 @@ public RequestEnrichingInterceptor(StyxHeaderConfig styxHeaderConfig) {
 
     @Override
     public StyxObservable<HttpResponse> intercept(HttpRequest request, Chain chain) {
-        return chain.proceed(enrich(request, chain.context().isSecure()));
+        return chain.proceed(enrich(request, chain));
```
mikkokar: I would pass in the `Context` only, as `enrich` doesn't need the full chain.

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpPipelineHandler.java
kvosper: I only moved this method, it's not new.

components/proxy/src/test/scala/com/hotels/styx/routing/handlers/ConditionRouterConfigSpec.scala
kvosper: I put back the old `HttpInterceptorContext.create()` but not `HttpInterceptorContext.create(boolean)` as I assume we do not want to set a precedent of following that convention (see my reply to your other comment of why I think the `create` methods are bad).

support/api-testsupport/pom.xml
kvosper: I moved it because I got briefly confused about the module relationships due to the `styx-common` dependency not being situated near `styx-api`, and wanting to prevent such confusion in the future. I don't think it merits an entirely separate PR though.

PR #281, opened 2018-09-13
Update ContentCutoffPlugin and enable test in DoubleSubscriptionPluginSpec

### LINKED ISSUE #203: [1.0] DoubleSubscribingPluginSpec fails in 1.0
## The problem
`DoubleSubscribingPluginSpec` has been disabled as it does not work in 1.0.

## Acceptance criteria
* Investigate and fix

kvosper:
Fixes #203

### DIFF
--- a/system-tests/e2e-suite/src/test/java/com/hotels/styx/plugins/ContentCutoffPlugin.java
+++ b/system-tests/e2e-suite/src/test/java/com/hotels/styx/plugins/ContentCutoffPlugin.java
@@ -35,7 +35,8 @@ public class ContentCutoffPlugin implements Plugin {
                         .flatMap(it -> {
                             LOGGER.info("do something with full content: " + it.bodyAs(UTF_8));
 
-                            return StyxObservable.of(response); // Error: Returns an already consumed response.
+                            // Error: Returns an already consumed response.
+                            return StyxObservable.of(response);
                         })
                 );
     }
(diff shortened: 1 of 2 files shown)


system-tests/e2e-suite/src/test/scala/com/hotels/styx/plugins/DoubleSubscribingPluginSpec.scala
```diff
@@ -65,9 +63,7 @@ class DoubleSubscribingPluginSpec extends FunSpec
 
 
   describe("Styx as a plugin container") {
-
-    // TODO: Mikko: Styx 2.0 API: Test fails. Look into it.
-    ignore("Tolerates plugins that break the content observable chain") {
+    it("Tolerates plugins that break the content observable chain") {
       val testClient = aggregatingTestClient("localhost", styxServer.httpPort)
 
       withTestClient(testClient) {
```
mikkokar: I think this test should assert that a PluginError has been logged. It could also assert that relevant metrics have been incremented.

Otherwise this test suggests there is nothing unusual about breaking a content observable chain.

system-tests/e2e-suite/src/test/java/com/hotels/styx/plugins/ContentCutoffPlugin.java
```diff
@@ -15,27 +15,29 @@
...
-                                }));
-
+    public StyxObservable<HttpResponse> intercept(HttpRequest httpRequest, Chain chain) {
+        return chain
+                .proceed(httpRequest)
+                .flatMap(response -> response.toFullResponse(1000000)
+                        .flatMap(it -> {
+                            LOGGER.info("do something with full content: " + it.bodyAs(UTF_8));
+
+                            return StyxObservable.of(response); // Error: Returns an already consumed response.
```
mikkokar: I would move the comment to the line above the return statement.

PR #282, opened 2018-09-14
Streamline Styx HTTP Client interfaces

mikkokar:
Streamline Styx HTTP Client interfaces so that they can be published together with 1.0 API.

To sum up the changes:
* The main HTTP client interfaces are moved into the `com.hotels.styx.client` package.
* The `StyxHttpClient` has been renamed to `StyxBackendServiceClient`. This is more descriptive, and it frees the more succinct `StyxHttpClient` to the more commonly used `FullHttpClient`.

* The `FullHttpClient` interface is renamed to `HttpClient` which in the next iteration will be receiving a fluent interface to deal with both full and streaming HTTP requests, and to adjust contextual attributes per request basis.

What do you think?

Does this renaming go too far or are there simpler approaches?


components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.java
```diff
@@ -0,0 +1,398 @@
...
+import static java.util.Objects.nonNull;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
+import static java.util.stream.StreamSupport.stream;
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * A configurable HTTP client that uses connection pooling, load balancing, etc.
+ */
+public final class StyxBackendServiceClient implements BackendServiceClient {
```
kvosper: It would be nice if this class could be named `BackendServiceClient`, though that would obvious conflict with the interface name.

PR #284, opened 2018-09-17
Remove announcer (fixes #268)

### LINKED ISSUE #268: Remove Announcer class
## The problem
The class `com.hotels.styx.api.Announcer` seems redundant.

## Detailed description
`Announcer` holds a list of listeners and uses relatively complex logic (using `java.lang.reflect.Proxy`) to iterate through and call their listen methods.

 Specifically, it creates a proxy object implementing the same listener interface, and then when methods are called on that object, it propagates the call to all of the listeners in its list.

It appears that this could be just as easily accomplished by using the list of listeners directly and calling `forEach(...)` to execute the listen methods. This suggests that `Announcer` is totally unnecessary.

## Acceptance criteria
* Use simple alternative as suggested above
* Remove `Announcer` class

kvosper:
Fixes #268


components/client/src/main/java/com/hotels/styx/client/OriginsInventory.java
```diff
@@ -88,7 +88,7 @@
     private static final HealthyEvent HEALTHY = new HealthyEvent();
     private static final UnhealthyEvent UNHEALTHY = new UnhealthyEvent();
 
-    private final Announcer<OriginsChangeListener> inventoryListeners = Announcer.to(OriginsChangeListener.class);
+    private final List<OriginsChangeListener> inventoryListeners = new CopyOnWriteArrayList<>();
```
mikkokar: Please just move the `Announcer` class away from `styx-api` into `styx-common`. Then these other classes can still keep using it.

kvosper: Also, please note that the `Announcer` is used inside the API class `AbstractRegistry` so it cannot be moved without also moving that one.

kvosper:
Postponed for now.

PR #288, opened 2018-09-18
Improve styx client

mikkokar:
TBD ...


components/client/src/main/java/com/hotels/styx/client/HttpClient.java
```diff
@@ -17,22 +17,108 @@
...
+     * A HTTP request transaction.
      *
-     * The sendRequest will be sent when the returned observable is subscribed to.
-     * In order to cancel the ongoing request, just unsubscribe from it.
+     * This interface allows client attributes and context to be customised
+     * for each request without having to rely on configured default values
+     * in the client.
+     *
+     */
+    interface Transaction {
```
kvosper: This interface is inside `HttpClient`, but `HttpClient` does not use it.

components/client/src/main/java/com/hotels/styx/client/HttpClient.java
```diff
@@ -17,22 +17,108 @@
...
+        /**
+         * Sends a HTTP request message using this client.
+         *
+         * @deprecated use {@link this::send} instead.
+         *
+         * @param request a full HTTP request object
+         * @return a future of full HTTP request object
+         */
+        @Deprecated
+        default CompletableFuture<FullHttpResponse> sendRequest(FullHttpRequest request) {
```
kvosper: Why create a method that is already deprecated?

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -96,73 +156,165 @@ private static Origin originFromRequest(FullHttpRequest request, Boolean isHttps
...
 
-        public Builder responseTimeoutMillis(int responseTimeout) {
-            this.responseTimeout = responseTimeout;
+        /**
+         * Sets the TCP connection timeout.
+         *
+         * @param timeoutMs desired TCP connection timeout
+         * @return this {@link Builder}
+         */
+        public Builder connectTimeout(int timeoutMs) {
```
kvosper: We usually have a signature like `(int timeout, TimeUnit unit)` for these kinds of method.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -44,39 +47,96 @@
...
-    private final Connection.Factory connectionFactory;
-    private final boolean isHttps;
-
-    private StyxHttpClient(Builder builder, boolean isHttps) {
-        this.userAgent = Optional.ofNullable(builder.userAgent);
-        this.connectionSettings = builder.connectionSettings;
-        this.maxResponseSize = builder.maxResponseSize;
-        this.connectionFactory = builder.connectionFactory;
-        this.isHttps = isHttps;
+    private final Builder transactionParameters;
```
kvosper: Storing the builder violates immutability.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClientTransaction.java
```diff
@@ -0,0 +1,74 @@
...
+package com.hotels.styx.client;
+
+import com.hotels.styx.api.FullHttpRequest;
+import com.hotels.styx.api.FullHttpResponse;
+import com.hotels.styx.client.netty.connectionpool.NettyConnectionFactory;
+
+import java.util.concurrent.CompletableFuture;
+
+class StyxHttpClientTransaction implements HttpClient.Transaction {
+    private StyxHttpClient.Builder transactionParameters;
```
kvosper: Storing builders violates immutability.
Also these fields are not `final`.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
```diff
@@ -54,13 +59,28 @@
      * @param requestOperationFactory used to create operation objects that send http requests via this connection
      */
     @VisibleForTesting
-    public NettyConnection(Origin origin, Channel channel, HttpRequestOperationFactory requestOperationFactory) {
+    public <T> NettyConnection(Origin origin, Channel channel, HttpRequestOperationFactory requestOperationFactory, HttpConfig httpConfig, SslContext sslContext) {
```
kvosper: Why is there a `<T>` here?

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
```diff
@@ -54,13 +59,28 @@
...
         this.channel = requireNonNull(channel);
         this.requestOperationFactory = requestOperationFactory;
         this.channel.pipeline().addLast(new TimeToFirstByteHandler(this));
         this.channel.closeFuture().addListener(future ->
                 listeners.announce().connectionClosed(NettyConnection.this));
+
+        addChannelHandlers(channel, httpConfig, sslContext);
+    }
+
+    private static void addChannelHandlers(Channel channel, HttpConfig httpConfig, SslContext sslContext) {
```
kvosper: If this is purely for testing, perhaps we should let the test specify which `ChannelHandler`s to add, so that we avoid adding too much code that is only used during tests?

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnectionFactory.java
```diff
@@ -100,19 +108,22 @@ private synchronized void bootstrap(ConnectionSettings connectionSettings) {
         }
     }
 
+    public CompletableFuture<Void> close() {
```
kvosper: Perhaps we should have a utility method somewhere for turning netty futures into `CompletableFuture`s somewhere, to avoid mixing it with this Styx-specific code.

components/client/src/test/unit/java/com/hotels/styx/client/StyxHttpClientTest.java
```diff
@@ -54,208 +54,281 @@
...
+        assertThat(threadExists("test-client"), is(false));
     }
 
-    @Test(expectedExceptions = OriginUnreachableException.class)
-    public void throwsOriginUnreachableExceptionWhenDnsResolutionFails() throws Throwable {
-        try {
-            httpClient().sendRequest(get("/foo.txt").header(HOST, "a.b.c").build()).get();
-        } catch (ExecutionException cause) {
-            throw cause.getCause();
+    private static Boolean threadExists(String threadName) {
```
kvosper: We could use java streams for this:
```
 private static Boolean threadExists(String threadName) {
        return Thread.getAllStackTraces().keySet().stream()
                .anyMatch(thread ->
                        thread.getName().startsWith(threadName));
    }
```

components/client/src/test/unit/java/com/hotels/styx/client/StyxHttpClientTest.java
```diff
@@ -54,208 +54,281 @@
...
 
-    private StyxHttpClient httpsClient() {
-        return new StyxHttpClient.Builder()
-                .connectionSettings(new ConnectionSettings(1000))
-                .tlsSettings(new TlsSettings.Builder()
-                        .authenticate(false)
-                        .build())
-                .responseTimeoutMillis(6000)
-                .build();
+    // TODO: Mikko:
```
kvosper: `TODO` left in PR.

components/client/src/test/unit/java/com/hotels/styx/client/StyxHttpClientTest.java
```diff
@@ -54,208 +54,281 @@
...
+        } catch (ExecutionException e) {
+            throw e.getCause();
         }
     }
 
-    private Connection mockConnection(HttpResponse response) {
-        Connection connection = mock(Connection.class);
-        when(connection.write(any(HttpRequest.class))).thenReturn(Observable.just(response));
-        return connection;
+    @Test(enabled = false)
```
kvosper: Disabled test.

components/client/src/main/java/com/hotels/styx/client/HttpClient.java
mikkokar: Due to backwards compatibility.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
mikkokar: Yeah, this is a good suggestion.

components/client/src/main/java/com/hotels/styx/client/HttpClient.java
kvosper: But as this is a new interface, does backwards-compatibility apply?

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
mikkokar: It does. But it saves lots of repeated code.

The `transactionParameters` stores the default client settings at the HTTP client. These settings are inherited into HTTP request transaction where the parameters can be overridden. Storing them as a builder offers a convenient way of inheriting the default settings into a transaction. For two reasons:

* avoids copying parameters into auxiliary data structures
* any copying related to parameter inheritance can be encapsulated in the builder class itself

Perhaps I'll need to rename `transactionParameters` to `defaultClientConfiguration`.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClientTransaction.java
mikkokar: I will mark the fields as `final`. Let's keep the discussion about usage of builder as a part of your previous comment. Therefore resolving this finding.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
kvosper: We have been consistent in our policy toward immutability so far, we cannot justify violating that just to type less code.

Besides, preserving immutability is the entire purpose of using the builder pattern. If we wanted to write mutable code we wouldn't use builders at all.

As a way to reduce code without introducing mutability, we could `clone` the builder, though there would still be internal mutability.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
mikkokar: This change is necessary to allow API consumer to choose whether to send a request over `http` or secure `https` protocol.

The problem we had is that `NettyConnectionFactory` can only provide channels that are either bound to `http` or `https` protocol. Therefore a factory that creates `http` channels cannot be used to send `https` traffic. This is fine for the styx proxy but doesn't work well with a stand-alone client.

This changes so that `NettyConnectionFactory` consumer can choose if it wants a `http` or `https` channel. That is, the factory has just enough information to build "bare" netty channels, and the actual channel configuration is injected from the consumer in a `createConnection` call.

mikkokar: Updated PR description with a rationale to this change.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnectionFactory.java
mikkokar: I could actually add one in `styx-common`.

components/client/src/test/unit/java/com/hotels/styx/client/StyxHttpClientTest.java
mikkokar: Okay. It is there to remind you guys to offer your opinion.

I think this test case should be removed if it doesn't run 100% of the time. Now it seems to depend on your ISP if it runs or not.

Shall we just get rid of this test? @a-dlatorre @kvosper what do you think?

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
mikkokar: The builder pattern in fact in Java is used as a substitute for optional constructor arguments (See Effective Java by Joshua Bloch). The problem with Java is that the constructors don't scale well with large numbers of parameters. It has little to do with immutability (although some).

Having said that. The class at the moment indeed has a problem with immutability. But this is not a problem with storing the builder itself. The problem is that the builder is not cloned (as you perhaps noticed?) before it is passed in to the `StyxHttpClient` constructor.

I will add a call to the `clone` method.

components/client/src/main/java/com/hotels/styx/client/HttpClient.java
mikkokar: Right. You are correct. This can be removed from the `HttpClient` interface but must be left in the `StyxHttpClient` class.

Will remove this.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
dvlato: Why don't you use the Builder(builder) constructor that the Builder provides instead of clone()?  Cloneable is not something we want to encourage and I would say it's been strongly advised against for years... 
Other than that, Builder is used to avoid creating 'incomplete objects'  (that's why we cannot use 'Java Beans' for that) and also avoid the limitations of constructors and static factories (that they do not provide either named or optional parameters) so we can construct objects in a clearer/more friendly way (it's usually complex objects, but we are using it in most cases). Even though it's true that it's one of the advantages of Builders, immutability can be achieved by using constructors or static factories...

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
mikkokar: By accident. It has now been removed.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
mikkokar: Hey all. I mistakenly referred to `clone`. The `Builder` already implements a method called `copy` that replicates the builder object. The client is instantiated from the `build` object by:

```
            return new StyxHttpClient(connectionFactory, this.copy());
```

Apologies for the confusion.

mikkokar: The client class is now immutable from the outside. At least in terms of `Builder` object.

components/client/src/test/unit/java/com/hotels/styx/client/StyxHttpClientTest.java
mikkokar: Removed this test.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -96,73 +157,166 @@ private static Origin originFromRequest(FullHttpRequest request, Boolean isHttps
...
+         * Note that an actual response can take considerably longer time to arrive than @{code responseTimeoutMillis}.
+         * This can happen if origin sends the response slowly. Origin may send headers first, and then
+         * slowly drip feed the response body. This is acceptable as long as the TCP connection does not
+         * experience longer inactivity than @{code responseTimeoutMillis} between any two consecutive
+         * data packets.
+         *
+         * @param responseTimeoutMs maximum tolerated inactivity on the TCP connection.
+         * @return this {@link Builder}
+         */
+        public Builder responseTimeout(int responseTimeoutMs) {
```
kvosper: Please apply the `int duration, TimeUnit timeUnit` signature here too.

PR #290, opened 2018-09-24
Fix origins reload thread leak

kvosper:
We have a bug in which thread pools are created each time origins are reloaded and never shut down.
This PR fixes the bug and adds a couple of small utility classes in the `e2e-suite` that may help with debugging & testing in the future.


components/client/src/main/java/com/hotels/styx/client/healthcheck/OriginHealthStatusMonitor.java
```diff
@@ -37,10 +38,11 @@
          *
          * @param id                the backend service id
          * @param healthCheckConfig configuration of the health check
-         * @param supplier          of the actual function which decided if the origin is healthy or not
+         * @param healthCheckFunction          of the actual function which decided if the origin is healthy or not
+         * @param client TODO
          * @return the monitor of the origin
          */
-        OriginHealthStatusMonitor create(Id id, HealthCheckConfig healthCheckConfig, Supplier<OriginHealthCheckFunction> supplier);
+        OriginHealthStatusMonitor create(Id id, HealthCheckConfig healthCheckConfig, Supplier<OriginHealthCheckFunction> healthCheckFunction, HttpClient client);
```
dvlato: Since we have moved the HttpClient creation outside of healthCheckFunction(), do we still need to use a supplier for the latter? I'm guessing it was used to  lazily create the HttpClient.

components/proxy/src/main/java/com/hotels/styx/proxy/BackendServicesRouter.java
```diff
@@ -154,20 +145,49 @@ public void onChange(Registry.Changes<BackendService> changes) {
                     .hostClientFactory(hostClientFactory)
                     .build();
 
-            pipeline = new ProxyToClientPipeline(newClientHandler(backendService, inventory, originStatsFactory), inventory);
+            pipeline = new ProxyToClientPipeline(newClientHandler(backendService, inventory, originStatsFactory), () -> {
+                inventory.close();
```
dvlato: It seems here (and above in `return runAsync(() -> {`... ) we might be not following the principle that says we should keep  statements in a function at the same level of abstraction.  Following up on that, it also seems the pipeline could contain the logic to close itself instead of transferring the responsibility to the caller.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -68,7 +68,7 @@ private StyxHttpClient(NettyConnectionFactory connectionFactory, Builder paramet
     /**
      * Indicates that a request should be sent using secure {@code https} protocol.
      *
-     * @return a {@HttpClient.Transaction} instance that allows fluent method chaining
+     * @return a {@link HttpClient.Transaction} instance that allows fluent method chaining
```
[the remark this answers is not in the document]
kvosper: Looks like there are no changes related to the PR anyway, I'll remove them and we can resolve this another time.

components/client/src/main/java/com/hotels/styx/client/healthcheck/OriginHealthStatusMonitor.java
kvosper: I tried removing the `Supplier`, but this caused tests to fail. I put it back since the change was out of the scope of this task.

components/proxy/src/main/java/com/hotels/styx/proxy/BackendServicesRouter.java
kvosper: The idea was to avoid bloating the `ProxyToClientPipeline` constructor signature by providing a kind of listener that can react to `close` events, but I am open to alternatives.

components/client/src/main/java/com/hotels/styx/client/healthcheck/monitors/ScheduledOriginHealthStatusMonitor.java
```diff
@@ -105,7 +116,14 @@ private void resetListeners(Set<Origin> origins) {
     @Override
     protected CompletableFuture<Void> stopService() {
         this.hostHealthMonitorExecutor.shutdown();
-        return completedFuture(null);
+
+        return runAsync(() -> {
+            try {
+                hostHealthMonitorExecutor.awaitTermination(10, SECONDS);
+            } catch (InterruptedException e) {
+                throw propagate(e);
```
mikkokar: Should we set the thread interrupted flag before propagating?
Also IIRC `Throwables.propagate` is deprecated from recent Guava versions.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/OriginsReloadThreadsSpec.scala
```diff
@@ -0,0 +1,116 @@
...
+  override protected def afterAll(): Unit = {
+    backend1.stop()
+    backend2.stop()
+    backend3.stop()
+
+    super.afterAll()
+  }
+
+  describe("Reload Origins Endpoint") {
+    it("should not leak threads") {
```
mikkokar: 1. Be careful with wording of the test case name:

`describe("Styx health checking")` ...

`   it ("cleans up resources after origins reload")`

2. Also please raise a GitHub issue for this problem, and add the issue number in a comment to this test case. The GitHub issue will become handy if this ever needs to be double committed to 0.7 branch in future.

components/client/src/main/java/com/hotels/styx/client/healthcheck/monitors/ScheduledOriginHealthStatusMonitor.java
dvlato: It seems that the ForkJoinPool clears the thread interruption status (in awaitWork) so it's not a huge issue if we don't clear the status, but the path execution is different and we should never discard the interruption status as a principle.
However, maybe we should not use the JoinFork pool for blocking tasks?

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/OriginsReloadThreadsSpec.scala
kvosper: Issue created: #291

PR #298, opened 2018-10-03
A ByteStream abstraction to represent streaming content.

mikkokar:
Raising this just to get some preliminary comments for this work.

* The PR adds a ByteStream class that abstracts a live asynchronous content stream. The ByteStream implements a reactive-streams `Publisher` interface and is implemented on top of reactor-core `Flux`. 

* Netty `ByteBuf` has been encapsulated inside a new Styx `Buffer` class.

More details to come soon ...


components/api/src/main/java/com/hotels/styx/api/AggregateOperator.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public void onSubscribe(Subscription subscription) {
+        if (this.subscription == null) {
+            this.subscription = subscription;
+            this.subscription.request(Long.MAX_VALUE);
+        } else {
+            subscription.cancel();
+            throw new IllegalStateException("Second onSubscribe event to AggregateOperator");
```
kvosper: Looks like we should use the same exception + message as the other method (create a `throwSubscriptionException` method or something?)

components/api/src/main/java/com/hotels/styx/api/Buffer.java
```diff
@@ -0,0 +1,73 @@
...
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Styx byte buffer abstraction.
+ *
+ */
+public final class Buffer {
+    private final ByteBuf delegate;
+
+    public Buffer(ByteBuf byteBuf) {
```
kvosper: By having this constructor, netty `ByteBuf` is still visible in our API. I'm not sure what the solution is.

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -59,27 +50,27 @@
  * chunk must be processed as they arrive. It is also useful for dealing with
  * very large content sizes, and in situations where content size is not known
  * upfront.
- * <p>
+ * <p>z
```
kvosper: `z`

components/api/src/test/java/com/hotels/styx/api/ByteStreamTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+        StepVerifier.create(stream)
+                .expectNext(buf1)
+                .expectNext(buf2)
+                .expectNext(buf3)
+                .verifyComplete();
+    }
+
+    @Test
+    public void publisherBackpressure() {
```
kvosper: Not a proper test name.

components/api/src/test/java/com/hotels/styx/api/ByteStreamTest.java
```diff
@@ -0,0 +1,185 @@
...
+                .expectSubscription()
+                .verifyComplete();
+
+        assertThat(buf1.delegate().refCnt(), is(0));
+        assertThat(buf2.delegate().refCnt(), is(0));
+        assertThat(buf3.delegate().refCnt(), is(0));
+    }
+
+    @Test
+    public void aggregatesContent() throws ExecutionException, InterruptedException {
```
kvosper: We should also have a test for when the aggregation goes over the maximum bytes.

components/client/src/main/java/com/hotels/styx/client/Transport.java
```diff
@@ -107,7 +108,7 @@ private synchronized void returnIfConnected(Optional<ConnectionPool> connectionP
         return origin
                 .map(ConnectionPool::borrowConnection)
                 .orElseGet(() -> {
-                    request.releaseContentBuffers();
+                    toObservable(request.body()).forEach(it -> it.delegate().release());
```
kvosper: Should this use `request.body().drop()` now?

components/api/src/test/java/com/hotels/styx/api/ByteStreamTest.java
mikkokar: `supportsReactiveBackpressure`?

mikkokar: `supportsBackpressure`?

components/api/src/main/java/com/hotels/styx/api/AggregateOperator.java
```diff
@@ -0,0 +1,91 @@
...
+        this.upstream = requireNonNull(upstream);
+        this.maxSize = maxSize;
+    }
+
+    public CompletableFuture<Buffer> apply() {
+        if (active.compareAndSet(false, true)) {
+            this.upstream.subscribe(this);
+            return future;
+        } else {
+            throw new IllegalStateException("Secondary subscription!");
```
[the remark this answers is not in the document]
mikkokar: I will change to:

`throw new IllegalStateException("ByteStreamAggregator may only be started once.");`

And for what it is worth I have changed the class name to `ByteStreamAggregator`. As this is not really an operator. It is a consumer of a byte stream, not an operator that modifies the stream.

components/api/src/main/java/com/hotels/styx/api/AggregateOperator.java
mikkokar: Changed to `throw new IllegalStateException("ByteStreamAggregator supports only one Producer instance.");`

components/api/src/main/java/com/hotels/styx/api/Buffer.java
mikkokar: Yeah. I still need to think about this.

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpResponseWriter.java
```diff
@@ -71,7 +72,9 @@
                 }
             });
 
-            Subscription subscriber = toRxObservable(response.body()).subscribe(new Subscriber<ByteBuf>() {
+            // TODO: Mikko
```
kvosper: TODOs should be converted to issues.

components/api/src/main/java/com/hotels/styx/api/ByteStreamAggregator.java
```diff
@@ -0,0 +1,91 @@
...
+    private final AtomicBoolean active = new AtomicBoolean();
+    private final CompositeByteBuf aggregated = compositeBuffer();
+    private Subscription subscription;
+
+    public ByteStreamAggregator(Publisher<Buffer> upstream, int maxSize) {
+        this.upstream = requireNonNull(upstream);
+        this.maxSize = maxSize;
+    }
+
+    public CompletableFuture<Buffer> apply() {
```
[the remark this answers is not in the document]
mikkokar: That would expose a half-constructed `ByteStreamAggregator` object to the `upstream`.

Also because the `ByteStreamAggregator` is package private, the `apply` method is never visible to Styx-API users, and therefore we don't have to solve this problem right now.

components/api/src/main/java/com/hotels/styx/api/ByteStreamAggregator.java
```diff
@@ -0,0 +1,91 @@
...
+
+class ByteStreamAggregator implements Subscriber<Buffer> {
+    private final Publisher<Buffer> upstream;
+    private final int maxSize;
+    private final CompletableFuture<Buffer> future = new CompletableFuture<>();
+    private final AtomicBoolean active = new AtomicBoolean();
+    private final CompositeByteBuf aggregated = compositeBuffer();
+    private Subscription subscription;
+
+    public ByteStreamAggregator(Publisher<Buffer> upstream, int maxSize) {
```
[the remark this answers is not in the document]
mikkokar: 👍

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -329,16 +303,16 @@ public Builder(HttpResponse response) {
...
+         * @param response      a full response for which the builder is based on
+         * @param byteStream a content byte stream
          */
-        public Builder(FullHttpResponse response, StyxObservable<ByteBuf> contentStream) {
+        public Builder(FullHttpResponse response, ByteStream byteStream) {
             this.status = statusWithCode(response.status().code());
             this.version = httpVersion(response.version().toString());
             this.headers = response.headers().newBuilder();
-            this.body = contentStream;
+            this.body = byteStream;
```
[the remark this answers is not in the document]
mikkokar: No. Added the null checks in Builder. Do you think it still warrants one here?

components/api/src/test/java/com/hotels/styx/api/ByteStreamTest.java
```diff
@@ -0,0 +1,198 @@
...
+        Buffer aggregated = stream.aggregate(100).get();
+        assertThat(decodeUtf8String(aggregated), is("abc"));
+    }
+
+    @Test
+    public void contentAggregationOverflow() throws ExecutionException, InterruptedException {
+        ByteStream stream = new ByteStream(Flux.just(buf1, buf2, buf3));
+
+        Throwable cause = stream.aggregate(2)
+                .handle((result, throwable) -> throwable)
```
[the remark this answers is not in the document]
mikkokar: No.
The code won't compile unless you map the outcome of `stream.aggregate()` into a `CompletableFuture<Throwable>`.

components/api/src/test/java/com/hotels/styx/api/ByteStreamTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    private String decodeUtf8String(Buffer buffer) {
+        return new String(buffer.content(), UTF_8);
+    }
+
+    private Buffer toUpperCase(Buffer buffer) {
+        return new Buffer(decodeUtf8String(buffer).toUpperCase(), UTF_8);
+    }
+
+    @Test(enabled = false)
```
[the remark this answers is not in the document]
mikkokar: Removed.

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpResponseWriter.java
mikkokar: Removed these todos. They are unnecessary.

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -200,40 +192,22 @@ public boolean isRedirect() {
      * @return a {@link StyxObservable}
      */
     public StyxObservable<FullHttpResponse> toFullResponse(int maxContentBytes) {
```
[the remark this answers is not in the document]
mikkokar: No not yet. Its type signature will change once the next phase of the work goes out.

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpPipelineHandler.java
```diff
@@ -495,7 +497,7 @@ private HttpResponse exceptionToResponse(Throwable exception, HttpRequest reques
 
         return responseEnhancer.enhance(HttpResponse.response(status), request)
                 .header(CONTENT_LENGTH, message.getBytes(UTF_8).length)
-                .body(StyxObservable.of(message), UTF_8)
+                .body(new ByteStream(RxReactiveStreams.toPublisher(Observable.just(new Buffer(message, UTF_8)))))
```
dvlato: Super minor: static import?

mikkokar: ack

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpResponseWriter.java
```diff
@@ -157,7 +158,7 @@ private void onWriteEmptyLastChunkOutcome(ChannelFuture writeOp) {
             return future;
         } catch (Throwable cause) {
             LOGGER.warn("Failed to convert response headers. response={}, Cause={}", new Object[]{response, cause});
-            response.releaseContentBuffers();
+            toObservable(response.body()).forEach(it -> Buffers.toByteBuf(it).release());
```
dvlato: This could be a stupid comment: do we need to expose a method to release the content buffers to plugins?  In that case, it could also be used here.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/StaticResponseHandler.java
```diff
@@ -46,7 +51,7 @@ private StaticResponseHandler(int status, String text) {
 
     @Override
     public StyxObservable<HttpResponse> handle(HttpRequest request, HttpInterceptor.Context context) {
-        return StyxObservable.of(HttpResponse.response(statusWithCode(status)).body(StyxObservable.of(text), UTF_8).build());
+        return StyxObservable.of(response(statusWithCode(status)).body(new ByteStream(toPublisher(Observable.just(new Buffer(text, UTF_8))))).build());
```
dvlato: Can't we use Flux.just here?

support/origins-starter-app/src/main/java/com/hotels/styx/support/origins/AppHandler.java
```diff
@@ -57,11 +61,11 @@ public AppHandler(Origin origin) {
                     request.queryParam("status").ifPresent(status ->
                             responseBuilder
                                     .status(httpResponseStatus(status))
-                                    .body(StyxObservable.of("Returning requested status (" + status + ")"), UTF_8)
+                                    .body(new ByteStream(toPublisher(Observable.just(new Buffer("Returning requested status (" + status + ")", UTF_8)))))
```
dvlato: As in other parts of coude, Is there any reason (visibility?) why we cannot use Flux.just() instead of creating an rxJava Observable with .just()  and transforming it to a publisher  ?

system-tests/e2e-suite/src/test/scala/com/hotels/styx/plugins/AggregatingPluginSpec.scala
```diff
@@ -65,7 +65,7 @@ class AggregatingPluginSpec extends FunSpec
 
     it("Gets response from aggregating plugin (with body)") {
       mockServer.stub("/body", responseSupplier(
-        () => response(OK).body(StyxObservable.from(Seq(chunk("a"), chunk("b"), chunk("c"), chunk("d"), chunk("e")).asJava)).build()
+        () => response(OK).body(new ByteStream(RxReactiveStreams.toPublisher(Observable.from(Seq(chunk("a"), chunk("b"), chunk("c"), chunk("d"), chunk("e")).asJava)))).build()
```
dvlato: static import?

system-tests/e2e-suite/src/test/scala/com/hotels/styx/client/OriginClosesConnectionSpec.scala
```diff
@@ -127,7 +126,7 @@ class OriginClosesConnectionSpec extends FunSuite
         .toStreamingRequest)
 
     responseObservable
-      .doOnNext((t: HttpResponse) => toRxObservable(t.body()).subscribe(contentSubscriber))
+      .doOnNext((t: HttpResponse) => RxReactiveStreams.toObservable(t.body()).subscribe(contentSubscriber))
```
dvlato: static import?

system-tests/e2e-suite/src/test/scala/com/hotels/styx/plugins/AsyncResponseContentSpec.scala
```diff
@@ -88,14 +85,14 @@ class AsyncDelayPlugin extends PluginAdapter {
     chain.proceed(request)
       .flatMap(asJavaFunction((response: HttpResponse) => {
 
-        val transformedContent: Observable[ByteBuf] = toRxObservable(response.body())
+        val transformedContent: Observable[Buffer] = RxReactiveStreams.toObservable(response.body())
```
dvlato: static import?

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpResponseWriter.java
mikkokar: This is in fact an excellent comment. In my opinion, if possible, we should not expose reference counts outside of `Buffer` at all.

However we need to ensure this indeed is the case, and one place we need to revisit is `ByteStream.aggregate` method. It returns an aggregated `Buffer` but the end-users have no way of releasing it yet.

I would like to implement these improvements in the next round though.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/StaticResponseHandler.java
mikkokar: Looks like Flux can be used here.

support/origins-starter-app/src/main/java/com/hotels/styx/support/origins/AppHandler.java
mikkokar: Yeah it looks it is OK to use Flux.

dvlato: I didn't want to add the comment in every occurrence but you can search for Observable.just()...

components/api/src/main/java/com/hotels/styx/api/Buffer.java
```diff
@@ -0,0 +1,73 @@
...
+package com.hotels.styx.api;
+
+import io.netty.buffer.ByteBuf;
+
+import java.nio.charset.Charset;
+
+import static io.netty.buffer.Unpooled.copiedBuffer;
+import static java.util.Objects.requireNonNull;
+
+/**
```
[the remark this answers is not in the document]
kvosper: I think we should avoid saying "A Styx byte buffer" and just say "A byte buffer". Otherwise it seems like we're implying that we have some kind of unique (unspecified) byte buffer behaviour. To us, it seems important because we're differentiating it from Netty `ByteBuf`, but someone looking at the docs for the first time wouldn't be thinking about that (and can see that it is a Styx class from the package name).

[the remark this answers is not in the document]
mikkokar: 👍 Makes sense.

components/api/src/main/java/com/hotels/styx/api/ByteStream.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Transform the stream by performing a mapping operation on each {@link Buffer} object.
+     *
+     * The mapping operation automatically maintains the @{link Buffer} reference counts as
+     * follows:
+     *
+     * <ul>
+     * <li> When the mapping function returns a new {@link Buffer} instance, the reference count for
+     * the old one is automatically decremented.</li>
+     * <li> When the mapping function modifies the {@link Buffer} in place, returning the same instance
```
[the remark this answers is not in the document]
mikkokar: I would leave this open for now. Performance-wise it would be sensible to allow in-place modifications for the buffer. WDYT?

PR #302, opened 2018-10-05
Log written vs acknowledged bytes from HttpResponseWriter

mikkokar:
Improve HttpResponseWriter logging:
- Log the number of bytes written to the socket
- Log the number of bytes acknowledged by completion of Netty ChannelFuture for the write operation.

### DIFF
--- a/components/server/src/test/java/com/hotels/styx/server/netty/connectors/HttpResponseWriterTest.java
+++ b/components/server/src/test/java/com/hotels/styx/server/netty/connectors/HttpResponseWriterTest.java
@@ -16,8 +16,11 @@
 package com.hotels.styx.server.netty.connectors;
 
 
+import ch.qos.logback.classic.Level;
 import com.google.common.collect.ImmutableList;
 import com.hotels.styx.api.HttpResponse;
+import com.hotels.styx.api.exceptions.TransportLostException;
+import com.hotels.styx.support.matchers.LoggingTestSupport;
 import io.netty.buffer.ByteBuf;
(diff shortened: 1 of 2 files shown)


components/server/src/test/java/com/hotels/styx/server/netty/connectors/HttpResponseWriterTest.java
```diff
@@ -375,6 +388,40 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpResponse response) th
...
+                        CompletableFuture<Void> future = writer.write(response);
+                        assertThat(future.isDone(), is(false));
+
+                        contentObservable.onNext(copiedBuffer("aaa", UTF_8));
+                        assertThat(future.isDone(), is(false));
+
+                        contentObservable.onNext(copiedBuffer("bbbb", UTF_8));
+                        assertThat(future.isDone(), is(false));
+
+                        contentObservable.onError(new TransportLostException(new InetSocketAddress("localhost", 5050), newOriginBuilder("localhost", 5050).build()));
```
dvlato: As a general rule, we might want to use InetAddress.getLoopbackAddress()  instead of "localhost"  so  we never get an error when trying to resolve "localhost".  In this case the resolution of "localhost"  wouldn't throw an error if it can be resolved but the loopbackadress would still prevent any resolution from happening.

mikkokar: 👍  Good point.

PR #309, opened 2018-10-11
Styx Eventual class for asynchronous events.

mikkokar:
Replaces `StyxObservable` with `Eventual` class, that asynchronously publishes a single event only.

A main difference between a Future and an Eventual supports deferred actions like Reactor Mono (in fact it is just a wrapper around Mono). That is, where as CompletableFuture action starts running straight away, an Eventual needs an explicit subscription to start the action.

The codebase still needs some tidying up, but you should get an impression how the final API will look like.


components/client/src/test/unit/java/com/hotels/styx/client/TransportTest.java
```diff
@@ -67,111 +70,150 @@ public void setUp() {
...
         subscriber.unsubscribe();
-        verify(pool).closeConnection(any(Connection.class));
+
+        verify(pool, never()).closeConnection(any(Connection.class));
+        verify(pool, never()).returnConnection(any(Connection.class));
     }
 
     @Test
-    public void releasesIfRequestIsUnsubscribedAfterHeaders() {
+    public void returnsConnectionBackToPool_delayedResponseError() {
```
kvosper: Another test that does not follow naming conventions.

components/client/src/test/unit/java/com/hotels/styx/client/TransportTest.java
```diff
@@ -67,111 +70,150 @@ public void setUp() {
...
-        verify(pool, never()).closeConnection(any(Connection.class));
+        assertThat(subscriber.getOnNextEvents().isEmpty(), is(true));
+        assertThat(subscriber.getOnErrorEvents().size(), is(1));
 
-        subscription.unsubscribe();
         verify(pool).closeConnection(any(Connection.class));
     }
 
+    @Test
+    public void terminatesConnection_bodyIsUnsubscribed() {
```
kvosper: Again.

components/common/src/main/java/com/hotels/styx/api/ResponseEventListener.java
```diff
@@ -0,0 +1,154 @@
...
+ */
+public class ResponseEventListener {
+    private final Observable<HttpResponse> publisher;
+    private Runnable cancelAction;
+    private Consumer<Throwable> responseErrorAction;
+    private Consumer<Throwable> contentErrorAction;
+    private Runnable onCompletedAction;
+
+    private final StateMachine<State> fsm = new StateMachine.Builder<State>()
+            .initialState(State.INITIAL)
```
kvosper: These `State` instances can be statically imported to make the code more readable.

components/common/src/main/java/com/hotels/styx/api/ResponseEventListener.java
```diff
@@ -0,0 +1,154 @@
...
+    ResponseEventListener(Observable<HttpResponse> publisher) {
+        this.publisher = publisher;
+    }
+
+    public static ResponseEventListener from(rx.Observable<HttpResponse> publisher) {
+        return new ResponseEventListener(publisher);
+    }
+
+    public ResponseEventListener whenCancelled(Runnable action) {
+        this.cancelAction = action;
```
kvosper: These "when" methods should do null checks.

components/common/src/main/java/com/hotels/styx/api/ResponseEventListener.java
```diff
@@ -0,0 +1,154 @@
...
+                return State.TERMINATED;
+            })
+            .transition(State.STREAMING, ContentCancelled.class, event -> {
+                cancelAction.run();
+                return State.TERMINATED;
+            })
+            .onInappropriateEvent((state, event) -> state)
+            .build();
+
+    ResponseEventListener(Observable<HttpResponse> publisher) {
```
kvosper: I see lots of things that could be `private`, and fields that could be `final` in this class, such as the constructor and in the inner classes further down. Please resolve them all.

components/common/src/test/java/com/hotels/styx/api/ResponseEventListenerTest.java
```diff
@@ -0,0 +1,171 @@
...
+                        .build()))
+                .whenCancelled(() -> cancelled.set(true))
+                .apply()
+                .toBlocking()
+                .first();
+
+        assertThat(cancelled.get(), Matchers.is(false));
+
+        StepVerifier.create(evResponse2.body())
+                .then(() -> assertThat(cancelled.get(), Matchers.is(false)))
```
kvosper: Please statically import `Matchers.is`.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/StandardHttpPipeline.java
```diff
@@ -77,19 +76,19 @@ public StandardHttpPipeline(List<HttpInterceptor> interceptors, HttpHandler hand
...
                 try {
                     return interceptor.intercept(request, chain);
                 } catch (Throwable e) {
-                    return StyxObservable.error(e);
+                    return Eventual.error(e);
                 }
             }
-            return fromRxObservable(toRxObservable(client.handle(request, this.context))
-                    .compose(StandardHttpPipeline::sendErrorOnDoubleSubscription));
+            return new Eventual<>(RxReactiveStreams.toPublisher(RxReactiveStreams.toObservable(client.handle(request, this.context))
```
kvosper: Looks like these two `RxReactiveStreams` methods can be statically-imported. It would also be nice to lay out the code in a way that makes it more clear why we turn a publisher to an observable and then back again.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/OriginCancellationMetrics.scala
```diff
@@ -67,7 +67,15 @@ class OriginCancellationMetrics extends FunSpec
       styxServer.metricsSnapshot.count("origins.app-1.requests.cancelled").get should be(0)
     }
 
-    it("Is incremented on an origin error") {
+
+    ignore("Is incremented on an origin error") {
```
kvosper: This ignored test will be to be changed into a functioning test before the PR can be approved.

system-tests/example-styx-plugin/src/main/java/loadtest/plugins/AsyncRequestPluginFactory.java
```diff
@@ -41,7 +41,7 @@ public Plugin create(PluginFactory.Environment environment) {
         }
 
         @Override
-        public StyxObservable<HttpResponse> intercept(HttpRequest request, Chain chain) {
+        public Eventual<HttpResponse> intercept(HttpRequest request, Chain chain) {
 
 
 //            resp1 = chain.context().styxObservable(1)
```
kvosper: Commented code should be removed.

components/api/src/main/java/com/hotels/styx/api/StreamingHttpMessage.java
```diff
@@ -93,4 +95,16 @@ default boolean chunked() {
         return HttpMessageSupport.chunked(headers());
     }
 
+    default void consume() {
```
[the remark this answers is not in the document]
mikkokar: The logic is the following:

* The `drop` removes the chunks from the initial ByteStream.
* The `aggregate(1)` just triggers the stream to be subscribed to.
* The empty buffer from aggregator is finally down referenced and released.

Also, the `CompletableFuture` starts executing eagerly, even before the `get` is called. So `thenApply` is going to be just fine. You can verify this yourself by putting together a quick test, etc.

components/api/src/main/java/com/hotels/styx/api/StreamingHttpMessage.java
```diff
@@ -93,4 +95,16 @@ default boolean chunked() {
...
+        body().drop().aggregate(1)
+                .thenApply(buffer -> {
+                    buffer.delegate().release();
+                    return null;
+                });
+    }
+
+    default void drop() {
```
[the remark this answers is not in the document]
mikkokar: No because the `drop` on the message still consumes the byte stream arriving from the TCP connection. It just drops each byte chunk that is received while retaining the publisher chain between styx server and styx client.

This message on the other hand drops the message. It is saying that we want to drop this message as it is and we are not interested in consuming the content. 

As a result the underlying TCP connection gets closed.

components/api/src/main/java/com/hotels/styx/api/StreamingHttpMessage.java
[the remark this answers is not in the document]
kvosper: Perhaps there should be a unit test?

components/api/src/main/java/com/hotels/styx/api/StreamingHttpMessage.java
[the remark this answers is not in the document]
kvosper: So this method must be used before any of the content is read?

components/common/src/test/java/com/hotels/styx/api/ResponseEventListenerTest.java
```diff
@@ -0,0 +1,171 @@
...
+        contentError = new AtomicReference<>();
+        completed = new AtomicBoolean();
+    }
+
+    @Test
+    public void doesntFireUnnecessaryEvents() {
+        Observable<HttpResponse> publisher = Observable.just(response(OK).body(new ByteStream(Flux.just(new Buffer("hey", UTF_8)))).build());
+        TestSubscriber<HttpResponse> subscriber = TestSubscriber.create();
+
+        Observable<HttpResponse> evResponse2 = ResponseEventListener.from(publisher)
```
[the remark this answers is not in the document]
mikkokar: 👍

components/common/src/test/java/com/hotels/styx/api/RequestsTest.java
```diff
@@ -0,0 +1,202 @@
...
+    @BeforeMethod
+    public void setUp() {
+        publisher = TestPublisher.create();
+        request = HttpRequest.get("/").body(new ByteStream(publisher)).build();
+        response = HttpResponse.response(OK).body(new ByteStream(publisher)).build();
+        completed = new AtomicReference<>();
+    }
+
+    @Test
+    public void requestDoFinallyActivatesWhenSuccessfullyCompleted() {
```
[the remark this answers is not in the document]
kvosper: OK. I was distracted by my IDEs "code duplication" warnings.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/HttpRequestOperation.java
```diff
@@ -154,16 +156,17 @@ private static boolean requestIsOngoing(RequestBodyChunkSubscriber bodyChunkSubs
...
-                removeProxyBridgeHandlers(nettyConnection);
-                if (requestIsOngoing(requestRequestBodyChunkSubscriber.get())) {
-                    LOGGER.warn("Origin responded too quickly to an ongoing request, or it was cancelled. Connection={}, Request={}.",
-                            new Object[]{nettyConnection.channel(), this.request});
-                    nettyConnection.close();
-                }
-            }
-        }).map(response -> response.newBuilder().build());
+        return observable.map(response ->
+                        Requests.doFinally(response, cause -> {
```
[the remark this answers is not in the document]
mikkokar: In all honesty the `ResponseEventListener` class got implemented after this change. However, it is not necessary to roll out a full response event listener in this instance because all we do here is to remove Netty handlers from the channel. We don't care if it is because cancellation or because of end of the stream. We just need an "end-of-stream" signal that triggers the channel clean up.

Additionally, the `ResponseEventListener` coordinates between reactive signals from a Styx `Eventual` and a `ByteStream`. For example it collapses response `Eventual` cancellation with `ByteStream` cancellation. It provides a single `whenCancelled` interface for the event listener regardless if the response was cancelled via `Eventual` or `ByteStream`.  It also deals with some interdependencies and signal ordering issues involved in response cancellation.

So because response event listener adds complexity by executing code that is not relevant for this use case, it is better to use a simpler and more straightforward `doFinally` hook to trigger a cleanup.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/OriginCancellationMetrics.scala
mikkokar: I will move the assertion into ChunkedDownloadSpec.

components/api/src/main/java/com/hotels/styx/api/StreamingHttpMessage.java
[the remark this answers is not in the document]
mikkokar: Let's remove this from the API for now, and make a note of this issue.

Basically the only way to "read" content before calling this method is to subscribe with a 3rd partly subscriber like so:

```
        flux = Flux.from(response.body())
                .subscribe(buf -> doSomething(buf))
```

So now it would be illegal to call `drop` while this subscription is ongoing. This is because it causes a 2nd subscription to the live data stream which is not supported.

components/common/src/main/java/com/hotels/styx/api/ResponseEventListener.java
```diff
@@ -0,0 +1,160 @@
...
+                return TERMINATED;
+            })
+            .onInappropriateEvent((state, event) -> state)
+            .build();
+
+    private ResponseEventListener(Observable<HttpResponse> publisher) {
+        this.publisher = requireNonNull(publisher);
+    }
+
+    public static ResponseEventListener from(rx.Observable<HttpResponse> publisher) {
```
dvlato: rx.Observable -> Observable?

components/client/src/test/unit/java/com/hotels/styx/client/TransportTest.java
mikkokar: We should discuss and clarify our naming conventions. Could you clarify what aspect doesn't conform to naming conventions? 

In the meanwhile I will rename this test to: `returnsConnectionBackToPoolDueToCancelledHeaders`.

components/client/src/test/unit/java/com/hotels/styx/client/TransportTest.java
mikkokar: Will rename to `returnsConnectionBackToPoolDueToDelayedResponseError`.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/StandardHttpPipeline.java
```diff
@@ -77,19 +77,19 @@ public StandardHttpPipeline(List<HttpInterceptor> interceptors, HttpHandler hand
...
                 try {
                     return interceptor.intercept(request, chain);
                 } catch (Throwable e) {
-                    return StyxObservable.error(e);
+                    return Eventual.error(e);
                 }
             }
-            return fromRxObservable(toRxObservable(client.handle(request, this.context))
-                    .compose(StandardHttpPipeline::sendErrorOnDoubleSubscription));
+            return new Eventual<>(toPublisher(RxReactiveStreams.toObservable(client.handle(request, this.context))
```
dvlato: Am I reading this correctly? You are transforming from publisher to observable and then back to publisher?

components/proxy/src/test/scala/com/hotels/styx/routing/interceptors/RewriteInterceptorSpec.scala
```diff
@@ -16,7 +16,7 @@
 package com.hotels.styx.routing.interceptors
 
 import com.hotels.styx.api.HttpResponse.response
-import com.hotels.styx.api.{HttpInterceptor, HttpRequest, HttpResponse, StyxObservable}
+import com.hotels.styx.api._
```
dvlato: Was this changed intended?

components/api/src/main/java/com/hotels/styx/api/StreamingHttpMessage.java
```diff
@@ -93,4 +95,21 @@ default boolean chunked() {
...
+    /**
+     * Consume the message by discarding the message body.
+     * <p>
+     * This method reads the entire message body from the networks and black holes
+     * all the traffic. This has the benefit of keeping the underlying TCP connection
+     * open for connection pooling.
+     * <p>
+     */
+    default void consume() {
+        body().drop().aggregate(1)
```
dvlato: I don't find this method too clear either.  Also, will the aggregate() complete exceptionally and need to generate a stack trace?  I don't have a suggestion on how to improve it though...

components/api/src/test/java/com/hotels/styx/api/EventualTest.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.testng.Assert.assertEquals;
+
+public class EventualTest {
+
+    @Test
+    public void newEventual() throws ExecutionException, InterruptedException {
```
kvosper: The name could be `createsFromFlux`.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/StandardHttpPipeline.java
mikkokar: It first turns the `Eventual` from `client.handle` into an `rx.Observable`, so that `sendErrorOnDoubleSubscription` can be composed in. After that the composed `rx.Observable` is converted back to publisher.

This needs to be tidied up but that could be a separate enhancement.

components/proxy/src/test/scala/com/hotels/styx/routing/interceptors/RewriteInterceptorSpec.scala
mikkokar: No

mikkokar: Will fix.

components/api/src/test/java/com/hotels/styx/api/EventualTest.java
mikkokar: Will rename to `createFromPublisher`.

components/common/src/main/java/com/hotels/styx/api/ResponseEventListener.java
mikkokar: Fixed.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/StandardHttpPipeline.java
kvosper: You could use lines + indentation to show which bits happen at which stage.

kvosper: It could be written line this, to show which parts happen when:
```
return new Eventual<>(
                    toPublisher(
                            toObservable(client.handle(request, this.context))
                                    .compose(StandardHttpPipeline::sendErrorOnDoubleSubscription)
                    )
            );
```

components/api/src/main/java/com/hotels/styx/api/StreamingHttpMessage.java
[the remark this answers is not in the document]
mikkokar: I have added an unit test in `StreamingHttpResponseTest` class.

components/api/src/main/java/com/hotels/styx/api/StreamingHttpMessage.java
mikkokar: Okay. I will take this comment onboard and we will do a separate round of improvements.

PR #310, opened 2018-10-12
Track current http requests in styx

salahqasem:
Added a new admin handler that will help in tracking only the current http requests to styx. for example if you noticed that there is a slowness in styx you can open that new admin handler "Current Requests" and it will show only the info of current HttpRequests in styx like uri, method, params ... etc and the running time of these requests, and the stack traces for the associated threads

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/CurrentRequestsHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/CurrentRequestsHandler.java
@@ -1,0 +1,174 @@
+/*
+  Copyright (C) 2013-2018 Expedia Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package com.hotels.styx.admin.handlers;
+
+import static com.google.common.net.MediaType.PLAIN_TEXT_UTF_8;
+import static com.hotels.styx.api.HttpHeaderNames.CONTENT_TYPE;
+import static com.hotels.styx.api.HttpResponseStatus.OK;
+import static java.lang.management.ManagementFactory.getThreadMXBean;
+import static java.lang.Thread.State.BLOCKED;
+import static java.lang.System.currentTimeMillis;
+import static java.lang.String.format;
+
+import java.lang.management.LockInfo;
+import java.lang.management.MonitorInfo;
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
+import java.nio.charset.Charset;
+
+import com.hotels.styx.api.HttpResponse;
+import com.hotels.styx.api.LiveHttpRequest;
+import com.hotels.styx.api.LiveHttpResponse;
+import com.hotels.styx.common.http.handler.BaseHttpHandler;
+import com.hotels.styx.server.track.CurrentRequestTracker;
+
+/**
+ * Admin handler that will help in tracking only the current HTTP requests to Styx.
+ */
+public class CurrentRequestsHandler extends BaseHttpHandler {
+
+    private static final Charset UTF_8 = Charset.forName("UTF-8");
+    private final ThreadMXBean threadMXBean;
+    private CurrentRequestTracker tracker;
+
+    public CurrentRequestsHandler(CurrentRequestTracker tracker) {
+        this.threadMXBean = getThreadMXBean();
+        this.tracker = tracker;
+    }
+
+    @Override
+    public LiveHttpResponse doHandle(LiveHttpRequest request) {
+        boolean withStackTrace = request.queryParam("withStackTrace")
+                .map(it -> "true".equals(it))
+                .orElse(false);
+
+        return HttpResponse
+                .response(OK)
+                .disableCaching()
+                .header(CONTENT_TYPE, PLAIN_TEXT_UTF_8)
+                .body(getCurrentRequestContent(withStackTrace), UTF_8, true)
+                .build()
+                .stream();
+    }
+
+    private String getCurrentRequestContent(boolean withStackTrace) {
+        StringBuilder sb = new StringBuilder();
+        tracker.getCurrentRequests().forEach(req -> {
+            sb.append("[\n");
+            sb.append(req.getRequest().replaceAll(",", "\n"));
+            sb.append("\n\n");
+            sb.append("running for: ");
+            sb.append(currentTimeMillis() - req.getStartingTimeMillies());
+            sb.append("ms\n");
+            if (req.isRequestSent()) {
+                sb.append("\nRequest state: Waiting response from origin.\n");
+            } else {
+                sb.append("\nRequest state: Plugins pipeline.\n");
+                sb.append("\n\n   ...   Thread Info: ...\n");
+                if (withStackTrace) {
+                    sb.append(getThreadInfo(req.getCurrentThread().getId()));
+                } else {
+                    sb.append("Name: ");
+                    sb.append(req.getCurrentThread().getName());
+                    sb.append("\n");
+                }
+            }
+            sb.append("]\n\n");
+        });
+
+        return sb.toString();
+    }
+
+    private String getThreadInfo(long threadId) {
+
+        StringBuilder sb = new StringBuilder();
+
+        final ThreadInfo t = threadMXBean.getThreadInfo(threadId, Integer.MAX_VALUE);
+        sb.append(format("\"%s\" id=%d state=%s", t.getThreadName(), t.getThreadId(), t.getThreadState()));
+        sb.append(getThreadState(t));
+
+        if (t.isSuspended()) {
+            sb.append(" (suspended)");
+        }
+
+        if (t.isInNative()) {
+            sb.append(" (running in native)");
+        }
+
+        sb.append("\n");
+        if (t.getLockOwnerName() != null) {
+            sb.append(format("     owned by %s id=%d%n", t.getLockOwnerName(), t.getLockOwnerId()));
+        }
+
+        sb.append(getThreadElements(t));
+        sb.append("\n");
+
+        sb.append(getThreadLockedSynchronizer(t));
+
+        return sb.toString();
+    }
+
+    private String getThreadState(ThreadInfo t) {
+
+        StringBuilder sb = new StringBuilder();
+
+        final LockInfo lock = t.getLockInfo();
+        if (lock != null && t.getThreadState() != BLOCKED) {
+            sb.append(format("%n    - waiting on <0x%08x> (a %s)", lock.getIdentityHashCode(), lock.getClassName()));
+            sb.append(format("%n    - locked <0x%08x> (a %s)", lock.getIdentityHashCode(), lock.getClassName()));
+        } else if (lock != null && t.getThreadState() == BLOCKED) {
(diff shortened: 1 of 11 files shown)


mikkokar:
Hi Salah, 

And thank you for your contribution!

Overall this is good addition to Styx proxy, you should consider the following:

1) Consider also tracking the `HttpPipelineHandler` state (ACCEPTING_REQUESTS, WAITING_FOR_RESPONSE, SENDING_RESPONSE, etc). This tells you straight away if the request is taking time on the way to the origins, with the origin, or its way back.

2) As an event base system, a stack trace is only valid while an event is being processed. That is, the stack trace is no longer relevant after the request has found its way out of styx towards the origins. Consider making it clear in the admin interface that the request has been proxied and invalidate the stack trace in that case. 

3) The response is not being tracked. Just checking that you are OK with this? Well response tracking could be added as a separate increment. But again if the request is being held back by a response processing, then showing the stack trace (see point 2 above) would be misleading.

As a new feature, this should have a short description in styx user manual: See `docs/user-guide.md`.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/CurrentRequestsHandler.java
```diff
@@ -0,0 +1,149 @@
...
+                sb.append(getThreadInfo(req.getCurrentThread().getId()));
+            } else {
+                sb.append("Name: ");
+                sb.append(req.getCurrentThread().getName());
+                sb.append("\n");
+            }
+            sb.append("]\n\n");
+        });
+
+        return sb.toString().getBytes();
```
mikkokar: `.getBytes` encodes a string using platform's default charset. In theory it could be anything. On the other hand it is important to ensure we are encoding in UTF_8 as advertised in the HTTP response. Therefore I suggest:

- First, modify `getCurrentRequestContent` to return a `String`
- Then, set the body using `.body(String, Charset, Bool)` overload like so:

```        
return FullHttpResponse.response(OK)
                .disableCaching()
                .header(CONTENT_TYPE, PLAIN_TEXT_UTF_8)
                .body(getCurrentRequestContent(withStackTrace), UTF_8, true)
                .build()
                .toStreamingResponse();
```

BTW to keep in line with rest of the styx code, please line up the fluent builder method calls on separate lines as in this example. :-)

salahqasem:
Hi Mikko,

Thanks for the feedback.

For point 1: do you mean registering a new event inside the transition state to log the request state inside the `CurrentRequest` bean? if yes, I think I just need to register `ACCEPTING_REQUESTS` and `WAITING_FOR_RESPONSE` since other statuses need to terminate the request. please correct me if I am wrong.

For point 2&3: Yes, I understand that, the response will not be tracked anymore and it will be removed from the `CurrentRequestTracker`, and the whole idea is only about tracking requests inside Styx itself, so if the request reached out of styx it will not be tracked anymore.

and I will add the documentation for this feature as you mentioned above.

components/server/src/main/java/com/hotels/styx/server/track/CurrentRequest.java
```diff
@@ -0,0 +1,50 @@
...
+import com.hotels.styx.api.HttpRequest;
+
+/**
+ * Bean that represent the current request.
+ */
+public class CurrentRequest {
+
+    private final String request;
+    private final long startingTimeMillies;
+    private Thread currentThread;
```
mikkokar: I think the `currentThread` needs to be volatile, as it is set from the different threads, and read from another.

salahqasem:
Hey Mikko, sorry how I can register the lambda callback in the admin interface?

mikkokar:
Hi Salah, 

Just wanted to clarify the point 2 about invalidating outdated thread information. 

From the request tracker point of view, there are two phases:

1. Request is on its way through the plugin pipeline. The request tracker has a valid stack trace for the request. So the tracker might show something like:

```
[
   HttpRequest{...id=123}
   runningFor: 0ms
   
   ... Thread Info: ...
   stack trace
   ...
]
```

2. The request has been proxied, and Styx is waiting for the response. In this case the request is still "ongoing", but the stack trace is irrelevant. This is because styx is just sitting idle and waiting for the response.  No threads are running.

```
[
   HttpRequest{...id=123}
   runningFor: 0ms
   
   ... Thread Info: ...
   NOT AVAILABLE
   ...
]
```

3. The response has been arrived. The request entry shall be removed from the response tracker.

```
[ empty ]
```

I wanted to bring this up because AFAICS now the implementation jumps from step 1 directly to step 3 without going through the step 2. My concern is that bypassing the step 2 may be misleading, as the users will see the stack trace and think that the response is blocked on a plugin, even if in reality it is held by an origin.

mikkokar:
Hey,

To use a lambda to obtain state:

```
    private State onLegitimateRequest(HttpRequest request, ChannelHandlerContext ctx) {
        statsSink.onRequest(request.id());
        CurrentRequestTracker.INSTANCE.trackRequest(request, () -> this.state().toString());  
```

The lambda parameter itself has a type of `Supplier<String>` etc. Then pass it through and store it
in `CurrentRequest`, which will have a new method:

```
public class CurrentRequest {
   ...

    public CurrentRequest(HttpRequest request, Supplier<String> stateSupplier) {
        ...
        this.stateSupplier = requireNonNull(stateSupplier);
    }


   public String httpPipelineState() {
        return this.stateSupplier.get();
    }
```

Something like that.

salahqasem:
Hi Mikko,

Yes, what is important about the request tracker is phase 1. in this phase you can know if the request delayed in styx or not. as a styx developers, if we get an incident that there is a slowness for requests or something like this, I need to know if the slowness is from my reverse proxy or not. so if I open this admin interface and I didn't find any request stuck inside styx then I can confirm that Styx working as expected, otherwise if there was a requests stuck in the styx then I need to find the root cause of this slowness, so the stack trace will be useful in such case. is that make sense?

salahqasem:
Mikko, do you mean that the request will be not removed from the tracker and the state of it will be on `WAITING_FOR_RESPONSE` ?

mikkokar:
Hey. Yes. That's correct. This is quite meaningful information:

```
[
   HttpRequest{...id=123}
   runningFor: 0ms
   
   ... Thread Info: ...
   NOT AVAILABLE : WAITING RESPONSE FROM ORIGIN
   ...
]
```

salahqasem:
@mikkokar I have fixed the review comments. please take a look when you have time and tell me if I missed anything.
Thanks!

components/proxy/src/main/java/com/hotels/styx/routing/handlers/StandardHttpPipeline.java
```diff
@@ -78,6 +80,9 @@ public StandardHttpPipeline(List<HttpInterceptor> interceptors, HttpHandler hand
 
         @Override
         public StyxObservable<HttpResponse> proceed(HttpRequest request) {
+
+            CurrentRequestTracker.INSTANCE.trackRequest(request);
+
             if (index < interceptors.size()) {
                 HttpInterceptor.Chain chain = new HttpInterceptorChain(this, index + 1);
                 HttpInterceptor interceptor = interceptors.get(index);
```
mikkokar: Hi Salah,

Apologies for getting back to you this late. 

I think there is a potential for a race condition. Let me explain.

* HttpPipelineHandler processes its events in a Netty IO thread. But not all plugins necessarily run on the Netty IO thread. Some of them may run asynchronously in a separate worker thread.

* `HttpPipelineHandler` enters in `WAITING_FOR_RESPONSE` right after it has submitted a received request to Styx HTTP pipeline. However a possible asynchronous plugin might still be processing the request on a different thread. And we'd like to show the stack trace of that plugin.

* A stack trace should be displayed *until* the request has traversed thought the entire styx pipeline. But `WAITING_FOR_RESPONSE` state is not necessarily a guarantee that it has done so.

* The stack trace should be withdrawn only after the request has left the HTTP pipeline.

To address this issue, you need to call the `CurrentRequestTracker` from the tail end of the pipeline with a special method to indicate that the request has been fully processed. For example, in `StandardHttpPipeline.proceed`:

```
        @Override
        public StyxObservable<HttpResponse> proceed(HttpRequest request) {

            CurrentRequestTracker.INSTANCE.trackRequest(request);

            if (index < interceptors.size()) {
                HttpInterceptor.Chain chain = new HttpInterceptorChain(this, index + 1);
                HttpInterceptor interceptor = interceptors.get(index);

                try {
                    return interceptor.intercept(request, chain);
                } catch (Throwable e) {
                    return StyxObservable.error(e);
                }
            }

            // HERE LIKE SO:
            // Notify the tracker that a request has traversed through the pipeline
            CurrentRequestTracker.INSTANCE.trackRequestSent(request);   

            return fromRxObservable(toRxObservable(client.handle(request, this.context))
                    .compose(StandardHttpPipeline::sendErrorOnDoubleSubscription));
        }
```

So I would add a special method to mark the request's departure from the pipeline. This would guarantee that the stack trace is displayed as long as necessary. But not too long.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/CurrentRequestsHandler.java
```diff
@@ -0,0 +1,163 @@
...
+        StringBuilder sb = new StringBuilder();
+        CurrentRequestTracker.INSTANCE.getCurrentRequests().forEach(req -> {
+            sb.append("[\n");
+            sb.append(req.getRequest().replaceAll(",", "\n"));
+            sb.append("\n\n");
+            sb.append("running for: ");
+            sb.append(System.currentTimeMillis() - req.getStartingTimeMillies());
+            sb.append("ms");
+            sb.append("\n\n   ...   Thread Info: ...\n");
+            if ("WAITING_FOR_RESPONSE".equalsIgnoreCase(req.getState())) {
```
mikkokar: See my other comment in `StandardHttpPipeline.proceed` method about pipeline handler state not being a sufficient condition for request being sent. It might still be held by some asynchronous plugin in a separate thread.

So you would have to change this part of the code.

However I would like to see the HTTP pipeline handler state as an additional piece of information:

```
    Styx server state: WAITING_FOR_RESPONSE
    running for: 1ms

    ... Thread Info: ...
    [stack trace for some asynchronous plugin]
```

salahqasem:
Hi @mikkokar I solved the review comments, please review again and tell me if I missed any.

Thanks!

mikkokar:
This PR is now showing changes that have already been published to master. A problem with rebase or merge.

Please rebase your feature branch to the head of the master and let's take it from there. Let me know if you need any help.

salahqasem:
@mikkokar I fixed the rebase issue. please take a look

components/proxy/src/main/java/com/hotels/styx/admin/handlers/CurrentRequestsHandler.java
```diff
@@ -0,0 +1,163 @@
...
+        return HttpResponse
+                .response(OK)
+                .disableCaching()
+                .header(CONTENT_TYPE, PLAIN_TEXT_UTF_8)
+                .body(getCurrentRequestContent(withStackTrace), UTF_8, true)
+                .build()
+                .stream();
+    }
+
+    private String getCurrentRequestContent(boolean withStackTrace) {
```
mikkokar: I would tweak the output bit more:

For proxied requests, which are waiting for origin responses, it currently says:

```
   ...   Thread Info: ...
Thread stack not available: Waiting response from origin.]
```

This looks bit funny as the response state is embedded in the "Thread Info". But in fact the "thread info" actually depends on the request state: The stack trace is relevant only when the response is going through the plugin pipeline. Therefore, let's promote a request state to be displayed *before* the stack trace.

I would remove the stack trace altogether when the request is waiting for a response from the origins. Just display this:

```
Request state: Waiting for response from origin.
```

When a stack trace is relevant (requests in a proxy pipeline), I would include it *after* the request state:

```
Request state: plugins pipeline.
Thread Info:
Thread stack not available: Waiting response from origin.]
```

In fact, even better if this could be formatted out structurally as a JSON string.

salahqasem:
@mikkokar I fixed the latest comments, could you please take a look

salahqasem:
Fixed. please review

PR #316, opened 2018-10-19
Update example plugin to latest api

kvosper:


plugin-examples/src/test/java/com/hotels/styx/ExamplePluginTest.java
```diff
@@ -44,15 +48,16 @@ public void addsExtraHeaders() throws Exception {
...
+            return Eventual.of(response(OK).build());
         };
 
         // an example request you expect your plugin to receive
-        HttpRequest request = get("/foo")
+        LiveHttpRequest request = get("/foo")
                 .build();
 
+
         //The method StyxFutures.await() in styx-common wraps future.get() including appropriate Exception handling.
```
dvlato: StyxFutures is not included in the API, so we should probably remove this comment.
Additionally, we have two utility classes for `CompetableFuture`s in the same module+package,  StyxFutures and CompletableFutures.  Maybe we want to take this chance to fix that.

PR #324, opened 2018-10-24
Upgrade fasterxml version in plugin-examples

kvosper:
Also removed some unused imports.


mikkokar:
Hi @kvosper . Could you please update the commentary for the Example Plugin code?
@dvlato  pointed out that the commentary still refers to the old `Observable` based APIs.

PR #326, opened 2018-10-25
Possible Fix of connection leak

taer:
Just posting this here to validate it's megability and test passing. Want to do more testing today to ensure it's working

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.hotels.styx</groupId>
     <artifactId>styx-parent</artifactId>
-    <version>0.7.10</version>
+    <version>0.7.11-SNAPSHOT</version>
   </parent>
 
   <packaging>pom</packaging>
--- a/components/api/pom.xml
+++ b/components/api/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>com.hotels.styx</groupId>
     <artifactId>styx-components</artifactId>
-    <version>0.7.10</version>
+    <version>0.7.11-SNAPSHOT</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
 
--- a/components/client/pom.xml
+++ b/components/client/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>com.hotels.styx</groupId>
     <artifactId>styx-components</artifactId>
-    <version>0.7.10</version>
+    <version>0.7.11-SNAPSHOT</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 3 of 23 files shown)


taer:
Reproduction instructions for pre-patch:
- One origin that routes to a "slow service". Tested with a backend that takes 6 seconds to respond
- setup origins file with 2 max connections and 2 pending.
- in 5 terminals, run ` while true; do  curl -H"host: hostHeaer" localhost:8080/service?ms=6000; done`
- 4 terminals should work, one should 503. They'll swap around.
- See dash http://localhost:19090/admin/dashboard/index.html   See 2 busy, 2 pending
- kill loops, wait for dash to refresh
- probably have 0 busy, 2 pending.
- repeat curl loops
- eventually, you'll have 2 busy, 2 pending after you kill all the loops

With this patch, pending sometimes hangs out at 2, but no busy connection leak. 

I'll dig more into the leaky pending.

mikkokar:
Hi @taer.

Okay. I have spotted an overlooked issue in request cancellation for pending connections.

Have another look at the `Transport` class. This somewhat modified example below shows the structure of the observable chain as built by `Transport`. An actual implementation is bit different but the underlying reasoning still applies:

```                
return connection(request, origin)
            .flatMap(connection -> {
                Observable<LiveHttpResponse> responseObservable = connection.write(request)
                        .map(response -> addOriginId(originId, response));

                return ResponseEventListener.from(responseObservable)
                        .whenCancelled(() -> closeIfConnected(origin, connection))
                        .whenResponseError(cause -> closeIfConnected(origin, connection))
                        .whenContentError(cause -> closeIfConnected(origin, connection))
                        .whenCompleted(() -> returnIfConnected(origin, connection))
                        .apply();
            });
    }
```

Note that a call to `connection(request, origin)` returns a connection observable. It emits a `connection` when one becomes available. A `connection` is emitted immediately if there is one available in the pool. But if not (subscriber is "pending"), there could be a significant delay before a `connection` is emitted. If ever.

When a `connection` becomes available, it is transformed (`flatMap`) to a response observable. The request cancellation logic (`whenCancelled`) is applied to this response observable.

Now, consider what happens when Styx server (`HttpPipelineHandler`) cancels the request by unsubscribing from the response observable. Regardless of the underlying reactive framework (let it be Rx.Java or Reactor core) an `unsubscribe` signal gets delivered up the observable/publisher chain to the original publisher. 

For active connections, the response event listener picks up the unsubscription event and calls the relevant callback. But for pending connections, a `connection` is not yet available. Therefore the `flatMap` has not yet been called, and the response event listener has not yet been created.

There should be something in the connection pool that detects an `unsubscribe` event and removes the waiting subscriber accordingly.

taer:
No need for 0.7 patches anymore. We're on 1.0

PR #327, opened 2018-10-25
Reject requests using the CONNECT method ( Fixes #312)

### LINKED ISSUE #312: Reject the CONNECT method in Styx Proxy.
## The problem
HTTP requests using the CONNECT method should be rejected by Styx proxy with a 405 Method not allowed response code.  

## Detailed description
Styx proxy does not currently provide any support for CONNECT requests so requests with that method should be rejected (with the 405). The current faulty behaviour is that Styx accepts the requests but fails to process it, generating a 500 server side error. 

The status of support for TCP tunneling via the HTTP Connect  method is tracked in #311.

## Acceptance criteria
The proxy will return a 405 error on reception of a HTTP Request using the CONNECT method.

dvlato:


components/proxy/src/main/java/com/hotels/styx/startup/StyxPipelineFactory.java
```diff
@@ -90,11 +91,12 @@ private static HttpHandler styxHttpPipeline(StyxConfig config, HttpHandler inter
             builder.add(new HttpMessageLoggingInterceptor(longFormatEnabled));
         }
 
-        builder.add(new ConfigurationContextResolverInterceptor(EMPTY_CONFIGURATION_CONTEXT_RESOLVER));
-        builder.add(new UnexpectedRequestContentLengthRemover());
-        builder.add(new ViaHeaderAppendingInterceptor());
-        builder.add(new HopByHopHeadersRemovingInterceptor());
-        builder.add(new RequestEnrichingInterceptor(config.styxHeaderConfig()));
+        builder.add(new TcpTunnelRequestRejector()).
```
lukaszdrynski: it seems that you use different convention with the dot at the end instead of at the front

components/proxy/src/main/java/com/hotels/styx/proxy/interceptors/TcpTunnelRequestRejector.java
```diff
@@ -0,0 +1,46 @@
...
+    public Eventual<LiveHttpResponse> intercept(LiveHttpRequest request, Chain chain) {
+
+        if (HttpMethod.CONNECT.equals(request.method())) {
+            return Eventual.of(response(HttpResponseStatus.METHOD_NOT_ALLOWED).build());
+        } else {
+            return chain.proceed(request);
+        }
+    }
+
+
```
mikkokar: Overall this is what needs to be done. However please follow styx coding standards, specifically use static imports with METHOD_NOT_ALLOWED, and CONNECT. Also I would remove the extra empty lines.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/ProxySpec.scala
```diff
@@ -245,4 +245,16 @@ class ProxySpec extends FunSpec
     }
   }
 
+  describe("Handling CONNECT method") {
+    it("Should reject CONNECT requests") {
+      val request = get("/connectTest")
+        .addHeader(HOST, styxServer.proxyHost)
+        .method(HttpMethod.CONNECT)
```
mikkokar: Static import if possible.

PR #328, opened 2018-10-25
Add Transformer class for live HTTP objects

mikkokar:
To make Styx APIs safer for the end-users, this PR adds a separate `Transformer` alongside `Builder` in `LiveHttpRequest` and `LiveHttpResponse` classes. The transformer restricts what the API consumers can do 1) when creating a new live HTTP object and 2) when transforming one.

Some examples:

It is no longer possible to replace, and thus disconnect the reactive `ByteStream` chain like so:

        response().build()
                .newBuilder()
                .body(new ByteStream(...)).   // No longer available
                .build();

The transformer class now enforces API consumers to *transform* the byte stream with the lambda alternative. Like so:

        response().build()
                .newBuilder()
                .body(ByteStream::drop)
                .build();

Likewise it is no longer possible to supply a mapper lambda when creating a live object. This is no longer possible:

        response()
                .body(stream -> doSomething(stream))  // No longer available
                .build();


components/api/src/main/java/com/hotels/styx/api/LiveHttpResponse.java
```diff
@@ -258,10 +258,299 @@ public boolean equals(Object obj) {
...
+        /**
+         * Removes the header with the specified name.
+         *
+         * @param name The name of the header to remove
+         * @return {@code this}
+         */
+        BuilderTransformer removeHeader(CharSequence name);
+
+        /**
+         * (UNSTABLE) Removes body stream from this request.
```
dvlato: is this still applicable? I would anyway like to see a more specific description than ''stability" issues.

mikkokar: No I don't think so. This is safe to use since the `ByteStream` abstraction was added.
I will remove this comment accordingly.

PR #329, opened 2018-10-25
Remove mutable state (connectionRef) from Transport class.

taer:
Forward port the fix in https://github.com/HotelsDotCom/styx/pull/326 to master

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/Transport.java
+++ b/components/client/src/main/java/com/hotels/styx/client/Transport.java
@@ -15,16 +15,15 @@
  */
 package com.hotels.styx.client;
 
+import com.hotels.styx.api.Id;
 import com.hotels.styx.api.LiveHttpRequest;
 import com.hotels.styx.api.LiveHttpResponse;
-import com.hotels.styx.api.Id;
 import com.hotels.styx.api.ResponseEventListener;
 import com.hotels.styx.api.exceptions.NoAvailableHostsException;
 import com.hotels.styx.client.connectionpool.ConnectionPool;
 import rx.Observable;
 
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicReference;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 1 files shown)


components/client/src/main/java/com/hotels/styx/client/Transport.java
```diff
@@ -43,37 +42,31 @@ public Transport(Id appId, CharSequence originIdHeaderName) {
...
+                    Observable<LiveHttpResponse> responseObservable = connection.write(request)
+                            .map(response -> addOriginId(originId, response));
+
+                    return ResponseEventListener.from(responseObservable)
+                            .whenCancelled(() -> closeIfConnected(origin, connection))
+                            .whenResponseError(cause -> closeIfConnected(origin, connection))
+                            .whenContentError(cause -> closeIfConnected(origin, connection))
+                            .whenCompleted(() -> returnIfConnected(origin, connection))
+                            .apply();
+                });
```
mikkokar: Hi @taer 

The observable should be constructed like this:

```            @Override
            public Observable<LiveHttpResponse> response() {
                return connection.flatMap(connection -> {
                    Observable<LiveHttpResponse> responseObservable = connection.write(request)
                            .map(response -> addOriginId(originId, response));

                    return ResponseEventListener.from(responseObservable)
                            .whenCancelled(() -> closeIfConnected(origin, connection))
                            .whenResponseError(cause -> closeIfConnected(origin, connection))
                            .whenContentError(cause -> closeIfConnected(origin, connection))
                            .whenCompleted(() -> returnIfConnected(origin, connection))
                            .apply();
                });
            }
```

The `ResponseEventListener.from` takes a response observable and decorates it with event listeners triggering  at relevant points of the response lifecycle. It takes into account both response observable and its associated body byte stream. The `apply` method returns a new `Observable<LiveHttpResponse>` that should be returned for the downstream observers.

Also, I have pushed this to your branch. Please pull the lates changes to have a look!

mikkokar:
Hi @taer. This is a good PR even if it doesn't actually fix the leak. Therefore I am going to rename the PR title accordingly and merge.

For the record, the root cause for the connection leak is lack of cancel/`unsubscribe`handling for connection observables. More specifically, `Transport` class handles cancellations for response observables, which only works *after* a connection event is published from the connection observable. The cancellation therefore doesn't have any effect when subscriber (connection) is pending. 

As ultimate producers of connections, unsubscriptions must be dealt with in the connection pool, and in `NettyConnectionFactory`.

PR #334, opened 2018-11-08
Remove hamcrest from public dependency list

taer:
MorePreconditions was relying on hamcrest-all. But no one was calling the method in question.


taer:
Not ready yet

PR #338, opened 2018-11-09
Refactor StyxHostHttpClient

mikkokar:
Move functionality up into `StyxBackendServiceClient`:
- Origin ID header attachment
- Application name in the exceptions

Rationale: to make it easier to expose `StyxHostHttpClient` as a reusable component for 3rd party projects. Specifically, this improvement removes classes like `Origin` from the API and removes Styx proxy specific functionality like attaching origin IDs to HTTP responses.

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.java
+++ b/components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.java
@@ -48,6 +48,7 @@ import static com.google.common.collect.Lists.newArrayList;
 import static com.hotels.styx.api.HttpHeaderNames.CONTENT_LENGTH;
 import static com.hotels.styx.api.HttpHeaderNames.TRANSFER_ENCODING;
 import static com.hotels.styx.api.extension.service.StickySessionConfig.stickySessionDisabled;
+import static com.hotels.styx.client.StyxHeaderConfig.ORIGIN_ID_DEFAULT;
 import static com.hotels.styx.client.stickysession.StickySessionCookie.newStickySessionCookie;
 import static io.netty.handler.codec.http.HttpMethod.HEAD;
 import static java.util.Collections.emptyList;
@@ -74,6 +75,7 @@ public final class StyxBackendServiceClient implements BackendServiceClient {
     private final boolean contentValidation;
     private final String originsRestrictionCookieName;
     private final StickySessionConfig stickySessionConfig;
+    private final CharSequence originIdHeader;
 
     private StyxBackendServiceClient(Builder builder) {
         this.id = requireNonNull(builder.backendServiceId);
@@ -93,6 +95,7 @@ public final class StyxBackendServiceClient implements BackendServiceClient {
         this.metricsRegistry = builder.metricsRegistry;
         this.contentValidation = builder.contentValidation;
         this.originsRestrictionCookieName = builder.originsRestrictionCookieName;
+        this.originIdHeader = builder.originIdHeader;
     }
 
     @Override
@@ -146,8 +149,7 @@ public final class StyxBackendServiceClient implements BackendServiceClient {
             newPreviousOrigins.add(remoteHost.get());
 
             return ResponseEventListener.from(
-                        toObservable(host.hostClient().handle(request, HttpInterceptorContext.create()))
-                        .map(response -> addStickySessionIdentifier(response, host.origin())))
+                        toObservable(host.hostClient().handle(request, HttpInterceptorContext.create())).map(response -> addStickySessionIdentifier(response, host.origin())))
                     .whenResponseError(cause -> logError(request, cause))
                     .whenCancelled(() -> originStatsFactory.originStats(host.origin()).requestCancelled())
                     .apply()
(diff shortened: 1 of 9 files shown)


components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.java
```diff
@@ -146,8 +149,7 @@ private static boolean isHeadRequest(LiveHttpRequest request) {
             newPreviousOrigins.add(remoteHost.get());
 
             return ResponseEventListener.from(
-                        toObservable(host.hostClient().handle(request, HttpInterceptorContext.create()))
-                        .map(response -> addStickySessionIdentifier(response, host.origin())))
+                        toObservable(host.hostClient().handle(request, HttpInterceptorContext.create())).map(response -> addStickySessionIdentifier(response, host.origin())))
```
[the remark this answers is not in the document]
mikkokar: 👍

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.java
```diff
@@ -157,13 +159,29 @@ private static boolean isHeadRequest(LiveHttpRequest request) {
...
+                    .onErrorResumeNext(this::mapExceptionCauses);
         } else {
             RetryPolicyContext retryContext = new RetryPolicyContext(this.id, attempt + 1, null, request, previousOrigins);
             return retry(request, retryContext, previousOrigins, attempt + 1, new NoAvailableHostsException(this.id));
         }
     }
 
+    private Observable<LiveHttpResponse> mapExceptionCauses(Throwable cause) {
+        if (cause instanceof NoAvailableHostsException) {
+            return Observable.error(new NoAvailableHostsException(id));
```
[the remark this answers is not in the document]
mikkokar: The stack trace seems bit pointless for this exception because we know where and why the exception is originally thrown.

However this comment highlights an interesting issue: The `NoAvailableHostsException` is thrown from the `Transport` class when it doesn't have a connection pool configured. But then again it doesn't make sense to create a `Transport` object without pool. Instead, the exception should be thrown earlier. Before the `Transport` is even instantiated.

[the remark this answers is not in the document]
mikkokar: Ok. This is effectively dead code. It used to be the case that `Transport` checked if the chosen origin/pool was available or not. But since refactoring `StyxBackendServiceClient` it is no longer possible to use transport/send without providing a connection pool.

So this exception mapping can be removed.

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.java
```diff
@@ -383,12 +400,16 @@ public Builder originsRestrictionCookieName(String originsRestrictionCookieName)
             return this;
         }
 
+        public Builder originIdHeader(CharSequence originIdHeader) {
+            this.originIdHeader = originIdHeader;
```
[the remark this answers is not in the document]
mikkokar: 👍

components/client/src/main/java/com/hotels/styx/client/Transport.java
```diff
@@ -78,13 +67,7 @@ private synchronized void returnIfConnected(Optional<ConnectionPool> connectionP
                 .orElseGet(() -> {
                     // Aggregates an empty body:
                     request.consume();
-                    return Observable.error(new NoAvailableHostsException(appId));
+                    return Observable.error(new NoAvailableHostsException(Id.id("")));
```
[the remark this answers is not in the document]
mikkokar: This has been removed.

components/client/src/main/java/com/hotels/styx/client/Transport.java
```diff
@@ -15,76 +15,41 @@
...
-                }
+            private synchronized void closeIfConnected(ConnectionPool connectionPool, Connection connection) {
+                connectionPool.closeConnection(connection);
             }
 
-            private synchronized void returnIfConnected(Optional<ConnectionPool> connectionPool, Connection connection) {
-                if (connection != null && connectionPool.isPresent()) {
-                    connectionPool.get().returnConnection(connection);
-                }
+            private synchronized void returnIfConnected(ConnectionPool connectionPool, Connection connection) {
```
dvlato: Don't we ever need to synchronize on something different than the current object? In that case, do we actually need to set 'closeIfConnected' and/or 'returnIfConnected' as synchronized methods or should we aim to ensure they are idempotent?

mikkokar: Hey David. There is no longer need to synchronise on anything. I realised myself the `synchronized` modifiers were left in the methods. They should have been removed already. Perhaps you have commented an outdated version?

mikkokar: Thinking about it, perhaps I should remove these methods altogether.

dvlato: oh, sure, I see your commit now right before my comment.   You can remove them but I don't see too much of a difference between both options to be honest... I don't dislike the 'extra indirection' myself.

components/client/src/main/java/com/hotels/styx/client/Transport.java
```diff
@@ -15,76 +15,41 @@
...
+                            .whenCancelled(() -> closeIfConnected(connectionPool, connection))
+                            .whenResponseError(cause -> closeIfConnected(connectionPool, connection))
+                            .whenContentError(cause -> closeIfConnected(connectionPool, connection))
+                            .whenCompleted(() -> returnIfConnected(connectionPool, connection))
                             .apply();
                 });
             }
 
-            private synchronized void closeIfConnected(Optional<ConnectionPool> connectionPool, Connection connection) {
-                if (connection != null && connectionPool.isPresent()) {
```
kvosper: Have we confirmed that the connections are never null? It would make sense, but thought it worth checking.

mikkokar: A `null` value wouldn't be emitted under any circumstances (that would be a bug elsewhere in styx). 

Note that the Reactive Streams specification specifically forbids `null` values as events (Rule 2.13 https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.2/README.md). Although we are still in RxJava 1.0, our intention is to fully migrate to reactive-streams compatible implementation (Flux), ensuring it cannot be null.

PR #339, opened 2018-11-12
Refactor: Simplify StyxHostHttpClient

mikkokar:
Remove `Transport` and `HttpTransaction` classes from codebase.

Improve `ResponseEventListener` by detecting empty HTTP response observables. It now calls an `whenResponseError()` callback if a response observable completes without first emitting response headers (`LiveHttpResponse`).


components/client/src/test/unit/java/com/hotels/styx/client/StyxHostHttpClientTest.java
```diff
@@ -15,55 +15,202 @@
...
+                .expectNextCount(0)
+                .expectComplete()
+                .log()
+                .verify();
+
+        verify(pool).closeConnection(any(Connection.class));
+    }
+
+    @Test
+    public void releasesConnectionWhenRequestFailsBeforeHeaders() {
```
kvosper: Looks like this might be a mistake?
```suggestion
    public void releasesConnectionWhenResponseFailsBeforeHeaders() {
```

components/common/src/test/java/com/hotels/styx/api/ResponseEventListenerTest.java
```diff
@@ -167,5 +168,19 @@ public void firesResponseContentError() {
...
 
+        ResponseEventListener.from(publisher)
+                .whenResponseError(cause -> responseError.set(cause))
+                .apply()
+                .toBlocking()
+                .subscribe(testSubscriber);
+
+        testSubscriber.awaitTerminalEvent();
+        assertEquals(testSubscriber.getOnCompletedEvents().size(), 1);
+        assertTrue(responseError.get() instanceof RuntimeException);
```
kvosper: We could use `throw responseError.get();` in order to use the expectedException annotation.  What do you think?

components/client/src/test/unit/java/com/hotels/styx/client/StyxBackendServiceClientTest.java
```diff
@@ -98,7 +98,7 @@ public void setUp() {
 
     @Test
     public void sendsRequestToHostChosenByLoadBalancer() {
-        StyxHostHttpClient hostClient = mockHostClient(just(response(OK).build()));
+        StyxHostHttpClient hostClient = mockHostClient(Flux.just(response(OK).build()));
```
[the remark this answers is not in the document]
mikkokar: It is better to keep the `Flux` scope to distinguish from others like `Mono`.

components/client/src/test/unit/java/com/hotels/styx/client/StyxBackendServiceClientTest.java
```diff
@@ -224,10 +224,10 @@ public void stopsRetriesWhenRetryPolicyTellsToStop() {
 
     @Test
     public void retriesAtMost3Times() {
-        StyxHostHttpClient firstClient = mockHostClient(Observable.error(new OriginUnreachableException(ORIGIN_1, new RuntimeException("An error occurred"))));
-        StyxHostHttpClient secondClient = mockHostClient(Observable.error(new OriginUnreachableException(ORIGIN_2, new RuntimeException("An error occurred"))));
-        StyxHostHttpClient thirdClient = mockHostClient(Observable.error(new OriginUnreachableException(ORIGIN_3, new RuntimeException("An error occurred"))));
-        StyxHostHttpClient fourthClient = mockHostClient(Observable.error(new OriginUnreachableException(ORIGIN_4, new RuntimeException("An error occurred"))));
+        StyxHostHttpClient firstClient = mockHostClient(Flux.error(new OriginUnreachableException(ORIGIN_1, new RuntimeException("An error occurred"))));
```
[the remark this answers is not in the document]
mikkokar: It is better to keep the `Flux` scope to distinguish from other publishers like `Mono`.

components/common/src/test/java/com/hotels/styx/api/ResponseEventListenerTest.java
mikkokar: I would leave the `expectedException` annotation for situations where an exception is actually *thorwn*. But here the exception is merely emitted through the publisher instead of being thrown. Therefore I would leave the code as it is.

kvosper: TestNG also has an annotation that can check the message of the exception, but it we don't need to do that here, no problem.

PR #340, opened 2018-11-13
Remove singleton dependency from request tracker unit tests.

mikkokar:
Also tidy up the code a bit.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/CurrentRequestsHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/CurrentRequestsHandler.java
@@ -15,13 +15,11 @@
  */
 package com.hotels.styx.admin.handlers;
 
-import static com.google.common.net.MediaType.PLAIN_TEXT_UTF_8;
-import static com.hotels.styx.api.HttpHeaderNames.CONTENT_TYPE;
-import static com.hotels.styx.api.HttpResponseStatus.OK;
-import static java.lang.management.ManagementFactory.getThreadMXBean;
-import static java.lang.Thread.State.BLOCKED;
-import static java.lang.System.currentTimeMillis;
-import static java.lang.String.format;
+import com.hotels.styx.api.HttpResponse;
+import com.hotels.styx.api.LiveHttpRequest;
+import com.hotels.styx.api.LiveHttpResponse;
+import com.hotels.styx.common.http.handler.BaseHttpHandler;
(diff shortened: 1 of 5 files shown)


components/proxy/src/main/java/com/hotels/styx/admin/handlers/CurrentRequestsHandler.java
```diff
@@ -15,34 +15,34 @@
...
+import static java.lang.System.currentTimeMillis;
+import static java.lang.Thread.State.BLOCKED;
+import static java.lang.management.ManagementFactory.getThreadMXBean;
 
 /**
  * Admin handler that will help in tracking only the current HTTP requests to Styx.
  */
 public class CurrentRequestsHandler extends BaseHttpHandler {
-
     private static final Charset UTF_8 = Charset.forName("UTF-8");
```
dvlato: I think we are using StandardCharsets.UTF_8 elsewhere (with a static import)

components/server/src/main/java/com/hotels/styx/server/track/CurrentRequestTracker.java
```diff
@@ -63,8 +58,4 @@ public void endTrack(LiveHttpRequest request) {
     public Collection<CurrentRequest> getCurrentRequests() {
```
dvlato: Per our code conventions this should probably be named "currentRequests()"

components/server/src/main/java/com/hotels/styx/server/track/CurrentRequestTracker.java
```diff
@@ -15,23 +15,19 @@
...
 /**
  * Manger class to manage the current requests.
  */
 public class CurrentRequestTracker implements RequestTracker {
-
-    private static ConcurrentHashMap<Object, CurrentRequest> currentRequests = new ConcurrentHashMap<>();
-
     public static final CurrentRequestTracker INSTANCE = new CurrentRequestTracker();
 
-    private CurrentRequestTracker() {
```
dvlato: Instead of letting the CurrentRequestTracker be the default, I would rather set the appropriate visibility (hopefully package private will suffice) and annotate the method with  @VisibleForTesting.  Probably we should redesign something if need public visibility as this is supposed to be a singleton...

dvlato: That's quite unfortunate.  I would still at least create a explicit constructor with the @VisibleForTesting annotation. Another option would be to have a private constructor and a static factory method that has the visible for testing annotation.

PR #342, opened 2018-11-14
Additional metrics

kvosper:
Publishing additional metrics provided by OperatingSystemMXBean.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/sets/OperatingSystemMetricSet.java
+++ b/components/proxy/src/main/java/com/hotels/styx/metrics/reporting/sets/OperatingSystemMetricSet.java
@@ -1,0 +1,66 @@
+/*
+  Copyright (C) 2013-2018 Expedia Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package com.hotels.styx.metrics.reporting.sets;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricSet;
+import com.sun.management.OperatingSystemMXBean;
+import com.sun.management.UnixOperatingSystemMXBean;
+
+import java.lang.management.ManagementFactory;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Creates a set of gauges that monitor metrics provided by the JVM regarding the OS.
+ */
+public final class OperatingSystemMetricSet implements MetricSet {
+    private final OperatingSystemMXBean bean;
+
+    public OperatingSystemMetricSet() {
+        this.bean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    }
+
+    @Override
+    public Map<String, Metric> getMetrics() {
+        Map<String, Gauge<?>> gauges = new HashMap<>();
+
+        castIfInstance(bean, UnixOperatingSystemMXBean.class).ifPresent(unixBean -> {
+            gauges.put("fileDescriptors.max", unixBean::getMaxFileDescriptorCount);
+            gauges.put("fileDescriptors.open", unixBean::getOpenFileDescriptorCount);
+        });
+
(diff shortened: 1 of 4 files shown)


mikkokar:
Please ensure that documentation is updated accordingly, if necessary.

kvosper:
> Please ensure that documentation is updated accordingly, if necessary.

I have made the documentation consistent with the other `jvm.` metrics.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/sets/OperatingSystemMetricSet.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public final class OperatingSystemMetricSet implements MetricSet {
+    private final OperatingSystemMXBean bean;
+
+    public OperatingSystemMetricSet() {
+        this.bean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    }
+
+    @Override
+    public Map<String, Metric> getMetrics() {
```
dvlato: Can we declare this as  "? extends Metric"? I haven't checked MetricSet.

kvosper: It doesn't compile that way, I'm afraid.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/sets/OperatingSystemMetricSet.java
```diff
@@ -0,0 +1,66 @@
...
+
+        gauges.put("memory.physical.free", bean::getFreePhysicalMemorySize);
+        gauges.put("memory.physical.total", bean::getTotalPhysicalMemorySize);
+        gauges.put("memory.virtual.committed", bean::getCommittedVirtualMemorySize);
+
+        gauges.put("swapSpace.free", bean::getFreeSwapSpaceSize);
+        gauges.put("swapSpace.total", bean::getTotalSwapSpaceSize);
+
+        // Using the "wrong" type then casting it allows us to avoid writing (Gauge<...>) in every single "put" call above
+        return (Map) gauges;
```
dvlato: If we cannot change the return type, maybe add a 'supresswarnings' for this statement.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/sets/OperatingSystemMetricSet.java
```diff
@@ -0,0 +1,66 @@
...
+        Map<String, Gauge<?>> gauges = new HashMap<>();
+
+        castIfInstance(bean, UnixOperatingSystemMXBean.class).ifPresent(unixBean -> {
+            gauges.put("fileDescriptors.max", unixBean::getMaxFileDescriptorCount);
+            gauges.put("fileDescriptors.open", unixBean::getOpenFileDescriptorCount);
+        });
+
+        gauges.put("cpu.process.load", bean::getProcessCpuLoad);
+        gauges.put("cpu.process.time", bean::getProcessCpuTime);
+        gauges.put("cpu.system.load", bean::getSystemCpuLoad);
```
mikkokar: It would seem more logical to call them `os.process.cpu.load` and `os.system.cpu.load` etc. WDYT?

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/sets/OperatingSystemMetricSet.java
mikkokar: The comment could be clearer. Consider

> Using a `Map<String, Gauge<?>> to avoid casting each metric provider to `Gauge<?>` above, but at the expense of having to cast the return value to `Map`.

components/proxy/src/test/java/com/hotels/styx/startup/CoreMetricsTest.java
```diff
@@ -120,7 +120,20 @@ public void registersJvmMetrics() {
...
+
+                "jvm.os.cpu.process.load",
+                "jvm.os.cpu.process.time",
+                "jvm.os.cpu.system.load",
+                "jvm.os.memory.physical.free",
+                "jvm.os.memory.physical.total",
+                "jvm.os.memory.virtual.committed",
+                "jvm.os.swapSpace.free",
+                "jvm.os.swapSpace.total"
+                ));
```
mikkokar: It feels odd that the operating system metrics are scoped under JVM metrics. I would consider moving them to the root level. Like so:

`os.swapSpace.free` or `os.memory.virtual.committed`.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/sets/OperatingSystemMetricSet.java
kvosper: They are both measuring CPU load but differ only in whether it is just for the process or the whole system. It's not like there are separate CPUs for system and process.

PR #343, opened 2018-11-14
Deprecate Rx.Java from ConnectionPool and its consumers

mikkokar:
Migrate `ConnectionPool` interface and its consumers from Rx.Java to Reactor/Flux. 
Also migrate `BackendServiceClient` and `ResponseEventListener` classes to Reactor/Flux.

### DIFF
--- a/components/common/src/test/java/com/hotels/styx/api/ResponseEventListenerTest.java
+++ b/components/common/src/test/java/com/hotels/styx/api/ResponseEventListenerTest.java
@@ -15,24 +15,20 @@
  */
 package com.hotels.styx.api;
 
+import org.reactivestreams.Publisher;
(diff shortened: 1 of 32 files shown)


components/common/src/test/java/com/hotels/styx/api/ResponseEventListenerTest.java
```diff
@@ -53,97 +49,96 @@ public void setUp() {
 
     @Test
     public void doesntFireUnnecessaryEvents() {
```
kvosper: Could we include some description (either in the name or a comment) of what makes these events unnecessary? Is it just saying that the events being fired never happened, e.g.
```suggestion
    public void doesntFireEventsThatNeverOccurred() {
```
?

PR #349, opened 2018-11-21
ByteStream replacement in live HTTP response.

mikkokar:
Allows response body byte stream to be replaced.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/ByteStream.java
+++ b/components/api/src/main/java/com/hotels/styx/api/ByteStream.java
@@ -15,6 +15,7 @@
  */
 package com.hotels.styx.api;
 
+import io.netty.buffer.Unpooled;
 import org.reactivestreams.Publisher;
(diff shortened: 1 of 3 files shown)


components/api/src/main/java/com/hotels/styx/api/ByteStream.java
```diff
@@ -50,6 +51,17 @@ public ByteStream(Publisher<Buffer> stream) {
...
     }
 
+    /**
+     * Creates a new {@code ByteStream} from String.
+     *
+     * @param content ByteStream source
+     * @param charset Character set encoding
+     * @return ByteStream
+     */
+    public static ByteStream from(String content, Charset charset) {
```
dvlato: Do you think we should have a factory to construct the object from the String instead of from a byte source?

mikkokar: Yes I think we should. I'll add this.

PR #353, opened 2018-11-30
StyxHttpClient: add support for live requests

mikkokar:
Add support for sending LiveHttpRequests.

It is now possible to:

    client.streaming().send(liveRequest)


components/client/src/main/java/com/hotels/styx/client/connectionpool/ConnectionPools.java
```diff
@@ -59,6 +62,18 @@ public static ConnectionPool poolForOrigin(Origin origin, MetricRegistry metricR
                 metricRegistry);
     }
 
+    public static ConnectionPool create(String hostname, int port, ConnectionPoolSettings poolSettings) {
```
dvlato: Where is this used?

mikkokar: Nowhere yet.

PR #355, opened 2018-12-06
Adapt Version metric to a more flexible pattern 

dvlato:
Some users want to see the version value in graphite and that server does not support non-numeric metrics. Thus, we will send the metric as a number, but supporting build numbers separated with a Hyphen instead of a dot.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/Version.java
+++ b/components/proxy/src/main/java/com/hotels/styx/Version.java
@@ -88,7 +88,7 @@ public class Version {
 
     private Optional<String> extractFinalInt(String versionString) {
         Matcher m = VERSION_FORMAT.matcher(versionString);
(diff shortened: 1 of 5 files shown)


components/proxy/src/main/java/com/hotels/styx/Version.java
```diff
@@ -74,7 +80,28 @@ public String releaseVersion() {
         return releaseTag.substring(firstDot + 1);
     }
 
+    public Optional<Integer> buildNumber() {
```
mikkokar: Is this used anywhere beyond tests?

dvlato: umm the changes to CoreMetrics were uncommited... it should be fixed now.

PR #357, opened 2018-12-07
Use Long to represent Content-Length

mikkokar:


components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -571,7 +571,7 @@ Builder ensureContentLengthIsValid() {
 
         private static boolean isInteger(String contentLength) {
             try {
-                parseInt(contentLength);
+                parseLong(contentLength);
```
dvlato: do you think we should validate it is a positive integer / greater than 0?

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -728,10 +728,10 @@ private void ensureContentLengthIsValid() {
             return headerValues.isEmpty() ? Optional.empty() : Optional.of(headerValues.get(0));
         }
 
-        private static boolean isInteger(String contentLength) {
+        private static boolean isPositiveInteger(String contentLength) {
```
kvosper:
```suggestion
        private static boolean isNonNegativeInteger(String contentLength) {
```
Generally the term "positive integer" refers to an integer greater than zero. I'm not sure if there is an official ruling for numbers in Java, but this is the use of the term I am familiar with.

components/api/src/main/java/com/hotels/styx/api/LiveHttpRequest.java
```diff
@@ -1017,7 +1017,7 @@ private void ensureContentLengthIsValid() {
 
         private static boolean isInteger(String contentLength) {
```
kvosper: So live requests + responses do not have the `>= 0` check that full ones have?

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
mikkokar: Done

components/api/src/main/java/com/hotels/styx/api/LiveHttpRequest.java
mikkokar: My mistake. Got distracted when implementing this. :-/

Added.

PR #363, opened 2018-12-20
Don't deploy unnecessary modules

mikkokar:
Don't deploy JARs for `api-testsupport`, `origins-starter-app` or `testsupport` modules.

### DIFF
--- a/support/api-testsupport/pom.xml
+++ b/support/api-testsupport/pom.xml
@@ -51,4 +51,17 @@
     </dependency>
 
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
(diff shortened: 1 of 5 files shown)


support/api-testsupport/pom.xml
```diff
@@ -51,4 +51,17 @@
...
 
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
dvlato: Some of our users are declaring dependencies to styx-api-testsupport, can we remove the deployment for it? Also both styx-api-testsupport and styx-testsupport are referenced in the BOM file.

mikkokar: Ok. I'll review the styx-api-testsupport, and remove the two others from the BOM.

mikkokar: Hi @dvlato. I have put back styx-api-testsupport for now. Although it ought to go in future. I have also edited the BOM accordingly.

PR #366, opened 2018-12-21
Create a docker image from an official Styx release.

mikkokar:
This PR adds a "public" docker image built from an official Styx release. We could add this to a hub.docker.com. Going forward the docker build should be automated as a part of release process. Also there is a separate docker file for the development version. But hey, this is the first step!

Also need to think if a separate repository is better for the public docker file.

Build instructions:

`$ cd docker`
`$ docker build -t styx .`


docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ADD https://github.com/HotelsDotCom/styx/releases/download/${BASENAME}/${BASENAME}-linux-x86_64.zip ${APP_HOME}/styx.zip
+COPY default-docker.yml /styx/default-config/default.yml
+COPY styx-env.sh /styx/default-config/styx-env.sh
+COPY origins.yml /styx/default-config/origins.yml
+
+RUN unzip styx.zip \
+    && mv ${BASENAME} styx \
+    && rm styx.zip
+
+EXPOSE 8080 8443 9000
```
dvlato: We might want to expose the JMX port as well.

docker/origins.yml
```diff
@@ -0,0 +1,12 @@
+# See origins/origins-default.yaml for explanation of config format
+---
+- id: "app"
```
dvlato: We are missing some instructions on how to use and configure the Docker image.

docker/styx-env.sh
```diff
@@ -0,0 +1,38 @@
...
+JVM_OTHER_OPTS="${JVM_OTHER_OPTS:=-XX:+DisableExplicitGC -Djava.awt.headless=true -XX:+OptimizeStringConcat -XX:+UseBiasedLocking}"
+
+# Increase Netty String builder maximum size to 70KB.
+# See: https://github.com/netty/netty/issues/7092
+JAVA_OPTS="$JAVA_OPTS -Dio.netty.threadLocalMap.stringBuilder.maxSize=71680"
+
+# Set Netty leak detection level
+# JAVA_OPTS="$JAVA_OPTS -Dio.netty.leakDetectionLevel=advanced"
+
+# Enable remote debugging
```
dvlato: Remove commented out config which doesn't need to be documented.

PR #373, opened 2019-01-17
Improve log message on plugin load fail/Decouple failure handling from task execution

### LINKED ISSUE #372: Improve log message when plugin(s) cannot be loaded
## The problem
Currently the message just says `java.lang.RuntimeException: 1 plugins could not be loaded`

## Detailed description
The current message is not useful because it does not tell us which message or why

## Acceptance criteria
* Include a list of names of plugins that failed
* Include at least the exception messages for each failed plugin

kvosper:
This PR introduces a common class called `FailureHandlingStrategy` that can be used to decouple failure-handling logic from the logic of processing items in a list.

Demonstrating this decoupling are the changes in `PluginSuppliers`, which also fixes #372.

### DIFF
--- a/components/common/src/main/java/com/hotels/styx/common/FailureHandlingTaskProcessor.java
+++ b/components/common/src/main/java/com/hotels/styx/common/FailureHandlingTaskProcessor.java
@@ -1,0 +1,144 @@
+/*
+  Copyright (C) 2013-2019 Expedia Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package com.hotels.styx.common;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Allows various types of failure handling to be applied to a task running over a list of inputs.
+ *
+ * It can be as strict or lenient as desired - from failing the whole run immediately as soon as a single iteration fails,
+ * to just logging failures and returning successful results.
+ *
+ * The behaviour is configured by injecting lambdas when an instance of this class is built using the {@link Builder}.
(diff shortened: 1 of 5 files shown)


components/common/src/main/java/com/hotels/styx/common/FailureHandlingTaskProcessor.java
```diff
@@ -0,0 +1,144 @@
...
+                onEachFailure.accept(input, t);
+                failures.put(input, t);
+            }
+        });
+
+        if (!failures.isEmpty()) {
+            failuresPostProcessing.accept(failures);
+        }
+
+        return successes;
```
dvlato: do you think it would make sense to change the default behaviour to 'failure' instead of just returning the number of errors?

components/common/src/main/java/com/hotels/styx/common/FailureHandlingTaskProcessor.java
```diff
@@ -0,0 +1,144 @@
...
+        public Builder<T, R> doImmediatelyOnEachFailure(BiConsumer<T, Exception> onEachFailure) {
+            this.onEachFailure = requireNonNull(onEachFailure);
+            return this;
+        }
+
+        /**
+         * An action to take after processing all list items, if there were failures.
+         * This will only execute if the individual failures were did not rethrow their exceptions.
+         *
+         * This action may be used to throw an exception, otherwise the {@link #process(List, Task)} method
```
dvlato: I would say that documenting what happens in 'external' code (the _otherwise_ part) should generally be avoided. You can just maybe rephrase the sentence.

components/common/src/main/java/com/hotels/styx/common/FailureHandlingTaskProcessor.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * A task to execute. Similar to {@link java.util.function.Function}, except it is allowed to throw any type of
+     * {@link Exception}.
+     *
+     * @param <T> input type
+     * @param <R> output type
+     */
+    public interface Task<T, R> {
```
dvlato: If I were to read this the code for the first time, I would probably interpret 'Task' as something to do with concurrency... If you agree it could be misleading, can you think of a better name?

components/proxy/src/main/java/com/hotels/styx/proxy/plugin/PluginSuppliers.java
```diff
@@ -44,11 +46,20 @@
     private final PluginFactoryLoader pluginFactoryLoader;
     private final Environment environment;
 
+    private final FailureHandlingTaskProcessor<Pair<String, SpiExtension>, NamedPlugin> processor =
```
dvlato: The name of the variable could probably be improved...

components/common/src/main/java/com/hotels/styx/common/FailureHandlingTaskProcessor.java
kvosper: Is it really external if it is in the parent class that the builder creates?

PR #374, opened 2019-01-29
added examples

VivianLopes:


plugin-examples/src/main/java/com/hotels/styx/ExampleMessageIntercept.java
```diff
@@ -0,0 +1,78 @@
+package com.hotels.styx;
```
kvosper: This is not a valid Java class. I will explain why further down. To avoid submitting invalid classes in pull requests, please attempt to run the associated unit test(s) first.

plugin-examples/src/main/java/com/hotels/styx/ExampleMessageIntercept.java
```diff
@@ -0,0 +1,78 @@
...
+
+import static com.hotels.styx.api.HttpHeaderNames.CONTENT_TYPE;
+import static com.hotels.styx.api.HttpResponseStatus.OK;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * This shows an example of an inteceptor that responds to a received request
+ */
+
+public class Example implements Plugin {
```
kvosper: Java class names must match the file name. I would not choose either of `Example` (too vague) or `ExampleMessageIntercept` (unclear, not ending with a noun).

Perhaps `EarlyReturnExamplePlugin` would be a good name.

Another thing to note is that there is already a class called `ExamplePlugin`. That was fine when there was only one example plugin, but now that you have added another, it has become too vague. You should come up with a new name and rename it.

plugin-examples/src/main/java/com/hotels/styx/ExampleMessageIntercept.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public Eventual<LiveHttpResponse> intercept(LiveHttpRequest request, Chain chain) {
+        if (request.header("X-Respond").isPresent()) {
+            // request.consume();
```
kvosper: There should not be any commented code here.

Also please add a comment or two explaining what this code is doing - e.g. explain the branching condition and what happens in each branch.

plugin-examples/src/main/java/com/hotels/styx/ExampleMessageIntercept.java
```diff
@@ -0,0 +1,78 @@
...
+ * You can replace live content
+ * If you need to replace a message body based on information in the message headers and don't care what is in
+ * the message body. (E.g you need to add a HTTP message body based on a HTTP error code)
+ *
+ * You can transform a live HTTP message body using a replaceWith Bytestream operator such as in the example.
+ */
+
+//Example:
+
+chain.proceed(request)
```
kvosper: We have dangling Java code here that is outside of any valid part of the file. This will prevent it from compiling. Presumably it is meant to be commented too.

However, it doesn't really make sense to have all these comments/code examples underneath the class file. If we have one example as a proper class with accompanying test, shouldn't the rest of the examples be done in the same way?

plugin-examples/src/test/java/com/hotels/styx/ExampleMessageInterceptTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+
+/**
+ * This tests the behaviours added in the example plugin.
+ */
+
+public class PluginExampleTest {
```
kvosper: Java class names must match the file name. In the case of tests, we take the name of the class under test and add `Test` to the end.

plugin-examples/src/test/java/com/hotels/styx/ExampleMessageInterceptTest.java
```diff
@@ -0,0 +1,34 @@
...
+    @Test
+    public void returnsEarly() {
+        PluginExample plugin = new PluginExample();
```
kvosper: The class `PluginExample` does not exist. Please ensure that code compiles and tests pass before creating a PR. In the case of these examples, it is important to provide code that we have confirmed functional, so as to avoid confusing anyone who tries to learn from it.

plugin-examples/src/test/java/com/hotels/styx/EarlyReturnExamplePluginTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+
+/**
+ * This tests the behaviours added in the example plugin.
+ */
+
+public class EarlyReturnExamplePluginTest {
```
kvosper: We should also include a test that proves the behaviour when the `X-Respond` header is not present.

plugin-examples/src/test/java/com/hotels/styx/EarlyReturnExamplePluginTest.java
```diff
@@ -0,0 +1,34 @@
...
+    @Test
+    public void returnsEarly() {
```
kvosper:
```suggestion
    public void returnsEarlyWhenHeaderIsPresent() {
```

plugin-examples/src/main/java/com/hotels/styx/GeneralExamplePlugin.java
```diff
@@ -32,15 +32,15 @@
  * You can change it in whatever way you like (provided it still implements the Plugin interface),
  * and should rename it to something relevant to your project.
  */
-public class ExamplePlugin implements Plugin {
-    private final ExamplePluginConfig config;
+public class GeneralExamplePlugin implements Plugin {
```
kvosper: Perhaps instead, we should name it `ModifyHeadersExamplePlugin`, to clarify its behaviour (please not we would also need to rename the config, factory and test classes accordingly).

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
```diff
@@ -0,0 +1,64 @@
...
+import com.hotels.styx.api.LiveHttpRequest;
+import com.hotels.styx.api.LiveHttpResponse;
+import com.hotels.styx.api.plugins.spi.Plugin;
+
+import java.util.Map;
+
+import static java.util.Collections.emptyMap;
+import static java.util.Objects.requireNonNull;
+
+
```
kvosper: The comment describing the class should be up here.

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
```diff
@@ -0,0 +1,64 @@
...
+public class ReplaceContentByAggregationExample implements Plugin {
+
+
+    /**
+     * You can replace content by aggregation
```
kvosper:
```suggestion
     * You can replace content after aggregating it in memory.
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
```diff
@@ -0,0 +1,64 @@
...
+     * For example the message content contains a JSON object and you need to modify the object somehow.
```
kvosper:
```suggestion
     * For example, if the message content contains a JSON object and you need to modify the object somehow.
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
```diff
@@ -0,0 +1,64 @@
...
+     * You can aggregate the live HTTP message into a full HTTP message. Transform the content into a full message context
+     * and convert the results back to live HTTP message
+     */
+
+    @Override
+    public Eventual<LiveHttpResponse> intercept(LiveHttpRequest request, Chain chain) {
+
+        LiveHttpRequest newRequest = request.newBuilder()
```
kvosper: This `newRequest` is never used, so it should not be created.

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
```diff
@@ -0,0 +1,64 @@
...
+                .header("myRequestHeader", config.requestHeaderValue())
+                .build();
+
+        return chain.proceed(request)
+                .flatMap(response -> it.aggregate(10000))
+                .map(response -> {
+                    String body = response.bodyAs(UTF_8);
+                    return response.newBuilder()
+                            .body(modify(body), UTF_8)
```
kvosper: This line references a method called `modify`, but no such method exists.

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
```diff
@@ -0,0 +1,64 @@
...
+                            .build();
+                })
+                .map(HttpResponse::stream);
+    }
+}
+
+/**
```
kvosper: This comment should be at the top of the file (combined with other comment).

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
```diff
@@ -0,0 +1,64 @@
...
+ * This can be used when you need to do something with full content and when you need to replace content after looking
```
kvosper:
```suggestion
 * This can be used when you need the entire content in order to operate on it.
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
```diff
@@ -0,0 +1,64 @@
...
+ * into it
```
kvosper:
```suggestion
 * 
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
```diff
@@ -0,0 +1,64 @@
...
+ * You need to do something with full content. This uses more heap as the full response it transiently stored in heap
```
kvosper:
```suggestion
 * Please note that this uses more heap space as the full response is transiently stored in the heap.
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregation.java
kvosper: Also, there is no test for this class.

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExample.java
```diff
@@ -0,0 +1,58 @@
...
+import com.hotels.styx.api.LiveHttpRequest;
+import com.hotels.styx.api.LiveHttpResponse;
+import com.hotels.styx.api.plugins.spi.Plugin;
+
+import java.util.Map;
+
+import static java.util.Collections.emptyMap;
+import static java.util.Objects.requireNonNull;
+
+
```
kvosper: The comment describing the class should be up here.
Also, there is no test for this class.

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExample.java
```diff
@@ -0,0 +1,58 @@
...
+public class ReplaceLiveContentExample implements Plugin {
+
+    /**
+     * You can replace live content
```
kvosper:
```suggestion
     * You can replace content without aggregating it first, if the replacement does not depend on the original contents.
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExample.java
```diff
@@ -0,0 +1,58 @@
...
+     * If you need to replace a message body based on information in the message headers and don't care what is in
```
kvosper:
```suggestion
     * For example, if you need to replace a message body based on information in the message headers, regardless of 
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExample.java
```diff
@@ -0,0 +1,58 @@
...
+     * the message body. (E.g you need to add a HTTP message body based on a HTTP error code)
```
kvosper:
```suggestion
     * the original message body. (E.g you need to add a HTTP message body based on a HTTP error code)
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExample.java
```diff
@@ -0,0 +1,58 @@
...
+     * <p>
+     * You can transform a live HTTP message body using a replaceWith Bytestream operator such as in the example.
```
kvosper:
```suggestion
     * You can transform a live HTTP message body using the `replaceWith` Bytestream operator as shown in the example below.
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExample.java
```diff
@@ -0,0 +1,58 @@
...
+                .build();
+
+        return chain.proceed(request)
+                .map(response -> response.newBuilder()
+                        .body(body -> body.replaceWith(ByteStream.from("replacement", UTF_8)))
+                        .build());
+    }
+}
+
+/**
```
kvosper: This comment should be at the top of the file, combined with the other comment.

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExample.java
```diff
@@ -0,0 +1,58 @@
...
+This can be used to replace a message body without having to look into it, this will also save heap as the
```
kvosper:
```suggestion
* This can be used to replace a message body without having to look into it, which will also save heap space as the
```

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExample.java
```diff
@@ -0,0 +1,58 @@
...
+ live upstream response body is never stored in heap in full
```
kvosper:
```suggestion
*  live upstream response body is never stored in the heap in full.
```

plugin-examples/src/main/java/com/hotels/styx/EarlyReturnExamplePlugin
```diff
@@ -0,0 +1,50 @@
+/*
```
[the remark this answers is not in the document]
dvlato: The filename is still incorrect

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregationExample.java
```diff
@@ -0,0 +1,59 @@
...
+ * You can replace content after aggregating it in memory.
+ * For example, if the message content contains a JSON object and you need to modify the object somehow.
+ * You can aggregate the live HTTP message into a full HTTP message. Transform the content into a full message context
+ * and convert the results back to live HTTP message
+ *
+ *This can be used when you need the entire content in order to operate on it.
+ *Please note that this uses more heap space as the full response is transiently stored in the heap.
+ */
+
+public class ReplaceContentByAggregationExample implements Plugin {
```
dvlato: I cannot find any 'modifier' implementation  so I don't think this is 'usable' code.   Also the ModifyHeadersExamplePluginConfig field/attribute is not used.

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExample.java
```diff
@@ -0,0 +1,55 @@
...
+import com.hotels.styx.api.LiveHttpResponse;
+import com.hotels.styx.api.plugins.spi.Plugin;
+
+import java.util.Map;
+
+import static java.util.Collections.emptyMap;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * You can replace without aggregating it first, if the replacement does not depend on the original contents.
```
dvlato: You can replace... the body content?

plugin-examples/src/main/java/com/hotels/styx/ReplaceContentByAggregationExample.java
```diff
@@ -0,0 +1,59 @@
...
+    ModifyHeadersExamplePluginConfig config;
+
+    private Modifier modifier; // modifier is an interface representing a class that performs some modification
+
+    @Override
+    public Eventual<LiveHttpResponse> intercept(LiveHttpRequest request, Chain chain) {
+
+        chain.proceed(request)
+                .flatMap(response -> it.aggregate(10000))
```
kvosper: This code has syntax errors and does not compile.

plugin-examples/pom.xml
```diff
@@ -1,205 +1,82 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <parent>
```
dvlato: I believe we removed the 'parent' here so the pom file can be copied to real plugins that will not declare themselves as 'children' of  styx-parent.   Do we definitely need that?

plugin-examples/src/main/java/com/hotels/styx/EarlyReturnExamplePlugin.java
```diff
@@ -0,0 +1,50 @@
...
+ * the HTTP response
+ * The eventual object is returned immediately, even if the response has not yet arrived.
+ */
+
+public class EarlyReturnExamplePlugin implements Plugin {
+
+    @Override
+    public Eventual<LiveHttpResponse> intercept(LiveHttpRequest request, Chain chain) {
+        if (request.header("X-Respond").isPresent()) {
+            return Eventual.of(HttpResponse.response(OK)
```
dvlato: Should we consume the request body here?

plugin-examples/src/main/java/com/hotels/styx/ReplaceLiveContentExamplePlugin.java
```diff
@@ -0,0 +1,53 @@
...
+import com.hotels.styx.api.Eventual;
+import com.hotels.styx.api.LiveHttpRequest;
+import com.hotels.styx.api.LiveHttpResponse;
+import com.hotels.styx.api.plugins.spi.Plugin;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * You can replace without aggregating it first, if the replacement does not depend on the original contents.
```
dvlato: I find the beginning a bit unclear.  A suggestion you can quickly discard : (The) body content can be replaced without ... Something along those lines.

plugin-examples/pom.xml
```diff
@@ -1,205 +1,82 @@
...
     </dependency>
     <dependency>
       <groupId>io.reactivex</groupId>
       <artifactId>rxjava</artifactId>
-      <version>${rxjava.version}</version>
-      <scope>provided</scope>
     </dependency>
 
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
```
dvlato: As mentioned, all these changes in the dependencies mean that the pom file cannot be just copied to a new plugin. At some point we should make sure our BOM file does not expose all the dependencies of Styx and then have plugins import it...

kvosper:
See #432

PR #378, opened 2019-02-22
Refactor plugin loading

kvosper:
The primary focus of this PR is to decouple the Admin server from the plugin-loading, so that will be able to visit the admin server before the plugins have all been loaded.
That work is not totally finished by this PR (as it is already quite large and we don't want to add even more changes).

I have also taken the opportunity to tidy-up parts of the plugin-loading code that were messy.

It is a prerequisite to allowing Styx to have proper liveness/readiness checks, which needs the admin server to be able to start without waiting for the plugins.


components/proxy/src/main/java/com/hotels/styx/startup/extensions/FailureHandling.java
```diff
@@ -0,0 +1,118 @@
...
+final class FailureHandling {
+    private static final Logger LOGGER = getLogger(FailureHandling.class);
+
+    static final FailureHandlingStrategy<Pair<String, SpiExtension>, ConfiguredPluginFactory> PLUGIN_FACTORY_LOADING_FAILURE_HANDLING_STRATEGY =
+            new FailureHandlingStrategy.Builder<Pair<String, SpiExtension>, ConfiguredPluginFactory>()
+
+                    .doImmediatelyOnEachFailure((plugin, err) -> logIndividualFailure(new FailedPlugin(plugin), err))
+
+                    .doOnFailuresAfterAllProcessing(failures -> {
+                        throw new PluginStartupException(afterFailuresErrorMessage(transformKeys(failures, FailedPlugin::new)));
```
dvlato: it seems transformKeys()'s second argument is always  "FailedPlugin::new" . Then why do we need it be an argument at all?

components/proxy/src/main/java/com/hotels/styx/startup/extensions/FailureHandling.java
```diff
@@ -0,0 +1,118 @@
...
+
+        return format("%s plugin%s could not be loaded: failedPlugins=%s; failureCauses=%s",
+                failures.size(),
+                failures.size() == 1 ? "" : "s", // only pluralise plurals
+                failedPlugins,
+                causes
+        );
+    }
+
+    private static class FailedPlugin {
```
dvlato: Nitpicking again, but I wonder if we really need a class for this purpose.

components/proxy/src/main/java/com/hotels/styx/startup/extensions/PluginLoadingForStartup.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Load plugins.
+     *
+     * @param environment environment
+     * @param factories   if set, overrides config, otherwise plugins will be loaded from config
+     * @return plugins
+     */
+    public static List<NamedPlugin> loadPlugins(Environment environment, List<ConfiguredPluginFactory> factories) {
+        if (factories == null) {
```
dvlato: Same comment about null arguments.

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
```diff
@@ -174,7 +184,7 @@ public StyxServerComponents build() {
     public interface LoggingSetUp {
         LoggingSetUp DO_NOT_MODIFY = environment -> {
         };
-        LoggingSetUp FROM_CONFIG = environment -> setUpLogging(environment.configuration().logConfigLocation());
+        LoggingSetUp FROM_CONFIG = environment -> setUpLogging(environment.styxConfig().logConfigLocation());
```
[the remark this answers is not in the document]
kvosper: Both of these methods already existed. This is a change to which method is called. I will try to unify it though, as we probably don't need both.

components/proxy/src/main/java/com/hotels/styx/startup/extensions/ConfiguredPluginFactory.java
```diff
@@ -0,0 +1,60 @@
...
+    private final Function<Class<?>, Object> configProvider;
+
+    public ConfiguredPluginFactory(String name, PluginFactory pluginFactory, Function<Class<?>, Object> configProvider) {
+        this.name = requireNonNull(name);
+        this.pluginFactory = requireNonNull(pluginFactory);
+        this.configProvider = configProvider == null ? any -> null : configProvider;
+    }
+
+    public ConfiguredPluginFactory(String name, PluginFactory pluginFactory) {
+        this(name, pluginFactory, null);
```
dvlato: Why don't you add the "any -> null" here so you can enforce that configProvider is not null in above's constructor?  I don't know if it requires changing anything other than a few Tests...

PR #381, opened 2019-03-07
Improve Styx Dockerfile

mikkokar:
Generify Dockerfile so that it can be used for building both development images and official Styx release images.

**This change also removes the release version number from Styx assembly base directory.**
That is the top-level directory in a Styx release zip file. The version number in the base directory is pointless because it is in the release zip file anyway, and it unnecessarily complicates the Dockerfile and possibly other release scripts.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -8,7 +8,7 @@ To build Styx Docker image, run `make docker-image` from the Styx project root:
 
 To run the resulting Styx Docker image:
 
-    $ docker container run -d --name myStyx -p 8000:8080 -p 9000:9000 -p 8443:8443styxcore:latest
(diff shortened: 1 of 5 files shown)


docker/README.md
```diff
@@ -31,7 +31,8 @@ Ensure the the Styx configuration reads the origins from `/styx/config/` directo
 
 Finally, start the Styx Docker image using `/styx/config/styxconf.yml` as a configuration file:        
 
-    docker container run -d --name myStyx \
-        -p 8080:8080 -p 9000:9000 -p 8443:8443
+    docker container run -d  \
```
dvlato: maybe keep the name as in the previous examples and suggest --rm ?

mikkokar: Changed the code as you suggested. Please have a look again.

PR #383, opened 2019-03-07
Liveness/readiness checks

kvosper:
Adds endpoints for checking:
1. Styx is live (i.e. running)
2. Styx is ready (to serve traffic)


components/proxy/src/main/java/com/hotels/styx/startup/ProxyServerSetUp.java
```diff
@@ -39,15 +43,34 @@ public ProxyServerSetUp(PipelineFactory pipelineFactory) {
         this.pipelineFactory = requireNonNull(pipelineFactory);
     }
 
-    public HttpServer createProxyServer(StyxServerComponents config) {
-        HttpHandler pipeline = pipelineFactory.create(config);
+    public HttpServer createProxyServer(StyxServerComponents components) {
+        // TODO see https://github.com/HotelsDotCom/styx/issues/382
+
+        while (!arePluginsLoaded(components)) {
```
dvlato: Shouldn't we put a (configurable) max time limit for the loading of plugins/startup of the app so we won't block  here forever? This could be some limit in styxServer.startAsync().awaitRunning(xxx) which is unlimited now.

kvosper: That's a good idea, though perhaps it should be a separate issue & feature-branch?
It is already possible for plugins to block Styx start-up in the `master`.

dvlato: I would include it in this PR:
In a scenario such as Kubernetes where the 'ping' endpoint is used for liveness checks, before this PR Styxpres would fail liveness checks if it got stuck and would be restarted (and that might perhaps solve the issue). However, after this PR Styxpres would be considered alive and would not be restarted.

bestokes:
@mikkokar please can you review?

components/proxy/src/main/java/com/hotels/styx/admin/handlers/livenessreadiness/Liveness.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Provides an endpoint for checking that the Styx JVM is running.
+ */
+public final class Liveness {
+    public static final HttpHandler LIVENESS_HANDLER = new HttpMethodHandler.Builder()
+            .add(GET, response(OK)
+                    .header(CONTENT_TYPE, JSON_UTF_8)
+                    .body(jsonBody(), UTF_8)
+                    .build())
+            .build();
```
mikkokar: Should we set cache control headers to prevent the liveness response from being cached?

components/proxy/src/main/java/com/hotels/styx/admin/handlers/livenessreadiness/Readiness.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final HttpResponse LOADED_RESPONSE = response(OK)
+            .header(CONTENT_TYPE, JSON_UTF_8)
+            .body(jsonBody(true), UTF_8)
+            .build();
+
+    private static final HttpResponse NOT_LOADED_RESPONSE = response(SERVICE_UNAVAILABLE)
+            .header(CONTENT_TYPE, JSON_UTF_8)
+            .body(jsonBody(false), UTF_8)
+            .build();
+
```
mikkokar: Here again I think we should disable caching?

components/proxy/src/main/java/com/hotels/styx/startup/BaseService.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.concurrent.ExecutorService;
+
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.Executors.newSingleThreadExecutor;
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * A base to build services on. Adds concurrency, logging, error handling.
+ */
+public abstract class BaseService extends AbstractService {
```
mikkokar: We already have a `StyxService` and `AbstractStyxService` abstraction. Please look into them. We should use the existing classes instead. Unless there are compelling reasons not to do otherwise?

components/proxy/src/main/java/com/hotels/styx/startup/StyxPipelineFactory.java
```diff
@@ -66,18 +67,21 @@ public HttpHandler create(StyxServerComponents config) {
 
         boolean requestTracking = config.environment().configuration().get("requestTracking", Boolean.class).orElse(false);
 
+        List<NamedPlugin> plugins = config.environment().configStore()
+                .valuesStartingWith("plugins", NamedPlugin.class);
```
mikkokar: Consider config-store as global data,  and therefore avoid unnecessary lookups when possible. 
In this case I would pass `plugins` list as a method argument. It is already available at the call site.

components/proxy/src/test/java/com/hotels/styx/StyxServerTest.java
```diff
@@ -110,15 +111,14 @@ public void disablesResourceLeakDetectionByDefault() {
...
+    public void stopsTheServerWhenPluginFailsToStart() throws TimeoutException {
         StyxServer styxServer = null;
         try {
             styxServer = styxServerWithPlugins(ImmutableMap.of(
                     "foo", new NonStarterPlugin("foo"),
                     "mockplugin3", mock(Plugin.class)));
 
-            Service service = styxServer.startAsync();
-            eventually(() -> assertThat(service.state(), is(FAILED)));
+            tryToStartExpectingFailure(styxServer);
```
mikkokar: This change seems unnecessary? 

Ok, this is a matter of taste, but I like the old style better where the test body explicitly specifies the desired behaviour. Without hiding it into `tryToStartExpectingFailure` method. Especially given such a marginal reduction in lines of code.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/livenessreadiness/Liveness.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    private static String jsonBody() {
+        try {
+            return new ObjectMapper().writeValueAsString(
+                    ImmutableMap.of("live", "true"));
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
mikkokar: This always returns true? As such this liveness test is unnecessary because the requestor can probe any of the other URL endpoints regardless of the response.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/livenessreadiness/Readiness.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public static HttpHandler readinessHandler(ConfigStore configStore) {
+        return new HttpMethodHandler.Builder()
+                .add(GET, request -> generateResponse(configStore))
+                .build();
+    }
+
+    private static HttpResponse generateResponse(ConfigStore configStore) {
+        boolean loaded = configStore.get("loading.styx", Object.class).isPresent();
```
mikkokar: The attribute name `loading.styx` suggests that the server is still actively attempting to start up.

Why not to call it `server.state` with values of `loading` and `ready`? Also this is bit problematic because Styx can open multiple server ports: http and https. Presumably it will be ready when the both are open.

components/proxy/src/main/java/com/hotels/styx/startup/PluginStartupService.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    protected void doStart0() {
+        List<NamedPlugin> plugins = pluginFactories == null
+                ? loadPlugins(environment)
+                : loadPlugins(environment, pluginFactories);
+
+        plugins.forEach(plugin -> environment.configStore().set("plugins." + plugin.name(), plugin));
+        environment.configStore().set("loading.plugins", "finished");
+    }
+
```
mikkokar: Perhaps this should also initialise the plugins? Looks like the plugins init phase is still called from `ProxyServerSetup`. Seems bit contrived to load the plugins here, stall the proxy server setup until loading is complete, to finally initialise plugins there.

components/proxy/src/main/java/com/hotels/styx/startup/PluginStartupService.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    protected void doStart0() {
+        List<NamedPlugin> plugins = pluginFactories == null
+                ? loadPlugins(environment)
+                : loadPlugins(environment, pluginFactories);
+
+        plugins.forEach(plugin -> environment.configStore().set("plugins." + plugin.name(), plugin));
+        environment.configStore().set("loading.plugins", "finished");
```
mikkokar: We should collect all config store keys in static final constants. 

Please also think about the attribute names: Please establish a sensible naming convention and ensure they fall in a sensible hierarchy.

components/proxy/src/test/java/com/hotels/styx/StyxServerTest.java
kvosper: The test was changed because `StyxServer.startAsync` is actually asynchronous now. In the past it was actually blocking.

The new `tryToStartExpectingFailure` looks like:
```
        try {
            server.startAsync().awaitRunning(3, SECONDS);
            fail("Expected failure but succeeded");
        } catch (IllegalStateException e) {
            // Desired behaviour, so no action taken
        }
```
So reducing that to only 1 line is quite substantial.

components/proxy/src/main/java/com/hotels/styx/startup/ProxyServerSetUp.java
```diff
@@ -39,32 +48,51 @@ public ProxyServerSetUp(PipelineFactory pipelineFactory) {
...
+                .httpHandler(pipeline)
+                .build();
+
+        proxyServer.addListener(new PluginsNotifierOfProxyState(plugins), sameThreadExecutor());
+        return proxyServer;
+    }
+
+    private static void sleep(long duration, TimeUnit unit) {
+        try {
+            Thread.sleep(unit.toMillis(duration));
```
dvlato: As per Java best practices, at least set the interrupted status of the Thread when an InterruptedException cannot be re-thrown.

I would use function composition to deal with these exceptions (similar to your throwsException() function) instead of declaring a sleep function.

components/api/src/main/java/com/hotels/styx/api/extension/service/spi/StyxService.java
```diff
@@ -73,4 +76,35 @@
...
+            public CompletableFuture<Void> stop() {
+                return parent.stop();
+            }
+
+            @Override
+            public Map<String, HttpHandler> adminInterfaceHandlers() {
+                return parent.adminInterfaceHandlers();
+            }
+        };
+    }
```
mikkokar: This change seems stem from the synchronous server start in `ServerService` class. I think this change can be avoided.

If I understand correctly, we have two or three requirements:
1. Set the server state to "complete" when it starts successfully
2. Set the server state to "failed" when it fails to start.

and optionally 

3. Set the server state to "shut" etc when it is stopped 

You can achieve these by deriving `ServerService` directly from `AbstractStyxService` and overriding its `startService` method to start the HTTP server. You can then register a listener to get notified of the server state changes. Failing that, you could perhaps start the HTTP server in `CompletableFuture.supplyAsync` block and use `thenAccept` and `exceptionally` to capture the outcome.

This should knock off lots of code in `StyxService` and `BaseService` classes.

components/proxy/src/main/java/com/hotels/styx/startup/extensions/PluginStatusNotifications.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Status for a plugin to be in during start-up.
+     */
+    public enum PluginStatus {
+        LOADING_CLASSES("incomplete:loading-classes"),
+        LOADED_CLASSES("incomplete:loaded-classes"),
+        CONSTRUCTING("incomplete:constructing"),
+        CONSTRUCTED("incomplete:constructed"),
+        LIFECYCLE_STARTING("incomplete:lifecycle-starting"),
```
mikkokar: "lifecycle starting". This should really say "plugin starting".

components/proxy/src/main/java/com/hotels/styx/startup/extensions/PluginStatusNotifications.java
```diff
@@ -0,0 +1,93 @@
...
+        COMPLETE("complete"),
+        FAILED_WHILE_LOADING_CLASSES("failed:loading-classes"),
+        FAILED_WHILE_CONSTRUCTING("failed:constructing"),
+        FAILED_WHILE_LIFECYCLE_STARTING("failed:lifecycle-starting");
```
mikkokar: And this should also say "plugin starting".

codequality/checkstyle_rules.xml
```diff
@@ -354,39 +354,39 @@
...
     <property name="message" value="use EncodingSupport for encoding"/>
-    <property name="id" value="BannedEncodingLibraryReference_Error"/>
+    <!--<property name="id" value="BannedEncodingLibraryReference_Error"/>-->
   </module>
   <module name="RegexpSingleline">
     <property name="severity" value="error"/>
     <property name="format" value="null !=|null =="/>
     <property name="message" value="Check for null in reverse order"/>
-    <property name="id" value="ReverseOrderNullCheck_Error"/>
+    <!--<property name="id" value="ReverseOrderNullCheck_Error"/>-->
```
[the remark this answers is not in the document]
kvosper: Yes, this was committed by mistake. Next commit removes them.

components/api/src/main/java/com/hotels/styx/api/extension/service/spi/StyxService.java
```diff
@@ -73,4 +75,50 @@
...
+                return parent.adminInterfaceHandlers();
+            }
+
+            @Override
+            public String toString() {
+                return parent.toString();
+            }
+
+        };
+    }
```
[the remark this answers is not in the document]
kvosper: I'll look into this

components/proxy/src/main/java/com/hotels/styx/startup/extensions/PluginStatusNotifications.java
kvosper: Or just starting, since all the statuses are plugin-related.

components/common/src/main/java/com/hotels/styx/common/SequenceProcessor.java
```diff
@@ -0,0 +1,179 @@
...
+import static java.util.stream.Collectors.toMap;
+
+/**
+ * Provides the ability to easily apply side effects and failure handling to the processing of a sequence of inputs
+ * without cluttering code with nested loops, try-catch and if-else statements.
+ *
+ * @param <T> input type
+ * @param <E> output type
+ */
+public class SequenceProcessor<T, E> {
```
mikkokar: There are better ways to tackle our problem. Instead of implementing a bespoke sequence processor, you'd want go for monadic error handing, like `Try` or `Either` Scala types.

Consider implementing a simple `Try` class in Java. It won't be perfect but would have many advantages over a bespoke processor class: 
* It would work as a combinator with both java streams, reactor core flux, rx java and so on.
* More flexible as it is not limited to the operations supported by the sequence processor.
* Less code to maintain

For example:

    @Test
    public void testName() {
        List<Integer> inputs = asList(1, 2, 3, 4, 5, 6, 7);

        List<Try<String>> outcomes = inputs.stream()
                .map(number -> Try.tryBlock(() -> {
                    if (number % 2 == 0) {
                        throw new IllegalStateException();
                    }

                    return String.valueOf(number);
                }))
                .collect(Collectors.toList());

        List<String> successes = outcomes.stream()
                .filter(Try::isSuccess)
                .map(Try::result)
                .collect(Collectors.toList());

        List<Exception> errors = outcomes.stream()
                .filter(it -> !it.isSuccess())
                .map(Try::failure)
                .collect(Collectors.toList());
    }

    static class Try<T> {
        private final Exception exception;
        private final T result;

        Try(T result) {
            this.result = result;
            this.exception = null;
        }

        Try(Exception exception) {
            this.result = null;
            this.exception = exception;
        }

        public static <U, T> Try<T> tryBlock(Supplier<T> lambda) {
            try {
                return new Try<>(lambda.get());
            } catch (Exception e) {
                return new Try<>(e);
            }
        }

        public boolean isSuccess() {
            return this.exception == null;
        }

        public T result() {
            return this.result;
        }

        public Exception failure() {
            return this.exception;
        }
    }


With this the `SequenceProcessor` becomes obsolete.

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -115,6 +116,8 @@ public HttpServer build() {
         httpRouter.add("/admin/configuration/logging", new LoggingConfigurationHandler(styxConfig.startupConfig().logConfigLocation()));
         httpRouter.add("/admin/configuration/startup", new StartupConfigHandler(styxConfig.startupConfig()));
 
+        httpRouter.add("/admin/startup", new StartupStatusHandler(environment.configStore()));
```
mikkokar: Please provide an admin page link.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -131,34 +143,47 @@ private static boolean skipServerConfigValidation() {
...
         Map<String, StyxService> servicesFromConfig = components.services();
 
-        ProxyServerSetUp proxyServerSetUp = new ProxyServerSetUp(new StyxPipelineFactory());
+        PluginStartupService pluginStartupService = new PluginStartupService(components);
+
+        this.configStore = components.environment().configStore();
 
-        components.plugins().forEach(plugin -> components.environment().configStore().set("plugins." + plugin.name(), plugin));
+        StyxService adminServerService = new ServerService("adminServer", () ->
+                this.adminServer.store(createAdminServer(components)));
```
mikkokar: Why it is necessary to an `HttpServer` instance and wrap it into a `ServerReference` object?

Why not to create an `HttpServer` instance and pass it as is to `ServerService`? Like so:

    new ServerService(createAdminServer(components));

kvosper: The whole point of the `ServerService` is to run the creation/start-up process of the server. If you do it before construction, then the service does nothing, and we would not be able to start the servers asynchronously.
The `ServerReference` is because the `StyxServer` needs to receive the server after it has been created asynchronously.

mikkokar: `createAdminServer()` just creates a server object. It doesn't start it. The service is still be able to start it by invoking `startAsync`.

Okay. I see. The `createProxyServer` blocks until the plugins are loaded. This is somewhat confusing:

`createProxyServer()` returns an `HttpServer` instance which has `startAsync` method. Typical usage is:

    HttpServer server = createProxyServer();
    server.startAsync().awaitRunning();

Where does it block? On both lines. Surprisingly. The first line blocks at loading the plugins, the 2nd line blocks at waiting for the server to start.

IMO it would be more sensible to wait for plugins in the service startup method, rather than in HttpServer creation.

kvosper:
We have decided not to go ahead with this PR, however it did spawn two smaller pieces of work:

#389 
#390

Issue #382, opened 2019-03-07
Race condition in ConfigStore


kvosper:
## The problem
There seems to be a race condition in the method `watch(String, Class<T>)` in the class `com.hotels.styx.configstore.ConfigStore`.

## Detailed description
While running unit tests, some of the tests would not give consistent results - sometimes the test would pass, sometimes fail. Further investigation showed that an event was never being delivered.

The code was:
```
            components.environment().configStore().watch("loading.plugins", Object.class)
                    .toBlocking()
                    .first();
```
This code blocked indefinitely even though the value `loading.plugins` had been set.

Note: it was called from `ProxyServerSetUp.createProxyServer` and has been replaced by a workaround that uses polling.

## Acceptance criteria
(/) The race condition should no longer exist

mikkokar:
Config store is going to go away.

PR #386, opened 2019-03-12
Expose ConfigStore in API so it can be used by extensions and plugins

kvosper:
Currently, extensions cannot do very much - they can only read config and publish metrics.
Giving access to the `ConfigStore` (via an interface) would allow them to provide data and functionality to other parts of the system.

For example, a service extension could provide access to some data or an external system required by multiple plugins. This would prevent the plugins from duplicating code and data, or from depending upon each other.

Notes:
* We may not want to call the interface `ConfigStore` - if you have a better name, please suggest it.
* The `ConfigStore` interface currently only exposes `get` as a means of access due to issue #382, we could add the other methods after that issue is resolved.


mikkokar:
The `ConfigStore` is still very much experimental and therefore I'd hesitate to expose it in the APIs at this early stage.

We should discuss some practical use cases to determine if it is absolutely necessary.

PR #390, opened 2019-03-27
Service cleanup

kvosper:
This was branched from #389 so please review/merge that one first, or you will see the changes from both branches at once.

This PR fixes two types of problems with our `NettyServer` and `StyxServer` classes, with respect to behaving as services:

1. Previously, the `NettyServer` instances (proxy and admin) are doing some of the start-up before the service is actually `STARTING` (i.e. construction time or before). We move it to the correct location.
2. Some of the start/stop methods are blocking on child services when they are supposed to provide async functionality.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
@@ -94,9 +94,16 @@ public class AdminServerBuilder {
     public HttpServer build() {
         LOG.info("event bus that will be used is {}", environment.eventBus());
         StyxConfig styxConfig = environment.configuration();
-
-        Optional<Duration> metricsCacheExpiration = styxConfig.adminServerConfig().metricsCacheExpiration();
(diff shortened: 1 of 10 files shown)


components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -94,9 +94,16 @@ public AdminServerBuilder backendServicesRegistry(Registry<BackendService> backe
...
         StyxConfig styxConfig = environment.configuration();
+        AdminServerConfig adminServerConfig = styxConfig.adminServerConfig();
+
+        return new NettyServerBuilderSpec("Admin", environment.serverEnvironment(), new WebServerConnectorFactory())
+                .toNettyServerBuilder(adminServerConfig)
+                .handlerFactory(() -> createHttpRouter(styxConfig))
+                .build();
+    }
 
+    private StandardHttpRouter createHttpRouter(StyxConfig styxConfig) {
```
dvlato: can we use a more meaningful name?

Issue #391, opened 2019-03-27
URL parsing bug with http://example.com//foo/bar


taer:
When a URL comes into Styx like this

https://test.homeaway.co.uk//www.homeaway.co.uk/p6873759, it gets messed up inside of Styx. The request path that pops out in the request.path()

path=/p6873759

because styx uses URI.create to parse the path from the HTTP first line path, and a //foo/bar as URI encodes foo as the host.

 

It's here in Styx's Url class
```
        public static Builder url(String value) {
            URI uri = URI.create(value);
            return new Builder()
                    .scheme(uri.getScheme())
                    .authority(uri.getAuthority())
                    .path(uri.getRawPath())
                    .rawQuery(uri.getRawQuery())
                    .fragment(uri.getFragment()); 
```
 value here is just the contents from the first http line.
 

But, it also seems the downstream app is given the original request still.  So origin will see path of `//www.homeaway.co.uk/p6873759`, while styx logged and treated as `/p6873759`

mikkokar:
Browsers would have parsed the URL and constructed an HTTP request before it is sent to styx. This is how Styx sees is:

    GET //www.homeaway.co.uk/p6873759 HTTP/1.1
    Host: localhost:8080
    User-Agent: curl/7.63.0
    Accept: */*

Because `//www.homeaway.co.uk/p6873759 HTTP/1.1` starts with `//` the `URI.create` treats it as authority part, and the rest as a path. 

It is correct behaviour for Styx to proxy the request as `//www.homeaway.co.uk/p6873759`, but it would be incorrect to expose `/p6873759` bit alone as a path component. Also it would be unclear how it gets reassembled back to the correct form when proxied.

I will investigate.

mikkokar:
Styx proxies the correct URL to origins because `HttpRequestOperation.toNettyRequest` uses `url().toString()` in sending the request. This presumably puts together both authority and path components, thus re-assembling the original URL.

Okay. We now need to figure out how to extract the correct path component for use in Styx URL.

mikkokar:
Hi @taer. I have a fix in here: https://github.com/mikkokar/styx/tree/issue-391-url-parsing. Could you give it a try please?

PR #393, opened 2019-03-29
Fix URL parsing issue (#391)

mikkokar:
Fixes issue #391.

The sample URI from #391 is incorrectly exposed to the interceptor chain: 

* Sample URI: `https://test.homeaway.co.uk//www.homeaway.co.uk/p6873759`
* The [origin-form](https://tools.ietf.org/html/rfc7230#section-5.3.1) of the request is: `//www.homeaway.co.uk/p6873759`
* The path component exposed to the plugins is: `/p6873759`
* The request is proxied correctly.

Root cause: Styx calls Java `URI.create` which fails to parse the URI in its origin form. The problem disappears when we temporarily convert the URI to its [absolute-form](https://tools.ietf.org/html/rfc7230#section-5.3.2)  before calling `URI.create`.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/Url.java
+++ b/components/api/src/main/java/com/hotels/styx/api/Url.java
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2018 Expedia Inc.
+  Copyright (C) 2013-2019 Expedia Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -66,7 +66,7 @@ public final class Url implements Comparable<Url> {
      * @return scheme
      */
     public String scheme() {
-        return this.scheme;
+        return this.scheme != null ? this.scheme : "";
     }
 
     /**
@@ -121,7 +121,7 @@ public final class Url implements Comparable<Url> {
      * @return true if the URL is absolute.
      */
     public boolean isAbsolute() {
-        return path.startsWith("/");
+        return Objects.nonNull(scheme) && !scheme.isEmpty();
     }
 
     /**
@@ -131,7 +131,7 @@ public final class Url implements Comparable<Url> {
      * @see {@link #isAbsolute()}
      */
     public boolean isRelative() {
-        return !isAbsolute();
+        return scheme == null || scheme.isEmpty();
     }
 
     /**
@@ -455,7 +455,7 @@ public final class Url implements Comparable<Url> {
                     .fragment(uri.getFragment());
         }
 
-        Builder rawQuery(String rawQuery) {
+        public Builder rawQuery(String rawQuery) {
             this.queryBuilder = rawQuery == null ? null : new UrlQuery.Builder(rawQuery);
             return this;
         }
--- a/components/server/src/test/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoderTest.java
+++ b/components/server/src/test/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoderTest.java
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2019 Expedia Inc.
+  Copyright (C) 2013-2018 Expedia Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -47,7 +47,6 @@ import rx.observers.TestSubscriber;
 
 import java.net.URI;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import static com.google.common.base.Charsets.US_ASCII;
 import static com.google.common.base.Charsets.UTF_8;
@@ -70,7 +69,9 @@ import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
 import static io.netty.handler.codec.http.LastHttpContent.EMPTY_LAST_CONTENT;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.core.Is.is;
-import static org.testng.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 import static rx.RxReactiveStreams.toObservable;
 
 public class NettyToStyxRequestDecoderTest {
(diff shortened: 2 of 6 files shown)


mikkokar:
FYI @taer . Please give this a go if you can.

components/api/src/main/java/com/hotels/styx/api/Url.java
```diff
@@ -121,7 +121,7 @@ public boolean isFullyQualified() {
      * @return true if the URL is absolute.
      */
     public boolean isAbsolute() {
-        return path.startsWith("/");
```
dvlato: What is this method used for? We have changed from _it's an absolute path_ to i_t's an absolute URI_ .  What's the real purpose?

mikkokar: I can only guess. I believe the original intention was to determine if the request is in origin-form or absolute-form. My gut feeling is that we don't really need this method in the URL class.

As to why this change? It is because every path component always starts with `/` therefore it always returned true.

components/server/src/test/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoderTest.java
```diff
@@ -222,17 +221,23 @@ public void throwsBadRequestExceptionWhenRequestMessageContainsMoreThanOneHostHe
 
     @Test
     public void callsTheEscaperForUnwiseChars() throws Exception {
-        UnwiseCharsEncoder encoder = mock(UnwiseCharsEncoder.class);
+        AtomicBoolean encoderCalled = new AtomicBoolean();
+        UnwiseCharsEncoder encoder = x -> {
```
kvosper: Why replace mocking with this approach?

components/server/src/test/java/com/hotels/styx/server/netty/codec/UrlDecoderTest.java
```diff
@@ -0,0 +1,117 @@
...
+        assertThat(url.encodedUri(), is("//www.abc.com//abc123Z"));
+        assertThat(url.isAbsolute(), is(false));
+        assertThat(url.isRelative(), is(true));
+        assertThat(url.host(), is(Optional.empty()));
+        assertThat(url.query(), is(Optional.empty()));
+        assertThat(url.scheme(), is(""));
+    }
+
+    @Test
+    public void decodesOriginFormWithLowercaseHostHeader() {
```
kvosper: This test is identical to the one below and the first test in the file.

components/api/src/main/java/com/hotels/styx/api/Url.java
dvlato: I would at least mark it as deprecated then. I think from previous discussions our goal could be to completely remove Url from the API.

components/server/src/test/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoderTest.java
dvlato: I meant to comment on the same to be honest... I think keeping mock.verify() would make our code a bit more uniform.

components/server/src/test/java/com/hotels/styx/server/netty/codec/UrlDecoderTest.java
```diff
@@ -0,0 +1,117 @@
...
+        assertThat(url.isRelative(), is(true));
+        assertThat(url.host(), is(Optional.empty()));
+        assertThat(url.query(), is(Optional.empty()));
+        assertThat(url.scheme(), is(""));
+    }
+
+    @Test
+    public void decodesOriginFormWithUppercaseHostHeader() {
+        DefaultFullHttpRequest request = new DefaultFullHttpRequest(HTTP_1_1, GET, "/foo");
+        request.headers().add("host", "example.com");
```
dvlato: Shouldn't this be uppercase?

components/server/src/test/java/com/hotels/styx/server/netty/codec/UrlDecoderTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+import static io.netty.handler.codec.http.HttpHeaders.Names.HOST;
+import static io.netty.handler.codec.http.HttpMethod.GET;
+import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+public class UrlDecoderTest {
+    @Test
+    public void decodesOriginForm() {
```
dvlato: This test is redundant

components/server/src/test/java/com/hotels/styx/server/netty/codec/UrlDecoderTest.java
```diff
@@ -0,0 +1,117 @@
...
+        assertThat(url.encodedUri(), is("/foo"));
+        assertThat(url.isAbsolute(), is(false));
+        assertThat(url.isRelative(), is(true));
+        assertThat(url.host(), is(Optional.empty()));
+        assertThat(url.query(), is(Optional.empty()));
+        assertThat(url.scheme(), is(""));
+    }
+
+    @Test
+    public void decodesOriginFormIssue391() {
```
dvlato: Is using the issue number in the name of the test something we usually do? I would still have a name that tried to explain the issue though.

kvosper: Agreed. Let's describe the scenario in an issue-agnostic way for the test name, then maybe add the issue link in a comment above the test.

mikkokar:
Thanks for your comments @kvosper and @dvlato! I will address them soon.

components/server/src/test/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoderTest.java
mikkokar: By mistake. I have backed off this change.

components/server/src/test/java/com/hotels/styx/server/netty/codec/UrlDecoderTest.java
mikkokar: Actually it is different from the first test in the file. See the header case. But I have now made it different from the test below.

components/server/src/test/java/com/hotels/styx/server/netty/codec/UrlDecoderTest.java
mikkokar: Yes it should.

components/server/src/test/java/com/hotels/styx/server/netty/codec/UrlDecoderTest.java
mikkokar: I have made some changes to the test files. Please have another look.

PR #394, opened 2019-04-01
Use Kotlintest for unit-testing styx routing objects

mikkokar:
### Summary

* Replace Scala-based tests from styx-proxy project. 
* Add Kotlin-test framework to project and convert existing Scala-tests to Kotlin.

### Rationale:

Routing object tests rely heavily on yaml configuration that require multi-line strings. Both Kotlin & Scala support multi-line strings and are handy for this domain. But Kotlin collections and SAM conversions interoperate better with existing Java code. Perhaps it is just my Scala knowledge (or lack of) but Kotlin seems more effortless for this purpose.


components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/NonSanitizingGraphiteSenderTest.java
```diff
@@ -49,7 +49,7 @@ public void tearDown() {
         LocalNameServiceDescriptor.get().unset();
     }
 
-    @Test
+    @Test(enabled = false)
```
dvlato: Did you mean to disable this test?

components/proxy/src/test/kotlin/com/hotels/styx/ServerConfigSchemaTest.kt
```diff
@@ -0,0 +1,334 @@
...
+                  requestId:
+                    name: "X-Request-Id"
+              """.trimIndent()
+            )) shouldBe (Optional.empty())
+        }
+    }
+
+    describe("StyxHeaders Object") {
+
+        it("Aaccepts 'format' field for 'styxInfo' only") {
```
dvlato:
```suggestion
        it("Accepts 'format' field for 'styxInfo' only") {
```

components/proxy/src/test/kotlin/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistryTest.kt
```diff
@@ -0,0 +1,217 @@
...
+                it("should read backend services configured in the yaml file") {
+                    val registry = FileBackedBackendServicesRegistry.create("classpath:conf/origins/origins-for-jsontest.yml")
+
+                    StyxFutures.await(registry.start())
+                    val backendServices = registry.get()
+
+                    assert(backendServices.toList().size == 2)
+                    assert(backendServices.first() == backendServiceOne)
+
+                    StyxFutures.await(registry.stop())
```
dvlato: do we need to wait for the stop?  I suppose that means we cannot run the tests in parallel, it would be nice to improve that...

components/proxy/src/test/kotlin/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistryTest.kt
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    fun writeToFile(applications: List<BackendService>, path: String): File {
+        val output = File(createOnMissing(path))
+        Mapper.writer().writeValue(output, applications)
+        return output
+    }
+
+    fun createOnMissing(path: String): String {
+        val absolutePath = Paths.get(javaClass.getResource("/").toURI()).resolve(path).toFile()
```
dvlato: Why are you using both java.io and java.io here? is it because writeValue() above require a File?

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/BackendServiceProxyTest.kt
```diff
@@ -0,0 +1,128 @@
...
+import com.hotels.styx.routing.configBlock
+import com.hotels.styx.server.HttpInterceptorContext
+import io.kotlintest.shouldBe
+import io.kotlintest.shouldThrow
+import io.kotlintest.specs.StringSpec
+import reactor.core.publisher.Mono
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.CompletableFuture.completedFuture
+
+class BackendServiceProxyTest: StringSpec ({
```
dvlato: There's an space before the parenthesis, which is not consistent with the other tests.

components/proxy/src/test/kotlin/com/hotels/styx/routing/config/RoutingConfigParserTest.kt
```diff
@@ -0,0 +1,109 @@
...
+        """.trimIndent())
+
+        val routingObjectDef = RoutingConfigParser.toRoutingConfigNode(jsonNode)
+
+        routingObjectDef.shouldBeTypeOf<RouteHandlerDefinition>()
+
+        (routingObjectDef as RouteHandlerDefinition).name() shouldBe("main-router")
+        routingObjectDef.type() shouldBe("ConditionRouter")
+
+        routingObjectDef.config().shouldBeInstanceOf<JsonNode>()
```
dvlato: Since we saw that in case of an assertion error this only prints the objects toString() and not the classname,  we might want to somehow include a message with the actual class name .

components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/NonSanitizingGraphiteSenderTest.java
mikkokar: Hi @dvlato. Well, no and yes. Do you still remember that DNS unit test class we implemented ages ago? I think that caused the test to start failing. We need to think if these tests are sensible to keep around.

In the meanwhile I'll check if we can re-enable this.

components/proxy/src/test/kotlin/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistryTest.kt
mikkokar: Okay. I'll keep this in mind. This problem is inherited from the original Scala test. So it isn't introduced by this PR.

components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/NonSanitizingGraphiteSenderTest.java
mikkokar: Okay. This is now re-enabled. I had accidentally changed the Surefire plugin configuration, causing the failure.

components/proxy/src/test/kotlin/com/hotels/styx/proxy/backends/file/FileBackedBackendServicesRegistryTest.kt
mikkokar: This is also inherited from the original tests.

components/proxy/src/test/kotlin/com/hotels/styx/ServerConfigSchemaTest.kt
```diff
@@ -0,0 +1,334 @@
...
+package com.hotels.styx
+
+import com.hotels.styx.ServerConfigSchema.validateServerConfiguration
+import com.hotels.styx.infrastructure.configuration.ConfigurationParser
+import com.hotels.styx.infrastructure.configuration.ConfigurationSource.configSource
+import com.hotels.styx.infrastructure.configuration.yaml.YamlConfiguration
+import com.hotels.styx.infrastructure.configuration.yaml.YamlConfigurationFormat.YAML
+import io.kotlintest.shouldBe
+import io.kotlintest.specs.DescribeSpec
+import java.util.*
```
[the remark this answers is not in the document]
mikkokar: Good catch!

PR #401, opened 2019-04-15
Fix load test script on Mac OSX

mikkokar:
Small improvements:
- Specifically use python2.7 on Mac OSX
- Assume `wrk2` from home brew
- Optionally collect JFR logs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,11 @@ release: clean
 release-no-tests: clean
 	mvn install -Prelease,$(PLATFORM) -Dmaven.test.skip=true
 
+##
+## To run the load tests, install `wrk2` from homebrew (Mac OSX) using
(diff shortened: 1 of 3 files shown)


mikkokar:
Not sure what to do with this. We can't run the load tests locally any more. This is because we have scrapped the origins starter app. 

- We could run styx & origins in a docker image. But using this for load tests would work in Linux only. I don't know if Docker on other platforms still runs in a separate VM.

- We could use some other test origins on Mac OSX or other dev laptops. Back in the days we used to have Node.js test origins. But they seem to have disappeared from the code base. We can consider reviving them.

WDYT?

mikkokar:
Closing this old PR due to lack of recent activity.

PR #402, opened 2019-04-16
Implement shared class loading for matching paths

kvosper:
This PR addresses a situation where two or more plugins are loaded from the same JAR and they share information via HTTP Context.

In this scenario a plugin communicates to another one by storing a custom data class in the HTTP message context. The other plugin is supposed to read the stored data class. But because the two plugins are loaded with separate class loaders, the shared data class is loaded separately in each plugin, and therefore are not compatible with each other, resulting class cast exceptions like these:

```
java.lang.ClassCastException: x.y.MyDataClass cannot be cast to x.y.MyDataClass
```


kvosper:
This work requires more consideration on what our ideal behaviour is for plugins that share data/functionality between them.

PR #403, opened 2019-04-25
Validate http routing object configuration.

mikkokar:
Add capability to validate built-in HTTP routing object configurations.
Refactor config schema validator.


components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -137,61 +137,143 @@ public FieldValue value() {
...
-    public static class StringField extends FieldValue {
+    public static class StringField implements FieldValue {
+        @Override
+        public Optional<String> isCorrectType(List<String> parents, JsonNode parent, JsonNode element, Function<String, FieldValue> schemas) {
+            return element.isTextual() ? Optional.empty() : message(parents, element);
+        }
+
+        @Override
+        public String describe() {
+            return "STRING";
```
dvlato: Do we want to remove these hardcoded descriptions? Also, I don't know if we want to bring back a 'collection' of all the possible field types ... Something like the previous enum or a Kotlin sealed class.

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -217,12 +325,42 @@ public FieldValue elementType() {
...
+                if (correctType.isPresent()) {
+                    success.set(correctType);
+                }
+            });
+            return success.get();
+        }
+
+        private Optional<String> message(List<String> parents, JsonNode tree) {
+            String fullName = Joiner.on(".").join(parents);
+            return Optional.of(format("Unexpected field type. Field '%s' should be MAP ('%s'), but it is %s", fullName, this.elementType().describe(), tree.getNodeType()));
```
dvlato: *Here why I decided to add the comment to remove hardcoded strings for the field type. Still a minor issue of course...

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
```diff
@@ -30,133 +33,132 @@
...
+                atLeastOne("http", "https")
+        );
+
+        Schema.FieldValue logFormatSchema = object(
+                optional("enabled", bool()),
+                optional("longFormat", bool()),
+                atLeastOne("enabled", "longFormat")
+        );
+
+        STYX_SERVER_CONFIGURATION_SCHEMA_BUILDER = newDocument()
```
dvlato: We've seen users adding configuration parameters for their own benefit in the styx config; I understand those would break now. If we defined a styx branch/root and ignored the config parameters at the same level than styx in our validation, we would have a quick/dirty fix. Do you think it makes sense to add it?
E.g;
  styxConfig:
      proxy: ...
 customConfig: 
     thiswontfailvalidation:...

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -217,12 +325,42 @@ public FieldValue elementType() {
...
+
+        private Optional<String> validateMap(List<String> parents, JsonNode parent, JsonNode mapNode, Function<String, FieldValue> schemas) {
+            AtomicReference<Optional<String>> success = new AtomicReference<>(Optional.empty());
+
+            mapNode.fieldNames().forEachRemaining(key -> {
+                JsonNode entry = mapNode.get(key);
+                Optional<String> correctType = elementType.isCorrectType(push(parents, format("%s", key)), mapNode, entry, schemas);
+
+                if (correctType.isPresent()) {
+                    success.set(correctType);
```
kvosper: This doesn't seem right. Won't a correctly typed entry override a previous incorrectly typed entry to make the map "correct" overall?

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -232,23 +370,74 @@ public FieldValue elementType() {
...
+        Set<String> knownFields = ImmutableSet.copyOf(schema.fieldNames());
+
+        fieldsPresent.forEach(name -> {
+            if (!knownFields.contains(name)) {
+                throw new SchemaValidationException(format("Unexpected field: '%s.%s'", prefix, name));
+            }
+        });
+    }
+
+    static List<String> push(List<String> currentList, String element) {
```
kvosper:
```suggestion
    private static List<String> concatenate(List<String> currentList, String element) {
```

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -232,23 +370,74 @@ public FieldValue elementType() {
...
+        ArrayList<String> newList = new ArrayList<>(currentList);
```
kvosper:
```suggestion
        List<String> newList = new ArrayList<>(currentList);
```

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -270,6 +459,31 @@ public String schemaName() {
         public String discriminatorFieldName() {
             return discriminatorField;
         }
+
+        @Override
+        public Optional<String> isCorrectType(List<String> parents, JsonNode parent, JsonNode element, Function<String, FieldValue> schemas) {
+            JsonNode type = parent.get(discriminatorField);
+
```
kvosper: Would be helpful to have some comments in this logic to explain what/why.

components/common/src/test/java/com/hotels/styx/config/validator/DocumentFormatTest.java
```diff
@@ -443,86 +445,127 @@ public void validatesSubobjects() throws Exception {
...
-                        field("plugins", object("PluginsList"))
+                .typeExtension("ProxyTo", proxyTo)
+                .typeExtension("Redirection", redirection)
+                .rootSchema(object(
+                        field("httpPipeline", new Schema.RoutingObjectSpec())
                 ))
                 .build();
 
-        assertThat(validator.validateObject(rootObject), is(true));
+        boolean outcome1 = validator.validateObject(
```
kvosper: These variable names do not describe the difference in scenarios being tested. Should there be two different test methods instead?

components/common/src/test/java/com/hotels/styx/config/validator/DocumentFormatTest.java
```diff
@@ -443,86 +445,127 @@ public void validatesSubobjects() throws Exception {
...
+        DocumentFormat validator = newDocument()
+                .rootSchema(object(
+                        field("httpPipeline", object(
+                                field("type", string()),
+                                field("config", union("type"))
+                        ))
                 ))
                 .build();
+
+        boolean outcome1 = validator.validateObject(
```
kvosper:
```suggestion
        boolean outcome = validator.validateObject(
```

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
```diff
@@ -30,133 +33,132 @@
...
+    }
+
 
     private ServerConfigSchema() {
     }
 
     static Optional<String> validateServerConfiguration(YamlConfiguration yamlConfiguration) {
         try {
-            STYX_SERVER_CONFIGURATION_SCHEMA.validateObject(yamlConfiguration.root());
+            for (Map.Entry<String, Schema.FieldValue> schema: RoutingObjectFactory.BUILTIN_HANDLER_SCHEMAS.entrySet()) {
```
kvosper: This for-loop can be replaced with `BUILTIN_HANDLER_SCHEMAS.forEach(STYX_SERVER_CONFIGURATION_SCHEMA_BUILDER::typeExtension);`

components/proxy/src/main/java/com/hotels/styx/routing/config/RoutingObjectFactory.java
```diff
@@ -42,15 +43,36 @@
...
  */
 public class RoutingObjectFactory {
-    private static final Map<String, HttpHandlerFactory> BUILTIN_HANDLERS = ImmutableMap.<String, HttpHandlerFactory>builder()
-            .put("StaticResponseHandler", new StaticResponseHandler.Factory())
-            .put("ConditionRouter", new ConditionRouter.Factory())
-            .put("InterceptorPipeline", new HttpInterceptorPipeline.Factory())
-            .put("ProxyToBackend", new ProxyToBackend.Factory())
-            .build();
+    public static final ImmutableMap<String, Schema.FieldValue> BUILTIN_HANDLER_SCHEMAS;
+
```
kvosper: These fields don't need to have blank lines between them.

components/proxy/src/test/kotlin/com/hotels/styx/ServerConfigSchemaTest.kt
```diff
@@ -163,6 +163,14 @@ class ServerConfigSchemaTest : DescribeSpec({
         }
     }
 
+    describe("Incorrect configuration") {
+        it("Detects unknown configuration options") {
+            validateServerConfiguration(yamlConfig(minimalConfig + """
+                abc: 1
+            """.trimIndent())) shouldBe (Optional.of("Unexpected field: '.abc'"))
```
kvosper: Why is the field name prefixed with `.` - shouldn't that be just for delimiting between parts of the name?

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
mikkokar: The "collection of all possible types" was not very accommodating for extension, and therefore had to be removed. 

This new, refactored validator framework is now open for extension. For example, `RoutingObjectSpec` could have been implemented elsewhere and simply dropped in to the validator tree. This is only possible when each validator is just a self describing class implementing a `FieldValue` interface. 

It would have been really cumbersome to achieve this with the previous "collection of all possible types" implementation.

For the same reason it is necessary for each type to describe themselves for human readable error reporting.

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
mikkokar: The element type can be anything. Even a nested map. For this reason the `describe()` is really important. If I understood your comment correctly? :-)

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
mikkokar: TBH, there could be still better, off the shelf libraries for the validation. Ideally we'll get to deprecate this code some day.

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
mikkokar: Well, for this purpose there is:

```
                            optional("userDefined", object(opaque())),
```

Any custom attributes would go to that `userDefined` namespace, which is entirely opaque to the validator.

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -137,61 +137,143 @@ public FieldValue value() {
...
     /**
      * Represents a type of a schema field.
      */
-    public static class FieldValue {
-        public FieldType type() {
-            if (this instanceof IntegerField) {
-                return FieldType.INTEGER;
-            }
+    public interface FieldValue {
```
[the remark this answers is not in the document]
mikkokar: Correct. I will fix this.

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -137,61 +137,143 @@ public FieldValue value() {
...
-            if (this instanceof ListField) {
-                return FieldType.LIST;
+            FieldValue routingObjectDefinition = object(
+                    field("type", string()),
+                    optional("name", string()),
+                    optional("tags", list(string())),
+                    field("config", union("type"))
+            );
+
+            Optional<String> correctType = ref.isCorrectType(parents, parent, element, schemas);
```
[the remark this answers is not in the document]
mikkokar: 👍

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -137,61 +137,143 @@ public FieldValue value() {
...
             }
 
-            throw new IllegalStateException("Unknown field type: " + this.getClass() + " " + this);
+            // Failed
+            return message(parents, element);
+        }
+
+        private Optional<String> message(List<String> parents, JsonNode tree) {
+            String fullName = Joiner.on(".").join(parents);
+            return Optional.of(format("Unexpected field type. Field '%s' should be %s, but it is %s", fullName, describe(), tree.getNodeType()));
```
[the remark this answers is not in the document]
mikkokar: Hi @kvosper 

Regarding your comment:
> (except the describe is sometimes hardcoded for no reason)

`describe` provides a type information for validation error messages. Therefore they need to be informative for end users. As a result, class names etc cannot and won't be used.

Or did you have something else in mind?

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
mikkokar: Renamed local variables. Hopefully the code logic is clearer now.

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
mikkokar: +1 for private.
But no for `concatenate`. This is supposed to behave like a stack, and therefore `push` is more appropriate.

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
mikkokar: +1

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
mikkokar: 👍

components/proxy/src/main/java/com/hotels/styx/routing/config/RoutingObjectFactory.java
mikkokar: 👍

components/proxy/src/test/kotlin/com/hotels/styx/ServerConfigSchemaTest.kt
mikkokar: Okay. This allows us to avoid a special case in the code.

But let me have another look into this. I also would like to improve the messages.

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -137,61 +137,143 @@ public FieldValue value() {
...
     /**
      * Represents a type of a schema field.
      */
-    public static class FieldValue {
-        public FieldType type() {
-            if (this instanceof IntegerField) {
-                return FieldType.INTEGER;
-            }
+    public interface FieldValue {
+        Optional<String> isCorrectType(List<String> parents, JsonNode parent, JsonNode element, Function<String, FieldValue> schemas);
```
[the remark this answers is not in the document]
mikkokar: Adding this:

        /**
         * Validates a JsonNode value against this type.  Typically checks that a value
         * can be used as an instance of this type.
         *
         * @param parents A parentage attribute stack.
         * @param parent Value's parent node.
         * @param value Value to be inspected.
         * @param typeExtensions Provides type extensions.
         * @return
         */

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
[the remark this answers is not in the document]
mikkokar: I have now created a common `message` method.

components/common/src/test/java/com/hotels/styx/config/validator/DocumentFormatTest.java
mikkokar: Refactored. The variable names are no longer there.

components/common/src/test/java/com/hotels/styx/config/validator/DocumentFormatTest.java
mikkokar: Refactored these away.

components/proxy/src/test/kotlin/com/hotels/styx/ServerConfigSchemaTest.kt
mikkokar: I have addressed this issue.

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
```diff
@@ -30,133 +31,130 @@
...
+    }
+
 
     private ServerConfigSchema() {
     }
 
     static Optional<String> validateServerConfiguration(YamlConfiguration yamlConfiguration) {
         try {
-            STYX_SERVER_CONFIGURATION_SCHEMA.validateObject(yamlConfiguration.root());
+            BUILTIN_HANDLER_SCHEMAS.forEach(STYX_SERVER_CONFIGURATION_SCHEMA_BUILDER::typeExtension);
```
kvosper: Shouldn't `BUILTIN_HANDLER_SCHEMAS.forEach(STYX_SERVER_CONFIGURATION_SCHEMA_BUILDER::typeExtension);` be in the static initialiser instead? So that if we call this method repeatedly (e.g. in tests) we won't repeatedly add the same schemas?

components/proxy/src/main/java/com/hotels/styx/routing/config/RoutingObjectFactory.java
```diff
@@ -42,15 +43,31 @@
...
-            .build();
+    public static final ImmutableMap<String, Schema.FieldType> BUILTIN_HANDLER_SCHEMAS;
+    private static final ImmutableMap<String, HttpHandlerFactory> BUILTIN_HANDLER_FACTORIES;
+    private static final String STATIC_RESPONSE = "StaticResponseHandler";
+    private static final String CONDITION_ROUTER = "ConditionRouter";
+    private static final String INTERCEPTOR_PIPELINE = "InterceptorPipeline";
+    private static final String PROXY_TO_BACKEND = "ProxyToBackend";
+
+    static {
+        BUILTIN_HANDLER_FACTORIES = ImmutableMap.<String, HttpHandlerFactory>builder()
```
kvosper: Looks like we can just use
```
BUILTIN_HANDLER_FACTORIES = ImmutableMap.of(
                STATIC_RESPONSE, new StaticResponseHandler.Factory(),
                CONDITION_ROUTER, new ConditionRouter.Factory(),
                INTERCEPTOR_PIPELINE, new HttpInterceptorPipeline.Factory(),
                PROXY_TO_BACKEND, new ProxyToBackend.Factory());
```
instead of the builder here.
Probably the same for `BUILTIN_HANDLER_SCHEMAS` as well.

components/common/src/main/java/com/hotels/styx/config/schema/Schema.java
```diff
@@ -98,16 +110,69 @@ public boolean ignore() {
...
-        BOOLEAN,
-        OBJECT,
-        LIST,
-        MAP;
+    public interface FieldType {
+        /**
+         * Validates a JsonNode value against this type.  Typically checks that a value
+         * can be used as an instance of this type.
+         *
+         * @param parents A parentage attribute stack.
```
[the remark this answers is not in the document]
mikkokar: ok

components/common/src/test/java/com/hotels/styx/config/schema/SchemaTest.java
```diff
@@ -15,49 +15,725 @@
...
 public class SchemaTest {
+    private final ObjectMapper YAML_MAPPER = new ObjectMapper(new YAMLFactory())
+            .configure(FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .configure(AUTO_CLOSE_SOURCE, true);
+    private final Function<String, Schema.FieldType> EXTENSIONS = key -> null;
+
+
+    @Test(expectedExceptions = SchemaValidationException.class,
+            expectedExceptionsMessageRegExp = "Unexpected field type. Field 'myNokValue' should be INTEGER, but it is BOOLEAN")
+    public void integer_validatesIntegerValues() throws Exception {
```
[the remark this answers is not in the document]
mikkokar: It is supposed to serve as a grouping of tests. I would assume this is self evident and doesn't need further explaining.

If this becomes an issue we could change to Kotlin tests in future. IIRC we do use this convention in some other tests too, where we have a large number of unit tests crammed into one file.

components/proxy/src/main/java/com/hotels/styx/routing/config/RoutingObjectFactory.java
mikkokar: `ImmutableMap.of` takes maximum of four or five values. So after one (or two) more we'd have to revert back to builder. Therefore I'd leave it like this.

PR #404, opened 2019-04-29
Add insert/remove/watch functionality to Object Store

mikkokar:
Adds Object Store functionality for

- inserting new objects
- removing existing objects
- watching for changes

Concurrency issues:

- Objects are stored in a persistent map. Each modification (insert/remove) creates a new, immutable object store snapshot. The snapshot is then serially published to all watchers.

Watching for Changes:

- Consumers can listen for database changes by calling a `watch` method. It supports watching for the full database only. Watching for individual objects will be implemented in future when necessary.

- A `watch` returns a reactive streams `Publisher`. When subscribed, the publisher emits the current object store snapshot, followed by a new snapshot every time database changes. The publisher keeps emitting the change events in perpetuity until unsubscribed.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/configuration/ObjectStore.java
+++ b/components/api/src/main/java/com/hotels/styx/api/configuration/ObjectStore.java
@@ -28,8 +28,8 @@ public interface ObjectStore<T> {
     /**
      * A repository of named objects.
      *
-     * @param key Object's name.
-     * @return An object if known.
+     * @param key object's name
+     * @return an object if known
      */
     Optional<T> get(String key);
 }
--- a/components/api/src/main/java/com/hotels/styx/api/configuration/ObjectStore.java
+++ b/components/api/src/main/java/com/hotels/styx/api/configuration/ObjectStore.java
@@ -23,7 +23,7 @@ import java.util.Optional;
  *
  * @param <T> Stored object type
  */
-public interface ObjectStoreSnapshot<T> {
+public interface ObjectStore<T> {
 
     /**
      * A repository of named objects.
--- a/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
@@ -15,28 +15,79 @@
  */
 package com.hotels.styx.routing.db;
 
-import com.hotels.styx.api.configuration.ObjectStore
+import com.hotels.styx.api.configuration.ObjectStoreSnapshot
+import org.pcollections.HashTreePMap
+import org.pcollections.PMap
+import org.reactivestreams.Publisher
+import reactor.core.publisher.Flux
 import java.util.Optional
-import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CopyOnWriteArrayList
+import java.util.concurrent.atomic.AtomicReference
+import java.util.function.Consumer
 
 /**
  * Styx Route Database.
  */
-class StyxObjectStore<T> : ObjectStore<T> {
-    private val objects = ConcurrentHashMap<String, Record<T>>();
+
+private interface MyListener<T> : Consumer<ObjectStoreSnapshot<T>>
+
+class StyxObjectStore<T> : ObjectStoreSnapshot<T> {
+    private val objects: AtomicReference<PMap<String, Record<T>>> = AtomicReference(HashTreePMap.empty())
+
+    private val listeners = CopyOnWriteArrayList<MyListener<T>>()
 
     override fun get(name: String?): Optional<T> {
-        return Optional.ofNullable(objects.get(name))
+        return Optional.ofNullable(objects.get().get(name))
                 .map { it.payload }
     }
 
-    private fun insert(key: String, tags: Set<String>, payload: T) {
-        objects.put(key, Record(key, tags, payload))
-    }
-
-    override fun insert(key: String, record: T) {
+    fun insert(key: String, record: T) {
         insert(key, setOf(), record)
     }
 
-    data class Record<T>(val key: String, val tags:Set<String>, val payload: T)
+    /**
+     * Returns a Publisher that emits an event for any modification (insert/remove).
+     *
+     * TODO: Should return a database snapshot?
+     *       Should return a change indication (insert "foo", remove "bar")?
+     *
+     * Watch activates on subscription only.
+     * Watch removed on unsubscription.
+     */
+    fun watch(): Publisher<ObjectStoreSnapshot<T>> {
+        return Flux.push { sink ->
+            val listener = object : MyListener<T> {
+                override fun accept(objectStoreSnapshot: ObjectStoreSnapshot<T>) {
+                    sink.next(objectStoreSnapshot)
+                }
+            }
+
+            listeners.add(listener)
+
+            sink.onCancel {
+                listeners.remove(listener)
+            }
+
+            sink.onDispose {
+                listeners.remove(listener)
+            }
+        }
+    }
+
+    fun watchers() = listeners.size
+
+    private fun insert(key: String, tags: Set<String>, payload: T) {
+        val objectsV2 = objects.get().plus(key, Record(key, tags, payload))
+        objects.set(objectsV2)
+
+        listeners.forEach { listener -> listener.accept({
+            key -> Optional
+                .ofNullable(objectsV2.get(key))
+                .map { it.payload } }
+        )}
+
+    }
+
+
+    data class Record<T>(val key: String, val tags: Set<String>, val payload: T)
 }
--- a/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
@@ -20,6 +20,7 @@ import org.pcollections.HashTreePMap
 import org.pcollections.PMap
 import org.reactivestreams.Publisher
 import reactor.core.publisher.Flux
+import reactor.core.publisher.FluxSink
 import java.util.Optional
 import java.util.concurrent.CopyOnWriteArrayList
 import java.util.concurrent.Executors
@@ -107,15 +108,18 @@ class StyxObjectStore<T> : ObjectStore<T> {
             }
 
             watchers.add(watcher)
-
             queue {
-                sink.next(ObjectStore { key ->
-                    Optional
-                            .ofNullable(objects().get(key))
-                            .map { it.payload }
-                })
+                emitInitialSnapshot(sink)
             }
         }
+    }
+
+    private fun emitInitialSnapshot(sink: FluxSink<ObjectStore<T>>) {
+        sink.next(ObjectStore { key ->
+            Optional
+                    .ofNullable(objects().get(key))
+                    .map { it.payload }
+        })
     }
 
     internal fun watchers() = watchers.size
--- a/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
@@ -15,7 +15,7 @@
  */
 package com.hotels.styx.routing.db;
 
-import com.hotels.styx.api.configuration.ObjectStoreSnapshot
+import com.hotels.styx.api.configuration.ObjectStore
 import org.pcollections.HashTreePMap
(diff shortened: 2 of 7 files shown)


components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
```diff
@@ -15,28 +15,100 @@
...
-    private val objects = ConcurrentHashMap<String, Record<T>>();
+class StyxObjectStore<T> : ObjectStoreSnapshot<T> {
+    private val objects: AtomicReference<PMap<String, Record<T>>> = AtomicReference(HashTreePMap.empty())
+    private val watchers = CopyOnWriteArrayList<ChangeWatcher<T>>()
+    private val executor = Executors.newSingleThreadExecutor()
 
-    override fun get(name: String?): Optional<T> {
-        return Optional.ofNullable(objects.get(name))
+    override fun get(name: String): Optional<T> {
+        return Optional.ofNullable(objects().get(name))
```
dvlato: Would you prefer to use the indexing operator? objects()[name] .We could also do objects() -  instead of minus() , etc. but all these seem a bit 'magical' compared to the more verbose method invocations.

components/proxy/src/test/kotlin/com/hotels/styx/routing/db/StyxObjectStoreTest.kt
```diff
@@ -36,6 +51,185 @@ class StyxObjectStoreTest : FeatureSpec() {
...
+            scenario("Maintains database integrity in concurrent operations") {
+                val db = StyxObjectStore<String>()
+                val executor = Executors.newFixedThreadPool(4)
+
+                for (i in 1..1000) {
+                    executor.execute { db.insert("redirect-$i", "Record-$i") }
+                }
+
+                executor.shutdown()
+                executor.awaitTermination(2, TimeUnit.SECONDS)
```
kvosper: Please import `SECONDS`.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
```diff
@@ -15,28 +15,100 @@
...
 
 /**
  * Styx Route Database.
  */
-class StyxObjectStore<T> : ObjectStore<T> {
-    private val objects = ConcurrentHashMap<String, Record<T>>();
+class StyxObjectStore<T> : ObjectStoreSnapshot<T> {
+    private val objects: AtomicReference<PMap<String, Record<T>>> = AtomicReference(HashTreePMap.empty())
+    private val watchers = CopyOnWriteArrayList<ChangeWatcher<T>>()
+    private val executor = Executors.newSingleThreadExecutor()
```
dvlato: This Executor will never be stopped, this seems relevant for tests where we could instantiate the class several times. If we think that production code 'should be ok since there's a single StyxObjectStore that will exist for the whole lifetime of the application', it makes sense to convert this class into an object so the singleton nature is depicted.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
```diff
@@ -15,28 +15,100 @@
...
+
+    /**
+     * Returns a Publisher that emits an event at any modification.
+     *
+     * Watch activates on subscription only.
+     * Watch removed on unsubscription.
+     */
+    fun watch(): Publisher<ObjectStoreSnapshot<T>> {
+        return Flux.push { sink ->
+            val watcher = object : ChangeWatcher<T> {
```
dvlato: Can't we simplify this?

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
```diff
@@ -15,28 +15,100 @@
...
-    override fun insert(key: String, record: T) {
-        insert(key, setOf(), record)
+    private fun notifyWatchers(objectsV2: PMap<String, Record<T>>) {
+        watchers.forEach { listener ->
+            listener.accept(snapshot(objectsV2))
+        }
     }
 
-    data class Record<T>(val key: String, val tags:Set<String>, val payload: T)
+    private fun snapshot(snapshot: PMap<String, Record<T>>) = object : ObjectStoreSnapshot<T> {
```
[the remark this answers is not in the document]
mikkokar: I believe Kotlin supports SAM conversions only for Java interoperability. Not for within Kotlin:

> Also note that this feature works only for Java interop; since Kotlin has proper function types, automatic conversion of functions into implementations of Kotlin interfaces is unnecessary and therefore unsupported.

(from: https://kotlinlang.org/docs/reference/java-interop.html#sam-conversions)

This suggests you cannot substitute `ObjectStore` return type for Lambda. Or do you see something else?

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
mikkokar: Hi @dvlato is the indexing operator identical in its behaviour when a key is not found? Specifically, will it return a null value or does it just throw?

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
mikkokar: How about, if for now, I'd just declare `executor` in the companion object? Like so:
    
    companion object {
        val executor = Executors.newSingleThreadExecutor()
    }

This addresses thread leak in tests, and will work well in production code. Also this executor is likely to have a very low work load, it would work sufficiently well in production code even with multiple ObjectStore implementations/instances.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
mikkokar: Looks like we can. By using Kotlin type alias instead of Java interfaces. Given:

    private typealias ChangeWatcher<T> = (ObjectStore<T>) -> Unit

Then:

    val watcher: ChangeWatcher<T> = { sink.next(it) }

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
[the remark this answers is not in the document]
mikkokar: Okay, it now looks like this:

    private fun snapshot(snapshot: PMap<String, Record<T>>) = ObjectStore<T> { key: String ->
        Optional
                .ofNullable(snapshot.get(key))
                .map { it.payload }
    }

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
dvlato: It sounds safe enough (unless the test framework uses several classloaders to load the class! of course).

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
dvlato: Pretty nice :)

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
dvlato: The indexing operator is translated by an invocation to a get() function, so it will be completely identical, yes. See here how operator overloading works in Kotlin: https://kotlinlang.org/docs/reference/operator-overloading.html

components/proxy/src/test/kotlin/com/hotels/styx/routing/db/StyxObjectStoreTest.kt
```diff
@@ -36,6 +52,224 @@ class StyxObjectStoreTest : FeatureSpec() {
                 db.get("notfound") shouldBe Optional.empty()
             }
         }
+
+        feature("Insert") {
+            scenario("Notifies watchers for any change") {
+                val db = StyxObjectStore<String>()
+
+                StepVerifier.create<ObjectStore<String>>(db.watch())
```
kvosper:
```suggestion
                StepVerifier.create(db.watch())
```

components/proxy/src/test/kotlin/com/hotels/styx/routing/db/StyxObjectStoreTest.kt
```diff
@@ -36,6 +52,224 @@ class StyxObjectStoreTest : FeatureSpec() {
...
+                    for (i in 1..1000) {
+                        db.get("redirect-$i") shouldBe (Optional.of("Record-$i"))
+                    }
+                }
+            }
+
+            scenario("Replaces already existing object") {
+                val db = StyxObjectStore<String>()
+
+                StepVerifier.create<ObjectStore<String>>(db.watch())
```
kvosper:
```suggestion
                StepVerifier.create(db.watch())
```

components/proxy/src/test/kotlin/com/hotels/styx/routing/db/StyxObjectStoreTest.kt
```diff
@@ -36,6 +52,224 @@ class StyxObjectStoreTest : FeatureSpec() {
...
+                eventually(1.seconds, java.lang.AssertionError::class.java) {
+                    db.get("x") shouldBe Optional.empty()
+                    db.get("y") shouldBe Optional.empty()
+                }
+            }
+
+            scenario("Non-existent object doesn't trigger watchers") {
+                val db = StyxObjectStore<String>()
+
+                StepVerifier.create<ObjectStore<String>>(db.watch())
```
kvosper:
```suggestion
                StepVerifier.create(db.watch())
```

components/proxy/src/test/kotlin/com/hotels/styx/routing/db/StyxObjectStoreTest.kt
```diff
@@ -36,6 +52,224 @@ class StyxObjectStoreTest : FeatureSpec() {
...
+                        .expectNextCount(1)
+                        .then { db.remove("x") }
+                        .expectNoEvent(250.milliseconds)
+                        .thenCancel()
+                        .verify()
+
+                db.insert("y", "Y")
+
+                StepVerifier.create<ObjectStore<String>>(db.watch())
```
kvosper:
```suggestion
                StepVerifier.create(db.watch())
```

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
```diff
@@ -16,27 +16,136 @@
...
+            val watcher: ChangeWatcher<T> = { sink.next(it) }
+
+            sink.onDispose {
+                watchers.remove(watcher)
+            }
+
+            watchers.add(watcher)
+
+            queue {
+                sink.next(ObjectStore { key ->
```
dvlato: This watch() method is doing quite a few things so I would probably extract functions that can have a descriptive name, especially for this last bit.

mikkokar: Hmm... I could do something like this:

    fun watch(): Publisher<ObjectStore<T>> {
        return Flux.push { sink ->
            addWatcher(sink)

            queue {
                emitInitialSnapshot(sink)
            }
        }
    }

But then again `addWatcher` would not only create and add a watcher, but it would also configure `sink.onDispose` callback. Seems bit contrived. 

But I could still factor out `emitInitialSnapshot` method as above.

components/api/src/main/java/com/hotels/styx/api/configuration/ObjectStore.java
```diff
@@ -26,21 +26,10 @@
 public interface ObjectStore<T> {
 
     /**
-     * Gets an object with given key.
+     * A repository of named objects.
      *
-     * @param key
-     * @return
+     * @param key Object's name.
```
kvosper:
```suggestion
     * @param key object's name
```

components/api/src/main/java/com/hotels/styx/api/configuration/ObjectStore.java
```diff
@@ -26,21 +26,10 @@
...
+     * @return An object if known.
```
kvosper:
```suggestion
     * @return an object if known
```

PR #406, opened 2019-05-02
Open a RESTful API for Styx Object Store

mikkokar:
I'm adding a new `UrlPatternRouter`. To configure this router, you specify a regular expression pattern for each endpoint per HTTP verb. It also captures patterns from URL and exposes them as values in HTTP context. See the unit tests as an example.

I have also added a new `RoutingObjectHandler` that builds on `UrlPatternRouter`. It allows Styx Routing Object Store to be modified via RESTful API. It supports, addition, removal, and replacement of routing objects.

The ObjectStore is updated accordingly, to support these features. Notice that ObjectStore concurrency model is improved. Object store modifications (insert/remove) now take place, always, in the calling thread. The executor simply delivers the watch notifications.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
@@ -88,11 +88,11 @@ public class AdminServerBuilder {
 
     private Registry<BackendService> backendServicesRegistry;
 
-    public AdminServerBuilder(Environment environment, StyxServerComponents serverComponents) {
-        this.environment = environment;
-        this.configuration = environment.configuration();
+    public AdminServerBuilder(StyxServerComponents serverComponents) {
+        this.environment = requireNonNull(serverComponents.environment());
         this.routeDatabase = requireNonNull(serverComponents.routeDatabase());
         this.routingObjectFactory = requireNonNull(serverComponents.routingObjectFactory());
+        this.configuration = this.environment.configuration();
     }
 
     public AdminServerBuilder backendServicesRegistry(Registry<BackendService> backendServicesRegistry) {
@@ -133,7 +133,6 @@ public class AdminServerBuilder {
 
         RoutingObjectHandler routingObjectHandler = new RoutingObjectHandler(routeDatabase, routingObjectFactory);
         httpRouter.add("/admin/routing", routingObjectHandler);
-
         httpRouter.add("/admin/routing/", routingObjectHandler);
 
         // Dashboard
--- a/components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
@@ -32,6 +32,7 @@ import com.hotels.styx.admin.handlers.OriginsInventoryHandler;
 import com.hotels.styx.admin.handlers.PingHandler;
 import com.hotels.styx.admin.handlers.PluginListHandler;
 import com.hotels.styx.admin.handlers.PluginToggleHandler;
+import com.hotels.styx.admin.handlers.RoutingObjectHandler;
 import com.hotels.styx.admin.handlers.StartupConfigHandler;
 import com.hotels.styx.admin.handlers.StyxConfigurationHandler;
 import com.hotels.styx.admin.handlers.ThreadsHandler;
@@ -45,12 +46,16 @@ import com.hotels.styx.api.extension.service.spi.Registry;
 import com.hotels.styx.common.http.handler.HttpMethodFilteringHandler;
 import com.hotels.styx.common.http.handler.StaticBodyHttpHandler;
 import com.hotels.styx.proxy.plugin.NamedPlugin;
+import com.hotels.styx.routing.RoutingObjectRecord;
+import com.hotels.styx.routing.config.RoutingObjectFactory;
+import com.hotels.styx.routing.db.StyxObjectStore;
 import com.hotels.styx.server.HttpServer;
 import com.hotels.styx.server.StandardHttpRouter;
 import com.hotels.styx.server.handlers.ClassPathResourceHandler;
 import com.hotels.styx.server.netty.NettyServerBuilderSpec;
 import com.hotels.styx.server.netty.WebServerConnectorFactory;
 import com.hotels.styx.server.track.CurrentRequestTracker;
+import com.hotels.styx.startup.StyxServerComponents;
 import org.slf4j.Logger;
 
 import java.time.Duration;
@@ -78,12 +83,16 @@ public class AdminServerBuilder {
 
     private final Environment environment;
     private final Configuration configuration;
+    private final RoutingObjectFactory routingObjectFactory;
+    private StyxObjectStore<RoutingObjectRecord> routeDatabase;
 
     private Registry<BackendService> backendServicesRegistry;
 
-    public AdminServerBuilder(Environment environment) {
+    public AdminServerBuilder(Environment environment, StyxServerComponents serverComponents) {
         this.environment = environment;
         this.configuration = environment.configuration();
+        this.routeDatabase = requireNonNull(serverComponents.routeDatabase());
+        this.routingObjectFactory = requireNonNull(serverComponents.routingObjectFactory());
     }
 
     public AdminServerBuilder backendServicesRegistry(Registry<BackendService> backendServicesRegistry) {
@@ -121,6 +130,11 @@ public class AdminServerBuilder {
         httpRouter.add("/admin/origins/status", new OriginsInventoryHandler(environment.eventBus()));
         httpRouter.add("/admin/configuration/logging", new LoggingConfigurationHandler(styxConfig.startupConfig().logConfigLocation()));
         httpRouter.add("/admin/configuration/startup", new StartupConfigHandler(styxConfig.startupConfig()));
+
+        RoutingObjectHandler routingObjectHandler = new RoutingObjectHandler(routeDatabase, routingObjectFactory);
+        httpRouter.add("/admin/routing", routingObjectHandler);
+
+        httpRouter.add("/admin/routing/", routingObjectHandler);
 
         // Dashboard
         httpRouter.add("/admin/dashboard/data.json", dashboardDataHandler(styxConfig));
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/RoutingObjectHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/RoutingObjectHandler.java
@@ -33,12 +33,9 @@ import com.hotels.styx.routing.config.RoutingObjectDefinition;
 import com.hotels.styx.routing.config.RoutingObjectFactory;
 import com.hotels.styx.routing.db.StyxObjectStore;
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import static com.fasterxml.jackson.core.JsonParser.Feature.AUTO_CLOSE_SOURCE;
 import static com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES;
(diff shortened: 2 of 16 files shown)


components/proxy/src/main/java/com/hotels/styx/admin/handlers/UrlPatternRouter.java
```diff
@@ -0,0 +1,163 @@
...
+        public HttpHandler handler() {
+            return handler;
+        }
+
+        public List<String> placeholderNames() {
+            return placeholderNames;
+        }
+
+        private Pattern compilePattern(String pattern) {
+            Pattern x = Pattern.compile(":([a-zA-Z0-9-_]+)");
```
kvosper: This `x` can be a constant. And needs a proper name.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UrlPatternRouter.java
```diff
@@ -0,0 +1,163 @@
...
+            Matcher matcher = x.matcher(pattern);
+            return Pattern.compile(matcher.replaceAll("(?<$1>[a-zA-Z0-9-_]+)"));
+        }
+
+        private List<String> placeholcers(String pattern) {
+            Pattern x = Pattern.compile(":([a-zA-Z0-9-_]+)");
```
kvosper: Use constant as discussed above.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
```diff
@@ -82,13 +93,18 @@ class StyxObjectStore<T> : ObjectStore<T> {
...
-            if (nextVersion != objects()) {
-                objects.set(nextVersion)
-                notifyWatchers(nextVersion)
+        while (!objects.compareAndSet(current, new)) {
+            current = objects.get()
+            new = current.minus(key)
+        }
+
+        queue {
+            if (current != new) {
```
kvosper: Could we check this condition before deciding to queue a task?

components/proxy/src/test/kotlin/com/hotels/styx/admin/handlers/RoutingObjectHandlerTest.kt
```diff
@@ -0,0 +1,274 @@
...
+            routeDatabase.get("staticResponse").get().type shouldBe "StaticResponseHandler"
+            routeDatabase.get("staticResponse").get().config.shouldBeTypeOf<ObjectNode>()
+            routeDatabase.get("staticResponse").get().handler.shouldBeTypeOf<StaticResponseHandler>()
+        }
+
+        scenario("Retrieving objects") {
+            val handler = RoutingObjectHandler(routeDatabase, objectFactory)
+
+            handler.handle(
+                    HttpRequest.put("/admin/routing/objects/staticResponse")
```
kvosper: There are lots of tests that are set up this way. There should be some kind of shared setUp method for them (not for the test name `Injecting new objects` though, since the `put` is the thing being tested there).

components/proxy/src/test/kotlin/com/hotels/styx/admin/handlers/RoutingObjectHandlerTest.kt
```diff
@@ -0,0 +1,274 @@
...
+                              """
+                    .trimIndent()
+                    .trim()
+        }
+
+        scenario("Fetching all routing objects") {
+            val handler = RoutingObjectHandler(routeDatabase, objectFactory)
+
+            handler.handle(
+                    HttpRequest.put("/admin/routing/objects/staticResponse")
```
kvosper: We should be able to import these `HttpRequest` factory methods (`put`, `get` etc.)

components/proxy/src/test/kotlin/com/hotels/styx/admin/handlers/UrlPatternRouterTest.kt
```diff
@@ -0,0 +1,217 @@
...
+            val router = UrlPatternRouter.Builder()
+                    .post("/admin/apps/:appId/:originId") { request, context -> throw RuntimeException("Something went wrong") }
+                    .build()
+
+            val response = Mono.from(
+                    router.handle(post("/admin/apps/appx/appx-01").build(), HttpInterceptorContext.create())
+                            .flatMap { it.aggregate(10000) })
+                    .block()
+
+            response!!.status() shouldBe HttpResponseStatus.INTERNAL_SERVER_ERROR
```
kvosper:
```suggestion
            response!!.status() shouldBe INTERNAL_SERVER_ERROR
```

components/proxy/src/test/kotlin/com/hotels/styx/routing/db/StyxObjectStoreTest.kt
```diff
@@ -77,19 +88,17 @@ class StyxObjectStoreTest : FeatureSpec() {
 
             scenario("Maintains database integrity in concurrent operations") {
                 val db = StyxObjectStore<String>()
-                val executor = Executors.newFixedThreadPool(4)
+                val executor = Executors.newFixedThreadPool(8)
```
kvosper:
```suggestion
                val executor = newFixedThreadPool(8)
```

components/proxy/src/test/kotlin/com/hotels/styx/routing/db/StyxObjectStoreTest.kt
```diff
@@ -175,6 +180,27 @@ class StyxObjectStoreTest : FeatureSpec() {
...
+            scenario("Maintains database integrity in concurrent operations") {
+                val db = StyxObjectStore<String>()
+
+                // Populate database with data:
+                for (i in 1..10000) {
+                    db.insert("redirect-$i", "Record-$i")
+                }
+
+                // Then remove everyting, concurrently:
+                val executor = Executors.newFixedThreadPool(8)
```
kvosper:
```suggestion
                val executor = newFixedThreadPool(8)
```

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -78,12 +83,16 @@
 
     private final Environment environment;
     private final Configuration configuration;
+    private final RoutingObjectFactory routingObjectFactory;
+    private StyxObjectStore<RoutingObjectRecord> routeDatabase;
 
     private Registry<BackendService> backendServicesRegistry;
 
-    public AdminServerBuilder(Environment environment) {
+    public AdminServerBuilder(Environment environment, StyxServerComponents serverComponents) {
```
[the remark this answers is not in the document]
mikkokar: No. I will fix this.

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -122,6 +131,11 @@ private StandardHttpRouter adminEndpoints(StyxConfig styxConfig) {
         httpRouter.add("/admin/configuration/logging", new LoggingConfigurationHandler(styxConfig.startupConfig().logConfigLocation()));
         httpRouter.add("/admin/configuration/startup", new StartupConfigHandler(styxConfig.startupConfig()));
 
+        RoutingObjectHandler routingObjectHandler = new RoutingObjectHandler(routeDatabase, routingObjectFactory);
+        httpRouter.add("/admin/routing", routingObjectHandler);
```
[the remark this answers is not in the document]
mikkokar: I think this is necessary because `StandardHttpRouter` treats the two endpoints differently. But I'll have another look.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UrlPatternRouter.java
```diff
@@ -0,0 +1,163 @@
...
+        public Builder put(String regexp, HttpHandler handler) {
+            alternatives.add(new RouteDescriptor(HttpMethod.PUT, regexp, handler));
+            return this;
+        }
+
+        public Builder delete(String regexp, HttpHandler handler) {
+            alternatives.add(new RouteDescriptor(HttpMethod.DELETE, regexp, handler));
+            return this;
+        }
+
```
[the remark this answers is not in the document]
mikkokar: It is not necessary for this PR, but surely we can add this when needed.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
mikkokar: Yes I think that should be fine.

mikkokar:
Added functional tests.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UrlPatternRouter.java
```diff
@@ -116,11 +121,12 @@ public UrlPatternRouter build() {
         private final Pattern pattern;
         private final HttpHandler handler;
         private final List<String> placeholderNames;
+        private static final Pattern PLACEHOLDER_PATTERN = Pattern.compile(":([a-zA-Z0-9-_]+)");
 
         public RouteDescriptor(HttpMethod method, String pattern, HttpHandler handler) {
```
dvlato: For all the methods that declare this 'pattern' parameter (we are calling it 'regex' in some invocations as well), shouldn't we call it  "path", "pathExpression", "pathPattern" or something similar instead?

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UrlPatternRouter.java
```diff
@@ -140,15 +146,14 @@ public HttpHandler handler() {
             return placeholderNames;
         }
 
-        private Pattern compilePattern(String pattern) {
-            Pattern x = Pattern.compile(":([a-zA-Z0-9-_]+)");
-            Matcher matcher = x.matcher(pattern);
+        private static Pattern compilePattern(String pattern) {
+            Matcher matcher = PLACEHOLDER_PATTERN.matcher(pattern);
+
             return Pattern.compile(matcher.replaceAll("(?<$1>[a-zA-Z0-9-_]+)"));
```
dvlato: Should we use a name that depicts that we are replacing placeholders here for its appropriate regex? We could also completely  abstract the usage of 'standard Java regex', but I guess that's a bigger change.
Are we documenting the semantics of the UrlPatternRouter anywhere?

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
```diff
@@ -102,8 +102,8 @@ class StyxObjectStore<T> : ObjectStore<T> {
             new = current.minus(key)
         }
 
-        queue {
-            if (current != new) {
+        if (current != new) {
```
dvlato: I'm only seeing the last commit so I haven't reviewed throughly, but I guess you were queueing the comparison to  avoid concurrency issues. Is that not needed?

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UrlPatternRouter.java
```diff
@@ -15,37 +15,42 @@
...
  */
 public class UrlPatternRouter implements HttpHandler {
-    private static final Logger LOGGER = LoggerFactory.getLogger(UrlPatternRouter.class);
+    private static final Logger LOGGER = getLogger(UrlPatternRouter.class);
     private static final String PLACEHOLDERS_KEY = "UrlRouter.placeholders";
     private final List<RouteDescriptor> alternatives;
 
     private UrlPatternRouter(List<RouteDescriptor> alternatives) {
-        this.alternatives = alternatives;
+        this.alternatives = ImmutableList.copyOf(alternatives);
```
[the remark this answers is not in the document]
mikkokar: Normally `ImmutableList.copyOf` has been exempt from static imports.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/RoutingObjectHandler.java
```diff
@@ -81,7 +81,7 @@ public RoutingObjectHandler(StyxObjectStore<RoutingObjectRecord> routeDatabase,
 
                     try {
                         String object = routeDatabase.get(name)
-                                .map(record-> serialise(name, record))
+                                .map(record -> serialise(name, record))
                                 .orElseThrow(ResourceNotFoundException::new);
```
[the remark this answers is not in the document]
mikkokar: No.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UrlPatternRouter.java
mikkokar: I'm calling it `uriPattern` like they do in Play framework. WDYT?

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UrlPatternRouter.java
mikkokar: I'm not documenting this yet. I think it is better to iron out any issues before providing a detailed documentation.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
mikkokar: Now I cannot remember but intuitively it feels safe to do this.

dvlato: ah, sure. I hadn't checked the previous code.  Totally fine.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UrlPatternRouter.java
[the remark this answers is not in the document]
kvosper: To clarify, we try to use static imports when the key information is in the method name, which makes the class name just clutter. In this case, the method name alone is insufficient to demonstrate that we are instantiating an immutable list.

PR #409, opened 2019-05-13
Include server name indication (SNI) in outgoing TLS connections.

dvlato:
Add support for SNI in Styx-client, so the destination ServerName will be sent in outgoing TLS requests as per RFC 6066 section 3 (https://tools.ietf.org/html/rfc6066#section-3).

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/extension/service/TlsSettings.java
+++ b/components/api/src/main/java/com/hotels/styx/api/extension/service/TlsSettings.java
@@ -60,7 +60,7 @@ public class TlsSettings {
         this.protocols = ImmutableList.copyOf(builder.protocols);
         this.cipherSuites = ImmutableList.copyOf(builder.cipherSuites);
         this.sendSni = builder.sendSni;
-        this.sniHost = builder.sniHost;
+        this.sniHost = Optional.ofNullable(builder.sniHost);
     }
 
     private char[] toCharArray(String password) {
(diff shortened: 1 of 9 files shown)


components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
```diff
@@ -52,26 +54,33 @@
...
     }
 
-    private static void addChannelHandlers(Channel channel, HttpConfig httpConfig, SslContext sslContext) {
+    private static void addChannelHandlers(Channel channel, HttpConfig httpConfig, SslContext sslContext, boolean sendSni, String targetHost) {
         ChannelPipeline pipeline = channel.pipeline();
 
         if (sslContext != null) {
-            pipeline.addLast("ssl", sslContext.newHandler(channel.alloc()));
+            SslHandler sslHandler = sendSni
+                    ? sslContext.newHandler(channel.alloc(), targetHost, -1)
```
mikkokar: Is it correct to pass -1 as a peer port? Should we instead pass in the configured port number?

mikkokar: Also, is it necessary to test for `sendSni` flag given that the call site has already set the target accordingly? This is from the `NettyConnection` constructor: 

    sniHost.orElse(origin.host())

components/api/src/main/java/com/hotels/styx/api/extension/service/TlsSettings.java
```diff
@@ -94,6 +99,18 @@ public String trustStorePath() {
...
 
+    public boolean sendSni() {
+        return sendSni;
+    }
+
+    public Optional<String> sniHost() {
+        return sniHost;
+    }
+
+    public  String getSniHost() {
```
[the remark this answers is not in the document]
dvlato: It is used for the serialization to JSON. I think it was implemented the same way in HealthCheckConfig for the URI(). We could also implement a different serialization method for Optionals or remove the optional for these JSON-serializable objects and use nullability instead, but I decided to follow conventions.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnection.java
dvlato: The sendSni flag is to disable sni,  whereas the 'orElse' is to send the target hostname in SNI  instead of a preconfigured override.

It seems that -1 is a valid value as it is the one that will be set in the code with no SNI. The SNI  extension doesn't specify a port number but only a server name.  It's a shame that netty does not have an overload without a port number.

components/api/src/main/java/com/hotels/styx/api/extension/service/TlsSettings.java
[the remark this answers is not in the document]
mikkokar: Okay. Let's put a comment.

PR #412, opened 2019-05-16
Adds a path prefix routing object.

mikkokar:
Adds a new routing object called "PathPrefixRouter". It performs a routing decision based on longest path prefix.

Example usage:

```
    type: PathPrefixRouter
    config:
      routes:
        - { prefix: /, destination: root }
        - { prefix: /foo/, destination: foo }
```

The destination can be also inlined. Like so:

```
    type: PathPrefixRouter
    config:
      routes:
        - { prefix: /, destination: root }
        - prefix: /styx/
          destination:
            type: StaticResponseHandler
            config:
              status: 200
              content: "Hello, from Styx"
```


components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
```diff
@@ -0,0 +1,121 @@
...
+    );
+
+    PathPrefixRouter(List<Pair<String, HttpHandler>> routes) {
+        routes.forEach(entry -> this.routes.put(entry.key(), entry.value()));
+    }
+
+    public Optional<HttpHandler> route(LiveHttpRequest request) {
+        String path = request.path();
+
+        return routes.entrySet().stream()
```
dvlato: Why don't we use a TreeMap for this so we can use subMap()  of floorEntry() ? FloorEntry would solve an issue I see here:
 - If we have two routes such as  '/base/'  and '/base/admin' , we don't know which one is going to match /base/admin/version.txt , do we?

components/proxy/src/main/java/com/hotels/styx/routing/handlers/RouteRefLookup.java
```diff
@@ -0,0 +1,50 @@
...
+
+import static com.hotels.styx.api.HttpResponse.response;
+import static com.hotels.styx.api.HttpResponseStatus.NOT_FOUND;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Resolves a routing object reference from route database.
+ */
+public class RouteRefLookup implements Function<RoutingObjectReference, HttpHandler> {
```
dvlato: I would consider creating a interface for this instead of implementing Function directly. This is not as much to have a slightly easier to read code but basically a solution to overcome the problem that the IDEs 'Find usages' will find any Function instead of only Function<RoutingObjectReference, HttpHandler>

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/PathPrefixRouterTest.kt
```diff
@@ -0,0 +1,256 @@
...
+                    RoutingObjectFactory(RouteRefLookup(objectStore)),
+                    mockk(),
+                    mockk(),
+                    false)
+
+            val handler = PathPrefixRouter.Factory().build(listOf(), factoryContext, routingDef);
+
+            handler.handle(HttpRequest.get("/x").build())
+                    .toMono()
+                    ?.block()
```
kvosper:
```suggestion
                    .block()
```
`toMono` is a kotlin method with a non-nullable return type

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/PathPrefixRouterTest.kt
```diff
@@ -0,0 +1,256 @@
...
+                    ?.bodyAs(UTF_8) shouldBe "root"
+
+            handler.handle(HttpRequest.get("/foo/").build())
+                    .toMono()
+                    ?.block()
```
kvosper:
```suggestion
                    .block()
```

`toMono` is a kotlin method with a non-nullable return type

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/PathPrefixRouterTest.kt
```diff
@@ -0,0 +1,256 @@
...
+                          config:
+                             status: 200
+                             content: hello
+                """.trimIndent())
+
+            val handler = PathPrefixRouter.Factory().build(listOf(), defaultFactoryContext, routingDef);
+
+            handler.handle(HttpRequest.get("/foo").build())
+                    .toMono()
+                    ?.block()
```
kvosper:
```suggestion
                    .block()
```

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/RouteRefLookupTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    "Returns error handler when route object is not found" {
+        val routeDb = mockk<StyxObjectStore<RoutingObjectRecord>>()
+        every { routeDb.get(any()) } returns Optional.empty()
+
+        val response = RouteRefLookup(routeDb).apply(RoutingObjectReference("handler1"))
+                .handle(get("/").build())
+                .toMono()
+                ?.block()
```
kvosper:
```suggestion
                .block()
```

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/PathPrefixRouterTest.kt
mikkokar: Ah okay. Good catch. I'll change them.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/PathPrefixRouterTest.kt
mikkokar: 👍

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
```diff
@@ -0,0 +1,121 @@
...
+                .filter(entry -> path.startsWith(entry.getKey()))
+                .findFirst()
+                .map(Map.Entry::getValue);
+    }
+
+    public static class Factory implements HttpHandlerFactory {
```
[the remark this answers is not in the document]
mikkokar: It is to simplify unit testing.

Otherwise `PathPrefixRouter` would have to be a `HttpHandler` with a return type of `Eventual<LiveHttpResponse>`. This would complicate unit testing.

The `Factory` for every class in `routing.handlers.*` returns a routing object as a `HttpHandler` instance. From the consumer's point of view it doesn't matter if `PathPrefixRouter` is the underlying implementation, or merely just a part of it. But it being merely an internal component of an `HttpHandler` simplifies unit tests and the divergence is preferred.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
mikkokar: This code is mimics the routing logic from `BackendServicesRouter`. It is important to keep both components identical until we deprecate `BackendServicesRouter`.

It is better to minimise the number of variables while migrating backend services router to the new routing object model.

Once the migration is successfully concluded and tested, we shall then optimise this part as a separate increment. Therefore we should leave this as it is. For now.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/RouteRefLookup.java
```diff
@@ -31,20 +29,28 @@
...
+
+        public RouteDbRefLookup(StyxObjectStore<RoutingObjectRecord> routeDatabase) {
+            this.routeDatabase = requireNonNull(routeDatabase);
+        }
+
+        @Override
+        public HttpHandler apply(RoutingObjectReference route) {
+            return this.routeDatabase.get(route.name())
+                    .map(RoutingObjectRecord::getHandler)
+                    .orElse((x, y) -> Eventual.of(response(NOT_FOUND)
```
kvosper: During discussion we noticed that there may be a bug here and in many similar circumstances in which a `LiveHttpRequest` is not having its body consumed.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/RouteRefLookup.java
```diff
@@ -36,21 +36,25 @@
      * A StyxObjectStore based route reference lookup function.
      */
     class RouteDbRefLookup implements RouteRefLookup {
```
kvosper: The indentation below here is incorrect on several lines.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/RouteRefLookup.java
mikkokar: Fixed this handler. However I will review the usage of other handlers separately.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/RouteRefLookup.java
mikkokar: This has been fixed.

PR #414, opened 2019-05-20
Make all origin metrics relative to the scope they are used.

mikkokar:
## Problem Statement

The origin metric names are not relative to the scope they are used. This makes it very inconvenient to re-use code in other context.

For example the connection pool gets to decide a metric name prefix "origins.app.origin.connectionPool". Therefore all connection pool metrics are published with that prefix even if origins, application, or origin have no context in a particular use case. This limits pools usability for other purposes.

## Fix

This PR makes all origin metrics producing components re-usable for other purposes.

The appropriate metrics scope is now injected into all relevant components: 
* Origin Stats Factory
* Connection pool
* Origins inventory (origins status metrics)

### DIFF
--- a/components/client/src/test/unit/java/com/hotels/styx/client/StyxBackendServiceClientTest.java
+++ b/components/client/src/test/unit/java/com/hotels/styx/client/StyxBackendServiceClientTest.java
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2018 Expedia Inc.
+  Copyright (C) 2013-2019 Expedia Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


components/client/src/test/unit/java/com/hotels/styx/client/StyxBackendServiceClientTest.java
```diff
@@ -91,7 +92,9 @@
 
     @BeforeMethod
     public void setUp() {
-        metricRegistry = new CodaHaleMetricRegistry();
+        metricRegistry = new CodaHaleMetricRegistry()
+                .scope("origins")
```
kvosper:
```suggestion
                .scope("origins");
```

components/client/src/test/unit/java/com/hotels/styx/client/StyxBackendServiceClientTest.java
```diff
@@ -91,7 +92,9 @@
...
+        ;
```
kvosper:
```suggestion
        
```

PR #418, opened 2019-05-28
Routing object lifecycle management

mikkokar:
Lifecycle management is needed for cleaning up stateful resources that routing objects may have.

This is necessary for a new `HostProxy` object (under development) for proxying traffic to an individual remote host. It maintains a connection pool which needs shutting after the object is removed.

This PR introduces a new `RoutingObject` interface that extends an `HttpHandler`. It adds a `stop` lifecycle method. Each routing object implements a `RoutingObject` and the `stop` along with it. The default implementation for `stop` lifecycle method is to do nothing.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/routing/config/RoutingObjectFactory.java
+++ b/components/proxy/src/main/java/com/hotels/styx/routing/config/RoutingObjectFactory.java
@@ -20,12 +20,13 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.hotels.styx.Environment;
 import com.hotels.styx.api.Eventual;
-import com.hotels.styx.api.HttpHandler;
 import com.hotels.styx.config.schema.Schema;
 import com.hotels.styx.proxy.plugin.NamedPlugin;
+import com.hotels.styx.routing.RoutingObject;
 import com.hotels.styx.routing.RoutingObjectRecord;
 import com.hotels.styx.routing.db.StyxObjectStore;
 import com.hotels.styx.routing.handlers.ConditionRouter;
+import com.hotels.styx.routing.handlers.HostProxy;
 import com.hotels.styx.routing.handlers.HttpInterceptorPipeline;
 import com.hotels.styx.routing.handlers.PathPrefixRouter;
 import com.hotels.styx.routing.handlers.ProxyToBackend;
@@ -58,6 +59,7 @@ public class RoutingObjectFactory {
     private static final String INTERCEPTOR_PIPELINE = "InterceptorPipeline";
     private static final String PROXY_TO_BACKEND = "ProxyToBackend";
     private static final String PATH_PREFIX_ROUTER = "PathPrefixRouter";
+    private static final String HOST_PROXY = "HostProxy";
 
 
     static {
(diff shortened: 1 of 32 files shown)


components/proxy/src/main/java/com/hotels/styx/routing/config/RoutingObjectFactory.java
```diff
@@ -58,6 +58,7 @@
     private static final String INTERCEPTOR_PIPELINE = "InterceptorPipeline";
     private static final String PROXY_TO_BACKEND = "ProxyToBackend";
     private static final String PATH_PREFIX_ROUTER = "PathPrefixRouter";
+    private static final String HOST_PROXY = "HostProxy";
```
kvosper: It looks like this new field is not used?

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HttpInterceptorPipeline.java
```diff
@@ -36,40 +36,47 @@
...
 
-    private final StandardHttpPipeline handler;
+    private final RoutingObject handler;
+    private final StandardHttpPipeline pipeline;
 
-    public HttpInterceptorPipeline(List<HttpInterceptor> interceptors, HttpHandler handler, boolean trackRequests) {
+    public HttpInterceptorPipeline(List<HttpInterceptor> interceptors, RoutingObject handler, boolean trackRequests) {
         RequestTracker tracker = trackRequests ? CurrentRequestTracker.INSTANCE : RequestTracker.NO_OP;
-        this.handler = new StandardHttpPipeline(interceptors, handler, tracker);
+        this.handler = handler;
```
kvosper:
```suggestion
        this.handler = requireNonNull(handler);
```

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
```diff
@@ -93,9 +102,20 @@ public HttpHandler build(List<String> parents, Context context, RoutingObjectDef
...
+                public Eventual<LiveHttpResponse> handle(LiveHttpRequest request, HttpInterceptor.Context context) {
+                    return pathPrefixRouter.route(request)
                             .orElse((x, y) -> Eventual.error(new NoServiceConfiguredException(request.path())))
-                            .handle(request, ctx);
+                            .handle(request, context);
+                }
+
+                @Override
+                public CompletableFuture<Void> stop() {
+                    pathPrefixRouter.routes.forEach((route, value) -> value.stop());
```
kvosper:
```suggestion
                    pathPrefixRouter.routes.forEach((route, routingObject) -> routingObject.stop());
```

components/proxy/src/main/java/com/hotels/styx/routing/handlers/RouteRefLookup.java
```diff
@@ -30,7 +30,10 @@
  * Resolves a routing object reference from route database.
  */
 public interface RouteRefLookup {
-    HttpHandler apply(RoutingObjectReference route);
+    // Consider modifying this interface to return Optional<RoutingObject>.
```
kvosper: Is this comment still relevant?

mikkokar: Yeah. It was originally a TODO, but then decided to leave it as a normal comment. I'm bit indecisive at the moment.

PR #420, opened 2019-05-31
Remove dependency between server module and client module

### LINKED ISSUE #375: Can dependency between server and client can be removed
## Background
Currently styx-server depends on styx-client as a maven dependency. There are only few exception classes being used by styx-server, so maybe there is an opportunity there to reduce dependance.

## Acceptance criteria
* Check if the dependency can be removed,
* Propose a solution how to handle the exception classes used by styx-server from styx-client module,
* Remove the dependency if feasable.

---

One suggestion is to:

> Add an interceptor to map client specific exceptions to common styx exceptions.
> Remove the client specific exception mapping from the Styx HTTP Pipeline, and map styx common exceptions to specific response codes.
> Then we can remove the dependency between Styx Server and Client.
> However, as HttpPipelineHandler is part of Styx-Server, still needs to know about proxy related response codes, such as Gateway Timeout, etc. But then again HttpPipelineHandler already performs proxying related such as flow control.

> Ideally though HttpPipelineHandler as a server component would not have any proxying related logging in it. Instead the proxying related logic would all go to styx-proxy alone. Perhaps we should explore this split bit further?

kvosper:
Fixes #375 .


components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpPipelineHandler.java
```diff
@@ -520,8 +494,8 @@ private LiveHttpResponse exceptionToResponse(Throwable exception, LiveHttpReques
                 .build();
     }
 
-    private static HttpResponseStatus status(Throwable exception) {
-        return EXCEPTION_STATUSES.statusFor(exception)
+    private HttpResponseStatus status(Throwable exception) {
+        return exceptionStatuses.statusFor(exception)
```
mikkokar: I wouldn't 't bother doing this. I think it is better to remove exception mapping from this class altogether.

I would move the exception mapping out into the HTTP pipeline. We can then assume the pipeline maps the exceptions as required. We can then map any uncaught exceptions always to 500 internal server errors here.

kvosper: OK, I will investigate this.

mikkokar: My reasoning is that we can cut down dependencies (in wiring) between different the Styx classes. This somewhat roundabout way of keeping exception mapping in this class adds extra code in `ServerConnectorFactory` (in proxy) and in `HttpPipelineHandler` (in server).

We can get rid of this by mapping the exceptions in a separate interceptor that can be easily implemented and tested as a single unit. Also this HttpPipelineHandler gets bit simpler as all exceptions would always map to 500 internal server errors. This should simplify the unit tests too.

kvosper:
As this task is more complicated than it seems, and does not add much value, we are going to close without merging and maybe revisit it later.

PR #421, opened 2019-06-03
Fix client thread leak after origin reload.

mikkokar:
This pull request addresses two issues:

### A thread leak associated with origin reloads. 

The Styx `proxy.clientWorkerThreadsCount` setting was not honoured. Styx is supposed to have only one client thread pool that is shared among all backend services and `proxy.clientWorkderThradsCount` is the number of threads in this pool. 

Due to an error in styx code, this pool was instantiated once per each Styx backend service, resulting in a high number of client threads. More than configuration allowed.

Secondly, this resulted a thread leak. When new backend services were added, additional thread pools were being created. Because the thread pool was supposedly shared, it was never closed when a backend service was removed.

This PR ensures that the client thread pool is shared, and not separately created for each new backend service. I am also adding an integration test to ensure this leak won't happen in future.

### A "global" event loop for StyxHttpClient. 

This PR also adds a "global" event loop for `StyxHttpClient`. It lives in a static context processes all IO events for `StyxHttpClient`. Benefit is that a client is always ready to be used and doesn't need to be closed afterward. For example:

```
val client = StyxHttpClient.Builder().build()

client.send(get("/")...build())
    .wait()
    .status() shouldBe OK

# When using the global event loop it is no longer necessary to close
# the client afterwards. Hence the following line can be removed:
client.close()
```

As styx health check functionality uses this client, the health check IO events also go on to this global event loop.

The global pool size is derived from the CPU core count. That is 0.5xCPUs.


components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnectionFactory.java
```diff
@@ -93,6 +103,15 @@ private NettyConnectionFactory(Builder builder) {
         });
     }
 
+    @Override
+    public CompletableFuture<Void> close() {
+        if (eventLoopGroup == GLOBAL_CLIENT_EVENT_LOOP) {
+            return completedFuture(null);
+        }
+
+        return fromNettyFuture(eventLoopGroup.shutdownGracefully());
```
dvlato: I don't know if I understand this code correctly: if the eventLoopGroup has been injected externally (and thus could be potentially shared among several different connection factories) we close it here.  That doesn't seem correct. 
Also we are initializing a  Default Global Event Loop that's not going to be used in most cases.

components/proxy/src/main/java/com/hotels/styx/StyxPipelineFactory.java
```diff
@@ -87,7 +90,7 @@ private RoutingObject configuredPipeline(RoutingObjectFactory routingObjectFacto
                 })
                 .orElseGet(() -> {
                     Registry<BackendService> backendServicesRegistry = (Registry<BackendService>) services.get("backendServiceRegistry");
-                    return new StaticPipelineFactory(environment, backendServicesRegistry, plugins, requestTracking);
+                    return new StaticPipelineFactory(environment, backendServicesRegistry, plugins, eventLoopGroupFactory, requestTracking);
```
dvlato: It seems we are only passing along the configured eventLoopGroupFactory for 'static pipelines'  (I haven't really checked the other code path, maybe I'm missing something)

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnectionFactory.java
mikkokar: Your criticism is justified. Let me see if event loop can be closed outside.

components/proxy/src/main/java/com/hotels/styx/StyxPipelineFactory.java
mikkokar: Static pipeline is more important because it implements the documented styx functionality. 

All other uses are for routing objects that are more or less experimental at the moment, and we'd need to think about how to go forward with them in near future. I see that as a separate increment. I suspect once the routing object model is finalised in styx 1.1 we'll end up deprecating the `ProxyToBackend` and `BackendServiceProxy` objects as unnecessary.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnectionFactory.java
mikkokar: Reading your comment again, and looking at code, I notice that `eventLoopGroup` is not really injected. It is created with the factory that has been externally injected.

But with memoization it may have the effect you describe: The same `eventLoopGroup` could end up being shared with different connection factories, and one of the factories may end up closing it on behalf of others.

This feels an incorrect usage of `NettyConnectionFactory` rather than an error in the `NettyConnectionFactory` itself.

dvlato: Oh, sure, that's right. In this case we might be safe ourselves as we don't try to close the connectionFactory except for healthchecks  - and the StyxHttpClient used in healthchecks is using a different eventLoopGroup than the forwarding path...  But it seems it could lead to some issues.

components/proxy/src/test/java/com/hotels/styx/routing/StaticPipelineBuilderTest.java
```diff
@@ -74,7 +75,7 @@ public void appliesPluginsInOrderTheyAreConfigured() throws Exception {
                 interceptor("Test-B", appendResponseHeader("X-From-Plugin", "B"))
         );
 
-        HttpHandler handler = new StaticPipelineFactory(clientFactory, environment, registry, plugins, false).build();
+        HttpHandler handler = new StaticPipelineFactory(clientFactory, environment, registry, plugins, factory,false).build();
```
kvosper:
```suggestion
        HttpHandler handler = new StaticPipelineFactory(clientFactory, environment, registry, plugins, factory, false).build();
```

components/proxy/src/test/java/com/hotels/styx/routing/StaticPipelineBuilderTest.java
```diff
@@ -61,8 +63,7 @@ public void staticPipelineBuilderTest() {
 
     @Test
     public void buildsInterceptorPipelineForBackendServices() throws Exception {
-
-        HttpHandler handler = new StaticPipelineFactory(clientFactory, environment, registry, ImmutableList.of(), false).build();
+        HttpHandler handler = new StaticPipelineFactory(clientFactory, environment, registry, ImmutableList.of(), factory,false).build();
```
kvosper:
```suggestion
        HttpHandler handler = new StaticPipelineFactory(clientFactory, environment, registry, ImmutableList.of(), factory, false).build();
```

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/admin/FileBasedOriginsFileChangeMonitorSpec.kt
```diff
@@ -0,0 +1,130 @@
...
+
+            val reqToApp02 = HttpRequest.get("/app02/x")
+                    .header(HOST, styxServer().proxyHttpHostHeader())
+                    .build()
+
+            client.send(reqToApp01).wait().status() shouldBe OK
+            client.send(reqToApp02).wait().status() shouldBe BAD_GATEWAY
+
+            writeConfig(styxOriginsFile, configTemplate.format("appv2", "/app02/", mockServer.port()))
+            Thread.sleep(2000)
```
kvosper: We should use `io.kotlintest.eventually` instead of a long sleep.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/resiliency/OriginResourcesSpec.kt
```diff
@@ -0,0 +1,161 @@
...
+
+            val threadCountBefore = threadCount("Styx-Client-Worker")
+
+            "---" + (1..count)
+                    .map { appDeclaration("bbb-$it") }
+                    .joinToString("\n")
+                    .let { writeConfig(styxOriginsFile, it) }
+
+            // Wait for styx to pick up the new configuration
+            Thread.sleep(2000)
```
kvosper: We should use `io.kotlintest.eventually` instead of a long sleep.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/resiliency/OriginResourcesSpec.kt
```diff
@@ -0,0 +1,161 @@
...
+
+    init {
+        "Cleans up threads after use" {
+            "---" + (1..count)
+                    .map { appDeclaration("aaa-$it") }
+                    .joinToString("\n")
+                    .let { writeConfig(styxOriginsFile, it) }
+
+            // Wait for styx to pick up the configuration
+            Thread.sleep(2000)
```
kvosper: We should use `io.kotlintest.eventually` instead of a long sleep.

components/proxy/src/test/java/com/hotels/styx/routing/StaticPipelineBuilderTest.java
mikkokar: IntellJ argument name view tends to hide this kind of mistakes 🙄

kvosper: It's a very small thing, but it will probably be considered a checkstyle violation.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnectionFactory.java
```diff
@@ -60,14 +64,12 @@
     private Class<? extends Channel> clientSocketChannelClass;
 
     private NettyConnectionFactory(Builder builder) {
-        PlatformAwareClientEventLoopGroupFactory eventLoopGroupFactory = new PlatformAwareClientEventLoopGroupFactory(
-                builder.name,
-                builder.clientWorkerThreadsCount
-        );
-        this.eventLoopGroup = eventLoopGroupFactory.newClientWorkerEventLoopGroup();
+        this.clientSocketChannelClass = builder.channelClass;
+
```
kvosper: No need for blank lines between assignments here.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnectionFactory.java
```diff
@@ -129,30 +127,14 @@ protected void initChannel(Channel ch) {
...
-        /**
-         * Sets number of client worker threads.
-         *
-         * @param clientWorkerThreadsCount number of client worker threads
-         * @return this builder
-         */
-        public Builder clientWorkerThreadsCount(int clientWorkerThreadsCount) {
-            this.clientWorkerThreadsCount = clientWorkerThreadsCount;
+        public Builder nettyEventLoop(EventLoopGroup eventLoopGroup, Class<? extends Channel> channelClass) {
+            this.eventLoopGroup = eventLoopGroup;
```
kvosper:
```suggestion
            this.eventLoopGroup = requireNonNull(eventLoopGroup);
```

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnectionFactory.java
```diff
@@ -129,30 +127,14 @@ protected void initChannel(Channel ch) {
...
+            this.channelClass = channelClass;
```
kvosper:
```suggestion
            this.channelClass = requireNonNull(channelClass);
```

components/proxy/src/main/java/com/hotels/styx/proxy/BackendServicesRouter.java
```diff
@@ -115,7 +124,9 @@ public void onChange(Registry.Changes<BackendService> changes) {
             ConnectionPoolSettings poolSettings = backendService.connectionPoolConfig();
 
             Connection.Factory connectionFactory = connectionFactory(
-                    backendService,
+                    nettyEventLoopGroup, socketChannelClass,
```
kvosper:
```suggestion
                    nettyEventLoopGroup, 
                    socketChannelClass,
```

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyConnectionFactory.java
mikkokar: Please check this commit: a5d70cd123635ce59fde8acea9f28bc460fe92b8.

It is a different approach and removes the `close` method. 

Now the Netty `eventLoopGroup` and accompanying channel type are injected from outside. They are never closed. The client event loop group is created when styx starts and stays alive until shut.

The Styx health checks & Styx standalone client are now permanently fixed on "global" event loop group without option to inject its own. This can be improved if necessary.

components/proxy/src/test/java/com/hotels/styx/proxy/BackendServicesRouterTest.java
```diff
@@ -63,6 +69,15 @@
 
     private Environment environment;
 
+    private ClientEventLoopFactory factory = new PlatformAwareClientEventLoopGroupFactory("x", 0);
+    private EventLoopGroup eventLoopGroup = this.factory.newClientWorkerEventLoopGroup();
+    private Class<? extends SocketChannel> channelClass = this.factory.clientSocketChannelClass();
+
+    @AbstractAnnotationSpec.AfterAll
```
kvosper: This is a java test, not a kotlin test.
```suggestion
    @AfterClass
```

components/proxy/src/test/java/com/hotels/styx/routing/StaticPipelineBuilderTest.java
```diff
@@ -74,7 +80,7 @@ public void appliesPluginsInOrderTheyAreConfigured() throws Exception {
                 interceptor("Test-B", appendResponseHeader("X-From-Plugin", "B"))
         );
 
-        HttpHandler handler = new StaticPipelineFactory(clientFactory, environment, registry, plugins, false).build();
+        HttpHandler handler = new StaticPipelineFactory(clientFactory, environment, registry, plugins, eventLoopGroup, socketChannelClass,false).build();
```
kvosper:
```suggestion
        HttpHandler handler = new StaticPipelineFactory(clientFactory, environment, registry, plugins, eventLoopGroup, socketChannelClass, false).build();
```
Missing space is probably a checkstyle violation.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/resiliency/OriginResourcesSpec.kt
mikkokar: Not sure if I'd use here. The following step is not really a verification step, but rather a step to ensure that new client worker threads are being created.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/resiliency/OriginResourcesSpec.kt
mikkokar: See above.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/resiliency/OriginResourcesSpec.kt
kvosper: That's not the point.
1. Eventually can be given a short interval (e.g. 100 ms) will exit once the job is done, sleep(2000) will never spend less than 2 seconds. The more we use sleeps, the more we unnecessarily extend the duration of the build, which we do not want.
2. Sleep will exit after 2 seconds even if the job is not done. This makes the test behaviour nondeterministic. If some other process on the machine causes a delay, the test will not work properly and the whole build has to be run again.

mikkokar: I know the benefits of `eventual` @kvosper. I just wanted to keep the code less cluttered. But I'll think about something.

mikkokar: Okay. I did some improvements on this test. Including eventual. Have another look please.

PR #422, opened 2019-06-04
Enable config validator by default

kvosper:


components/proxy/src/main/java/com/hotels/styx/proxy/InterceptorPipelineBuilder.java
```diff
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2018 Expedia Inc.
+  Copyright (C) 2013-2019 Expedia Inc.
```
mikkokar: I would back these changes off.

PR #423, opened 2019-06-04
Host proxy object

mikkokar:
Adds a new routing object: HostProxy. It proxies traffic to single configured endpoint.

It supports:
* Styx connection pool (with expiring connections)
* TLS settings
* Response timeout configuration
* Origin/Application and connection pool metrics.
* A metric prefix for backwards compatibility

The following "features" are omitted from this pull request:
* Request logging at styx client level. This shouldn't be necessary as logging could be implemented as a separate routing object.
* Origin/Application metrics are not de-registered after the host proxy object is removed. Styx application routing at present doesn't de-register them anyway, and therefore we'd have to implement this as a separate increment.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
+++ b/components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
@@ -1,0 +1,238 @@
+/*
+  Copyright (C) 2013-2019 Expedia Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package com.hotels.styx.routing.handlers;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.net.HostAndPort;
+import com.hotels.styx.api.Eventual;
+import com.hotels.styx.api.HttpInterceptor;
+import com.hotels.styx.api.LiveHttpRequest;
+import com.hotels.styx.api.LiveHttpResponse;
+import com.hotels.styx.api.extension.Origin;
+import com.hotels.styx.api.extension.service.ConnectionPoolSettings;
+import com.hotels.styx.api.extension.service.TlsSettings;
+import com.hotels.styx.client.Connection;
+import com.hotels.styx.client.OriginStatsFactory;
+import com.hotels.styx.client.StyxHostHttpClient;
+import com.hotels.styx.client.applications.metrics.OriginMetrics;
+import com.hotels.styx.client.connectionpool.ConnectionPool;
+import com.hotels.styx.client.connectionpool.ExpiringConnectionFactory;
+import com.hotels.styx.client.connectionpool.SimpleConnectionPoolFactory;
+import com.hotels.styx.client.netty.connectionpool.NettyConnectionFactory;
+import com.hotels.styx.config.schema.Schema;
+import com.hotels.styx.infrastructure.configuration.yaml.JsonNodeConfig;
+import com.hotels.styx.routing.RoutingObject;
+import com.hotels.styx.routing.config.HttpHandlerFactory;
+import com.hotels.styx.routing.config.RoutingObjectDefinition;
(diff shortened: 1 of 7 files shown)


components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/HostProxyTest.kt
```diff
@@ -0,0 +1,131 @@
...
+                }
+
+                verify {
+                    client?.sendRequest(ofType(LiveHttpRequest::class))
+                }
+            }
+
+            scenario("Requests arriving at stopped HostProxy object") {
+                val exception = HostProxy(HostAndPort.fromString("localhost:80"), client).let {
+                    it.stop()
```
kvosper: Should our test know that `stop` is synchronous, despite returning a `CompleteableFuture`? If it becomes async at any point in the future this test will start failing.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/HostProxyTest.kt
```diff
@@ -0,0 +1,131 @@
...
+            scenario("Port defaults to 80") {
+                val factory = HostProxy.Factory()
+
+                val hostProxy = factory.build(listOf(), context.get(), routingObjectDef("""
+                          type: HostProxy
+                          config:
+                            host: localhost
+                        """.trimIndent())) as HostProxy
+
+                hostProxy.hostAndPort.let {
```
dvlato: let() here doesn't seem extremely useful as we are not reading the returned value (e.g, for chaining) or redefining the name of "it" . If we want to reduce verbosity a little bit, we could use run()  for these cases - or apply() if we want to return the receiver object.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
```diff
@@ -0,0 +1,238 @@
...
+                            "connectTimeoutMillis",
+                            "socketTimeoutMillis",
+                            "pendingConnectionTimeoutMillis",
+                            "connectionExpirationSeconds")
+            )),
+            optional("responseTimeoutMillis", integer()),
+            optional("metricPrefix", string())
+    );
+
+    private final String message;
```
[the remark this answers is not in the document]
mikkokar: This class is just 10 lines of code (excluding schema & other boiler plate). Don't you think `unexpectedTrafficErrorMessage` is bit over the top in that context?

How about just `errorMessage`?

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/HostProxyTest.kt
mikkokar: Not at the moment. We'll fix this when it starts failing.

I have a separate work item in the pipeline to fix it. But that goes beyond scope of this PR.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/HostProxyTest.kt
mikkokar: Fair point. Although I'd prefer `let` because it is also familiar from other languages and therefore bit more self explanatory. Also, without `it` the `run` block might appear bit magical, but also cleaner:

                hostProxy.hostAndPort.run {
                    hostText shouldBe "localhost"
                    port shouldBe 443
                }

I'm happy either way. 

What do you think?

dvlato: I would typically prefer let but in these simple cases run or even apply (I think "run" sounds a bit more obscure )  look neater; however  it adds some complexity in the writing/reading process so it might be easier if we just stick to let{} unless we can't use it.

kvosper: I think it doesn't make a huge difference. I see it's been changed to `run`. I'm fine with that.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/HostProxyTest.kt
kvosper: I think my comment was unclear. What I was saying is that instead of calling `stop()` here, we should call something like `stop().get()`, since we should assume that a method with an asynchronous signature will execute asynchronously, even if we know that this particular case doesn't.

mikkokar: Actually no. The handler should start rejecting requests right after a call to `stop`, before the future has completed, and the handler has fully come to stop. The existing requests ought to drain gracefully. But that starts going out of scope and therefore is a separate enhancement.

PR #426, opened 2019-06-07
Update config documentation

kvosper:


docs/user-guide/configure-overview.md
```diff
@@ -12,192 +12,161 @@ using environment variables with the same name as the property.
...
+      sessionTimeoutMillis: 300000
+      sessionCacheSize: 20000
+      protocols:
+        - TLSv1.2
+      cipherSuites:
+       - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
+       - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+       - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+       - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+  #  0 -> availableProcessors / 2 threads will be used
```
dvlato: Do we want to explain what this value is for?  Also is this 0 ->  comment true or is that the case only when bossThreadsCount is null?

docs/user-guide/configure-overview.md
```diff
@@ -12,192 +12,161 @@ using environment variables with the same name as the property.
...
+    http:
+      # Server port for Styx admin interface.
+      port: 9000
+  # Number of threads for handling incoming connections on admin interface. 0 -> availableProcessors / 2 threads will be used.
+  bossThreadsCount: 1
+  # Number of worker threads for admin interface
+  # Worker threads are those performing all the asynchronous I/O operation on the inbound channel.
+  # 0 -> availableProcessors / 2 threads will be used
+  workerThreadsCount: 2
+  tcpNoDelay: true
```
dvlato: Are these values explained anywhere?

docs/user-guide/configure-overview.md
```diff
@@ -12,192 +12,161 @@ using environment variables with the same name as the property.
...
-        name: "X-Styx-Request-Id"
-        
-    # Enables request tracking. This is a debugging feature that shows information about
-    # each proxied request. Accepts a boolean value (true/false).
-    requestTracking: false
+# Enables request tracking. This is a debugging feature that shows information about
+# each proxied request. Accepts a boolean value (true/false).
+requestTracking: false
+
+# Allow and Encode the list of unwise chars.
```
dvlato: Can we maybe give an explanation for this one?

docs/user-guide/configure-load-balancing.md
```diff
@@ -2,6 +2,7 @@
 
 Styx supports three load balancing strategies:
 
+ - Power of Two
  - Round robin
```
dvlato: I think this is usually spelt 'round-robin', with a hyphen. In that case we should probably change all the later occurrences (like in the header).

dvlato:
```suggestion
 - Round-robin
```

docs/user-guide/configure-load-balancing.md
```diff
@@ -2,6 +2,7 @@
 
 Styx supports three load balancing strategies:
 
+ - Power of Two
```
dvlato:
```suggestion
 - Power of two
```

docs/user-guide/configure-overview.md
```diff
@@ -12,192 +12,161 @@ using environment variables with the same name as the property.
...
-        enabled: true
-        expirationMillis: 10000
-
-    loadBalancing:
-      strategy: #Check load balancing documentation for all the possible strategies
-        factory: {class: "com.hotels.styx.client.loadbalancing.strategies.PowerOfTwoStrategy$Factory"}
- 
-
-
-      adaptive:
```
dvlato: I see some config examples that still use the adaptive strategy (like styx-config-test.yml or styx-config-alternative.yml).  Can you please replace those references as well?

docs/user-guide/configure-overview.md
kvosper: This chunk of config is actually not new, but the same as before, only Git diff thinks it is new. Still, we can benefit from adding some explanation.

dvlato: Yes, I noticed it was not new but it seems something we could consider 'bringing the documentation up-to-date'.

docs/user-guide/configure-overview.md
```diff
@@ -12,192 +12,167 @@ using environment variables with the same name as the property.
...
+      protocols:
+        - TLSv1.2
+      # The cipher suites to enable, in the order of preference.
+      cipherSuites:
+       - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
+       - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+       - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+       - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+  # Boss threads are those establishing new channels.
+  #  If set to 0, availableProcessors / 2 threads will be used
```
dvlato: I think that if it's set to 0, it will use Netty's default which is availableProcessors * 2  ( https://github.com/netty/netty/blob/00afb19d7a37de21b35ce4f6cb3fa7f74809f2ab/transport/src/main/java/io/netty/channel/MultithreadEventLoopGroup.java#L40 ) .

docs/user-guide/configure-overview.md
```diff
@@ -12,192 +12,167 @@ using environment variables with the same name as the property.
...
+  bossThreadsCount: 1
+  # styx client worker threads are those performing all the asynchronous I/O operation to the backend origins.
+  # If set to 0, availableProcessors / 2 threads will be used
+  clientWorkerThreadsCount: 0
+  # Worker threads are those performing all the asynchronous I/O operation on the inbound channel.
+  # If set to 0, availableProcessors / 2 threads will be used
```
dvlato: Same comment as above.

docs/user-guide/configure-overview.md
kvosper: We do not leave the defaults up to Netty. However, I have discovered that the default varies depending on whether this property is not set, or whether the proxy config is entirely absent.

docs/user-guide/configure-overview.md
kvosper: For worker threads, Styx will replace 0 with availableProcessors / 2

docs/user-guide/configure-overview.md
kvosper: See https://github.com/HotelsDotCom/styx/issues/427

docs/user-guide/configure-overview.md
```diff
@@ -12,192 +12,166 @@ using environment variables with the same name as the property.
...
+       - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+       - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+       - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+  # Boss threads are those establishing new channels.
+  bossThreadsCount: 1
+  # Styx client worker threads are those performing all the asynchronous I/O operation to the backend origins.
+  # If set to 0, availableProcessors / 2 threads will be used
+  clientWorkerThreadsCount: 0
+  # Worker threads are those performing all the asynchronous I/O operation on the inbound channel.
+  # If set to 0, availableProcessors / 2 threads will be used
```
dvlato: I think I would still remove this "If set to 0", as IIRC this was not the case here. Then let's change the example to something different from 0!

components/server/src/main/java/com/hotels/styx/server/netty/NettyServerConfig.java
```diff
@@ -125,42 +135,84 @@ public int nioAcceptorBacklog() {
...
 
+    /*
+     TODO unused: https://github.com/HotelsDotCom/styx/issues/428
+     */
     public int maxContentLength() {
         return this.maxContentLength;
     }
 
+    /**
+     * Time in milliseconds Styx Proxy Service waits for an incoming request from client.
```
mikkokar: This description is bit vague. 

It would be worth clarifying that this parameter controls the amount of tolerated inactivity while the request is being received. If the client started sending the request, and then suddenly cuts it off, styx would detect this after `requestTimeoutMillis`.

PR #429, opened 2019-06-11
Load Balancing Group routing object

mikkokar:
Adds a `LoadBalancingGroup` routing object that balances load between other routing objects. Specifically it can load balance between `HostProxy` objects, and can be used with `PathPrefixRouter` to perform styx application routing. Currently, only `Pow2` load balancing algorithm is supported.


components/proxy/src/main/java/com/hotels/styx/routing/RoutingObjectAdapter.java
```diff
@@ -0,0 +1,67 @@
...
+    private final RoutingObject delegate;
+
+    private final LongAdder allRequests = new LongAdder();
+    private final LongAdder finishedRequests = new LongAdder();
+
+    /**
+     * Routing object adapater constructor.
+     * @param routingObject
+     */
+    public RoutingObjectAdapter(RoutingObject routingObject) {
```
dvlato: Should we use Kotlin here to use implicit delegation?

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
```diff
@@ -0,0 +1,244 @@
...
+                    optional("sslProvider", string()),
+                    optional("trustStorePath", string()),
+                    optional("trustStorePassword", string()),
+                    optional("protocols", list(string())),
+                    optional("cipherSuites", list(string())),
+                    optional("additionalCerts", list(object(
+                            field("alias", string()),
+                            field("certificatePath", string())
+                    ))),
+                    atLeastOne("trustAllCerts",
```
dvlato: is this validation correct? It seems that "protocols" and "cipherSuites" are not related with trustAllCerts/trustStorePath/additionalCerts and, of course, trustStorePassword could not be the "only one" here.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/LoadBalancingGroup.java
```diff
@@ -0,0 +1,193 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(LoadBalancingGroup.class);
+
+    private StyxObjectStore<RoutingObjectRecord> routeDatabase;
+    private StyxBackendServiceClient client;
+    private Disposable changeWatcher;
+
+
+    public LoadBalancingGroup(StyxBackendServiceClient client, Disposable changeWatcher) {
+        this.client = requireNonNull(client);
```
dvlato: Are these fields final?

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
```diff
@@ -107,8 +108,8 @@ private StyxServerComponents(Builder builder) {
                 .map(StyxServerComponents::readHttpHandlers)
                 .orElse(ImmutableMap.of())
                 .forEach((name, record) -> {
-                    RoutingObject handler = routingObjectFactory.build(ImmutableList.of(name), record);
-                    routeObjectStore.insert(name, new RoutingObjectRecord(record.type(), record.config(), handler))
+                    RoutingObjectAdapter adapter = new RoutingObjectAdapter(routingObjectFactory.build(ImmutableList.of(name), record));
```
dvlato: is adapter a good name here?

components/api/src/main/java/com/hotels/styx/api/configuration/ObjectStore.java
```diff
@@ -26,10 +28,17 @@
...
      * @return an object if known
      */
     Optional<T> get(String key);
+
+    /**
+     * Returns a list of full contents.
+     *
+     * @return a collection of all entries.
+     */
+    Collection<Map.Entry<String, T>> entrySet();
```
kvosper: Should we consider using `com.hotels.styx.common.Pair` instead of `Map.Entry`?

components/common/src/main/java/com/hotels/styx/api/ResponseEventListener.java
```diff
@@ -99,6 +106,11 @@ public ResponseEventListener whenCompleted(Runnable action) {
         return this;
     }
 
+    public ResponseEventListener whenFinished(Runnable action) {
```
kvosper: Needs documentation.

components/proxy/src/main/java/com/hotels/styx/routing/RoutingObjectAdapter.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.LongAdder;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Adapts routing object into Styx core.
+ * Provides a load balancing metric for the adapted routing object.
+ */
+public class RoutingObjectAdapter implements RoutingObject {
```
kvosper: This is called "Adapter" but the interface is the same as the delegate's interface. It seems that what it actually does is not adapting interfaces, but rather wrapping to add metrics.

Perhaps `RequestCountingRoutingObject`, `InstrumentedRoutingObject` or something like that might be a better name?

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
```diff
@@ -0,0 +1,244 @@
...
+import static com.hotels.styx.routing.config.RoutingSupport.missingAttributeError;
+import static java.lang.String.format;
+import static java.lang.String.join;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.CompletableFuture.completedFuture;
+
+/**
+ * A routing object that proxies all incoming traffic to a remote host.
+ */
+public class HostProxy implements RoutingObject {
```
kvosper: This diff seems to think `HostProxy` is a new class, when it was actually added in #423. I think you may need to rebase from `master`. I will continue reviewing after.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
mikkokar: Hi @dvlato could you clarify this please?

>  It seems that "protocols" and "cipherSuites" are not related with trustAllCerts/trustStorePath/additionalCerts and,

This schema models the existing configuration format using the rudimentary validator that is available at the moment.

For example, there is no way to tell the validator to fail the configuration if `trustStorePassword` appears without `trustStorePath`.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
```diff
@@ -0,0 +1,244 @@
...
+                            "connectTimeoutMillis",
+                            "socketTimeoutMillis",
+                            "pendingConnectionTimeoutMillis",
+                            "connectionExpirationSeconds")
+            )),
+            optional("responseTimeoutMillis", integer()),
+            optional("metricPrefix", string())
+    );
+
+    private final String message;
```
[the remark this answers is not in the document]
mikkokar: Apologies, this PR builds on another (HostProxy) PR. I should have postponed publishing this PR until HostProxy changes have been merged. 

Regardless, it they have been merged, and the `message` did gain a more meaningful name.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
```diff
@@ -0,0 +1,244 @@
...
+
+            HostAndPort hostAndPort = config.get("host")
+                    .map(HostAndPort::fromString)
+                    .map(it -> addDefaultPort(it, tlsSettings))
+                    .orElseThrow(() -> missingAttributeError(configBlock, join(".", parents), "host"));
+
+            return createHostProxyHandler(context, hostAndPort, poolSettings, tlsSettings, responseTimeoutMillis, metricPrefix);
+        }
+
+        private static HostAndPort addDefaultPort(HostAndPort hostAndPort, TlsSettings tlsSettings) {
```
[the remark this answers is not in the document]
mikkokar: Good point! I was unaware of this. Thanks for pointing out.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/LoadBalancingGroup.java
mikkokar: This change is outdated.

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
mikkokar: It adapts a `RoutingObject` instance to Styx core, adding some functionality to it (load balancing metric).

I tried to google for some industry standard names for this use case but drew a blank. I'm open for any sensible proposals.

components/proxy/src/main/java/com/hotels/styx/routing/RoutingObjectAdapter.java
mikkokar: See my response to @dvlato above.

It "adapts" a bare RoutingObject into styx core. It adds any functionality or metadata that may be necessary for styx to interoperate and use the routing object.

Let's add additional comments to the other, above thread.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
mikkokar: Correct.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
dvlato: As discussed offline this can be ignored.

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
dvlato: So I would say that in default OO patterns this is not an adapter but a Decorator.  An Adapter is supposed to bridge two interfaces , whereas Decorators wrap an existing interface providing additional functionality.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/HostProxyTest.kt
```diff
@@ -65,10 +73,51 @@ class HostProxyTest : FeatureSpec() {
...
+                                .body(ByteStream(Flux.never()))
+                                .build())
+
+                val hostProxy = HostProxy(HostAndPort.fromString("abc:80"), client, originMetrics)
+
+                hostProxy.handle(LiveHttpRequest.get("/").build(), HttpInterceptorContext.create())
+                        .toMono()
+                        .block()
+                        .let { response ->
+                            response.body()
```
kvosper: My IDE says `Unsafe use of a nullable receiver of type LiveHttpResponse?` Do you know what this means, and is it correct?

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
```diff
@@ -0,0 +1,218 @@
...
+            routeDb.insert("appx-01", RoutingObjectRecord.create("HostProxy", setOf("appX"), mockk(), StaticResponseHandler(200, "appx-01")))
+            routeDb.insert("appx-02", RoutingObjectRecord.create("HostProxy", setOf("appX"), mockk(), StaticResponseHandler(200, "appx-02")))
+            routeDb.insert("appx-03", RoutingObjectRecord.create("HostProxy", setOf("appX"), mockk(), StaticResponseHandler(200, "appx-03")))
+
+            routeDb.insert("appy-01", RoutingObjectRecord.create("HostProxy", setOf("appY"), mockk(), StaticResponseHandler(200, "appy-01")))
+            routeDb.insert("appy-02", RoutingObjectRecord.create("HostProxy", setOf("appY"), mockk(), StaticResponseHandler(200, "appy-02")))
+
+            routeDb.watch().waitUntil { it.entrySet().size == 5 }
+
+            val lbGroup = factory.build(listOf("appX"), RoutingContext(routeDb = routeDb).get(), routingObjectDef("""
```
kvosper: The method `RoutingContext.get` is a bit confusing. If the relationship between `RoutingContext` and `HttpHandlerFactory.Context` can't be refactored, perhaps `get` should be called `adapt` or something?

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
```diff
@@ -0,0 +1,218 @@
...
+                """.trimIndent())) as LoadBalancingGroup
+
+            scenario("Discovers origins with appropriate tag") {
+                val frequencies = mutableMapOf<String, Int>()
+
+                for (i in 1..100) {
+                    lbGroup.call(get("/").build())
+                            .bodyAs(UTF_8)
+                            .let {
+                                val current = frequencies.getOrDefault(it, 0)
```
kvosper: Two lines can be replaced by `frequencies.compute(it) { _, frequency -> (frequency ?: 0) + 1 }`

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
```diff
@@ -0,0 +1,218 @@
...
+                                frequencies[it] = current + 1
+                            }
+                }
+
+                println("Frequencies: " + frequencies)
+
+                frequencies["appx-01"]?.shouldBeGreaterThan(20)
```
kvosper: Is this a special case where a method can be called on `null`?

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
```diff
@@ -0,0 +1,218 @@
...
+                for (i in 1..100) {
+                    lbGroup.call(get("/").build())
+                            .bodyAs(UTF_8)
+                            .let {
+                                val current = frequencies.getOrDefault(it, 0)
+                                frequencies[it] = current + 1
+                            }
+                }
+
+                println("Frequencies: " + frequencies)
```
kvosper:
```suggestion
                
```
Remove debugging `println`s.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
```diff
@@ -0,0 +1,218 @@
...
+fun Publisher<ObjectStore<RoutingObjectRecord>>.waitUntil(duration: Duration = Duration.ofSeconds(1), predicate: (ObjectStore<RoutingObjectRecord>) -> Boolean) = this
+        .toFlux()
+        .filter(predicate)
+        .blockFirst(duration)
+
+
+fun HttpHandler.call(request: HttpRequest, maxContentBytes: Int = 100000) = this.handle(request.stream(), HttpInterceptorContext.create())
+        .flatMap { it.aggregate(maxContentBytes) }
+        .toMono()
+        .block()
```
kvosper:
```suggestion
        .block()

```
I think checkstyle complains about this. Though maybe it doesn't run on kotlin files.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/routing/LoadBalancingGroupSpec.kt
```diff
@@ -0,0 +1,491 @@
...
+                                  config:
+                                    origins: App-A
+                        """.trimIndent())
+
+                for (i in 1..50) {
+                    client.send(get("/").header(HOST, styxServer().proxyHttpHostHeader()).build())
+                            .wait()
+                            .let {
+                                withClue("Response should come from App-A instances only") {
+                                    println("it: $it")
```
kvosper:
```suggestion
                                    
```

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/routing/LoadBalancingGroupSpec.kt
```diff
@@ -0,0 +1,491 @@
...
+                                requestCookie("styx_origin_stickyApp", "NA"))
+                        .build())
+                        .wait()
+                        .let {
+                            it.status() shouldBe OK
+                            it.bodyAs(UTF_8) shouldMatch ("mock-server-0.")
+                        }
+            }
+
+//            scenario("!Routes to new origin when the origin indicated by sticky session cookie is no longer available") {
```
kvosper: Remove commented code and replace with issues if necessary.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
mikkokar: Are you referring to an empty line, or lack of?

kvosper: Our style says to end files with a newline.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
mikkokar: Please clarify. What method are you referring to? The get operator, or `shouldBeGreagerThan` or something else?

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
mikkokar: Thanks for letting me know. But I assume it is acceptable as it is?

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
kvosper: I'm confused about the use of `?.` - normally this would only call `shouldBeGreaterThan` if `frequencies["appx-01"]` were non-null, meaning it would pass the test by being null (which is not the expected behaviour), so I would expect `!!` to be used here instead.

However, further down I see `frequencies["appy-01"]?.shouldBeNull()` - and the method `shouldBeNull` could only have a reason to exist if can somehow be called on null values (though in that case wouldn't the `?.` be better replaced with just `.`?)

So it seems like a confusing feature of the test framework, as it leaves me uncertain about what I should expect.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
kvosper: You can leave it if you prefer.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
mikkokar: I suppose `RoutingContext` needs to be renamed. It really is a "prototype" object for building `HttpHandlerFactory.Context`s.

However I do think that `HttpHandlerFactory`, and therefore `HttpHandlerFactory.Context` also needs a more apt name. The problem is we already have a `RoutingObjectFactory` class.

`RoutingObjectFactory` builds a `RoutingObject` from a given Yaml configuration object (`RoutingObjectConfiguration`). If the configuration defines a new object, it delegates to a particular `HttpHandlerFactory` instance to build the desired object type. If the configuration is just a named reference, it returns a name resolution function wrapped in a lambda.

`HttpHandlerFactory` is an interface. There is a separate implementation for each actual routing object type (like load balancing group, host proxy, etc).

I welcome any suggestions.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
kvosper: If this information is needed to understand a test failure, it should be part of the failure message.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
mikkokar: Thinking about it, current `RoutingObjectFactory` functionality could be implemented directly in a `HttpHandlerFactory` interface as a static method.

Then we could remove the existing `RoutingObjectFactory`, and let the `HttpHandlerFactory` to take its name.

We would then have `RoutingObjectFactory.Context`, and the the Kotlin helper could be renamed as RoutingObjectFactoryContextPrototype` 😱or something along the lines. But I guess that would to be clearer?

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/HostProxyTest.kt
mikkokar: It is because compiler cannot prove that `response` won't be null under any circumstances.

Being correct we should use Kotlin's `?.` operator, but this then propagates further down the chain:
                            response?.body()
                                    ?.toFlux()
                                    ?.subscribe()
                                    ?.dispose()

This being test, it doesn't if we throw an NPE so I'd leave the code as it is.

mikkokar: Actually, I will use a null assertion operator: `!!.`:

                            response!!.body()
                                    .toFlux()
                                    .subscribe()
                                    .dispose()

components/proxy/src/main/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroup.kt
```diff
@@ -0,0 +1,164 @@
...
+                    routingObject,
+                    LoadBalancingMetricSupplier { routingObject.metric() })
+        }
+
+        private fun watchFailed(name: String, cause: Throwable) {
+            LOGGER.error("{} watch error - cause={}", name, cause)
+        }
+
+        private fun watchCompleted(name: String) {
+            LOGGER.error("{} watch complete", name)
```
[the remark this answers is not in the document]
mikkokar: I'd say `"{}: Illegal state: watch completed"`.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/StyxObjectStore.kt
```diff
@@ -171,10 +162,27 @@ class StyxObjectStore<T> : ObjectStore<T> {
...
+    private fun snapshot(snapshot: PMap<String, Record<T>>) = object: ObjectStore<T> {
+        override fun get(key: String?): Optional<T> {
+            return Optional.ofNullable(snapshot.get(key))
+                    .map { it.payload }
+        }
+
+        override fun entrySet(): Collection<Map.Entry<String, T>> = entrySet(snapshot)
+    }
+
+    private fun entrySet(snapshot: PMap<String, Record<T>>) = snapshot.entries.stream()
```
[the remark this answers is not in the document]
mikkokar: Nice!

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/routing/LoadBalancingGroupSpec.kt
mikkokar: Raised #434.

mikkokar: I'll leave this in as a comment:

```
  scenario("!Routes to new origin when the origin indicated by sticky session cookie is no longer available") {
     // See bug: https://github.com/HotelsDotCom/styx/issues/434
  }
```

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
mikkokar: Do you know if there is a way to print out this information after an assertion failed?

kvosper: Actually, I'm not sure. I know that we can do this in Java tests (using Hamcrest for example) but I don't know if there is a kotlintest equivalent. I would hope so.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
mikkokar: > I'm confused about the use of ?. - normally this would only call shouldBeGreaterThan if frequencies["appx-01"] were non-null, meaning it would pass the test by being null (which is not the expected behaviour), so I would expect !! to be used here instead.

Good point. This is clearly a Kotlin newbie error! This actually skips the validation altogether for null values! I'll review all other usages of this.

> However, further down I see frequencies["appy-01"]?.shouldBeNull() - and the method shouldBeNull could only have a reason to exist if can somehow be called on null values (though in that case wouldn't the ?. be better replaced with just .?)

Here you need to realise that `shouldBeNull` is not a method. It is an extension function on `Any?`. See: https://kotlinlang.org/docs/reference/extensions.html. That is, the value `frequencies["apply-01"]` is passed to `shouldBeNull` extension function that just asserts it is null. Therefore the safe call operator is totally unnecessary.

components/proxy/src/main/kotlin/com/hotels/styx/routing/RoutingObjectRecord.kt
```diff
@@ -19,5 +19,14 @@ import com.fasterxml.jackson.databind.JsonNode
 
 data class RoutingObjectRecord(
```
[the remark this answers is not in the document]
mikkokar: "A data class that stores the routing object with associated metadata?"  Are you sure this is helpful?

components/api/src/main/java/com/hotels/styx/api/configuration/ObjectStore.java
mikkokar: Possibly. Can you think of any pros/cons using `Pair` vs `Map.Entry`?

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
mikkokar: Okay. Decorator it is!

components/proxy/src/main/kotlin/com/hotels/styx/routing/RoutingObjectRecord.kt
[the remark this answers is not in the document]
kvosper: A comment should say "why" not "what". We don't need to describe what fields are in the class, but why we want this class to exist in the first place. I think just saying `An entry in the routing object database` would accomplish that goal.
On the other hand, if you think this is redundant, we can leave it out. It's not a huge issue.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
```diff
@@ -0,0 +1,244 @@
...
+            }
+
+            int defaultPort = Optional.ofNullable(tlsSettings)
+                    .map(it -> DEFAULT_TLS_PORT)
+                    .orElse(DEFAULT_HTTP_PORT);
+
+            return HostAndPort.fromParts(hostAndPort.getHostText(), defaultPort);
+        }
+
+        @NotNull
```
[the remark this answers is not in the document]
mikkokar: I'll look into this. Possibly as a separate enhancement.

components/proxy/src/main/java/com/hotels/styx/routing/RoutingObjectAdapter.java
mikkokar: I'll look into this too. Again possibly as a separate enhancement.

components/api/src/main/java/com/hotels/styx/api/configuration/ObjectStore.java
kvosper: `Map.Entry` does have a method `setValue(V value)` for mutating it (or throwing `UnsupportedOperationException`), which could give the illusion of mutability, however most Java programmers will be aware of the awkward way that the JCF handles immutable classes.

Otherwise, it doesn't really make a difference, so if `Map.Entry` is more convenient otherwise, I think we can put up with it.

components/proxy/src/main/kotlin/com/hotels/styx/routing/RoutingObjectRecord.kt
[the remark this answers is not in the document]
mikkokar: This is just a data class and the reason for "why" is that "it is just convenient to bundle this information together". I'm sure you agree?

This class is 100% agnostic to the routing object database or other consumers. The routing database could cease to exist, or cease to use this class at moment. Also this class could find other uses besides the routing object database. Would we then go back and revisit this comment?

For these reason I'd just leave this without any commentary.

components/proxy/src/test/kotlin/com/hotels/styx/routing/handlers/LoadBalancingGroupTest.kt
mikkokar: Do you have any examples how to do this in Hamcrest?

PR #430, opened 2019-06-11
Tidy up code in config classes

kvosper:
The code in these classes has not been looked at for a while.
This PR:
* Replaces calls to deprecated methods with preferred replacements
* Removes unused code
* Remove redundant field initialisation
* Adds missing `final` modifiers to fields
* Removes annotation values that duplicate default value
* Fixes inconsistent style in a couple of places

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/extension/service/HealthCheckConfig.java
+++ b/components/api/src/main/java/com/hotels/styx/api/extension/service/HealthCheckConfig.java
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2018 Expedia Inc.
+  Copyright (C) 2013-2019 Expedia Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -261,7 +261,7 @@ public final class HealthCheckConfig {
         /**
          * Sets the socket timeout for health-checks in a specified unit.
          *
-         * @param timeout  timeout in the specified unit
+         * @param timeout timeout in the specified unit
          * @param timeUnit time unit of timeout
          * @return this builder
          */
--- a/components/api/src/main/java/com/hotels/styx/api/extension/service/HealthCheckConfig.java
+++ b/components/api/src/main/java/com/hotels/styx/api/extension/service/HealthCheckConfig.java
@@ -20,7 +20,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
-import static com.google.common.base.Objects.toStringHelper;
+import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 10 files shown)


components/api/src/main/java/com/hotels/styx/api/extension/service/HealthCheckConfig.java
```diff
@@ -20,7 +20,7 @@
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
-import static com.google.common.base.Objects.toStringHelper;
+import static com.google.common.base.MoreObjects.toStringHelper;
```
mikkokar: Ideally we wouldn't use Guava for anything. It causes problems with plugins. Now when you are at it, perhaps replace the Guava based toString with IntelliJ generated, etc.

Furthermore this 300 lines of code looks like an ordinary data class. With Kotlin it would be perhaps 30 lines max.

components/api/src/main/java/com/hotels/styx/api/extension/service/StickySessionConfig.java
```diff
@@ -20,7 +20,7 @@
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
-import static com.google.common.base.Objects.toStringHelper;
+import static com.google.common.base.MoreObjects.toStringHelper;
```
mikkokar: As above regarding toString.

mikkokar: This is also a data class, a candidate for converting to Kotlin.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteConfig.java
```diff
@@ -20,7 +20,7 @@
 
 import java.util.Optional;
 
-import static com.google.common.base.Objects.toStringHelper;
+import static com.google.common.base.MoreObjects.toStringHelper;
```
mikkokar: Another toStringHelper. Consider using alternative implementation.

components/server/src/main/java/com/hotels/styx/server/HttpConnectorConfig.java
```diff
@@ -18,6 +18,8 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Objects;
 
+import static com.google.common.base.MoreObjects.toStringHelper;
+
```
mikkokar: Consider using alternative toString implementation.

components/server/src/main/java/com/hotels/styx/server/HttpsConnectorConfig.java
```diff
@@ -26,7 +26,7 @@
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
-import static com.google.common.base.Objects.toStringHelper;
+import static com.google.common.base.MoreObjects.toStringHelper;
```
mikkokar: Also here.

components/api/src/main/java/com/hotels/styx/api/extension/service/HealthCheckConfig.java
kvosper: This is PR is intended to just be small incremental changes. I think we should create an issue for larger changes like removing Guava, porting to Kotlin etc.

PR #432, opened 2019-06-12
Add plugin examples to Styx

kvosper:
This PR adds some plugin example classes.


plugin-examples/pom.xml
```diff
@@ -1,205 +1,82 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <parent>
```
dvlato: If we add the parent as a dependency, all the parent pom's dependencies will be transitive dependencies of the plugin. This means that if we wanted to shade it with all its dependencies in a later stage we would re-add (with a different package name) many of Styx dependencies we don't really need...

dvlato: I would probably recommend using composition (i.e., importing a pom) here. In any case, it would be great if you could verify that if we try to use the add the plugin to Styx it's easy to include the required dependencies only.

kvosper: This change was added to ensure that the plugin examples build properly as part of the Styx project. I will see if any alternative is possible.

plugin-examples/pom.xml
```diff
@@ -2,48 +2,34 @@
...
-      <version>${styx.version}</version>
-      <!-- The dependency is explicitlu needed only during compile time, styx is already available in the runtime-->
-      <scope>provided</scope>
+      <version>1.0-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>io.reactivex</groupId>
       <artifactId>rxjava</artifactId>
       <version>${rxjava.version}</version>
-      <scope>provided</scope>
```
dvlato: I would keep the provided scope as this is still not needed in our "plugin bundle"

plugin-examples/pom.xml
```diff
@@ -2,48 +2,34 @@
...
   </properties>
 
-
   <dependencies>
     <dependency>
       <groupId>com.hotels.styx</groupId>
       <artifactId>styx-api</artifactId>
-      <version>${styx.version}</version>
-      <!-- The dependency is explicitlu needed only during compile time, styx is already available in the runtime-->
-      <scope>provided</scope>
```
dvlato: I would still keep the provided scope as we don't need to package/bundle the api with our plugin.

PR #435, opened 2019-06-14
Config validation on programmatic start

kvosper:
Have config validation take place when StyxServer is created programmatically, e.g. for tests.


components/proxy/src/main/kotlin/com/hotels/styx/StyxConfigValidation.kt
```diff
@@ -0,0 +1,58 @@
...
+            throw SchemaValidationException(message);
+        }
+    }
+}
+
+enum class OutcomeType {
+    PASS, FAIL, NO_VALIDATION
+}
+
+data class Outcome(val type: OutcomeType, val error: String? = null)
```
mikkokar: Use sealed classes to represent outcome:

```
sealed class OutcomeType
object ValidationSuccess: OutcomeType()
data class ValidationFailure(val error: String): OutcomeType()
object NoValidation: OutcomeType()
```

They allow you to associate an error string specifically to validation failure only. As an added benefit the `error` can then be a non-nullable `String`.

components/proxy/src/main/kotlin/com/hotels/styx/StyxConfigValidation.kt
```diff
@@ -0,0 +1,58 @@
...
+val LOG: Logger = getLogger("com.hotels.styx")
+
+/**
+ * Used to perform validation on StyxConfig.
+ * StyxConfig can only be validated if using YamlConfiguration underneath.
+ */
+// TODO it would be nice if we could implement this without violating encapsulation by knowing whether YAML is used internally
+// TODO but that should be designed carefully so that our class model still makes sense.
+fun validate(config: StyxConfig) {
+    val outcome: Outcome = config.yamlConfiguration()!!.map {
```
mikkokar: The null safety operator (`!!`) looks out of place. Is it necessary? `config.yamlConfiguration` should always return an optional. 

If Kotlin keeps complaining, try tagging with `NotNull` annotation?

components/proxy/src/test/java/com/hotels/styx/StyxConfigValidationKtTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.testng.annotations.Test;
+
+import static ch.qos.logback.classic.Level.ERROR;
+import static ch.qos.logback.classic.Level.INFO;
+import static com.hotels.styx.common.testing.ExceptionExpectation.expect;
+import static com.hotels.styx.support.matchers.LoggingEventMatcher.loggingEvent;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+public class StyxConfigValidationKtTest {
```
mikkokar: `StyxConfigValidationKtTest`
`____________________^^____`

`Kt` as in Kotlin? In a Java file?

Please implement this in Kotlin. You will get an exception matchers for free, without having to add yet another class. Check our existing Kotlin tests for examples.

components/common/src/main/java/com/hotels/styx/common/testing/ExceptionExpectation.java
```diff
@@ -0,0 +1,59 @@
...
+     * Runs the given action and validates whether it matches the expected exception type.
+     * If it does, the exception is returned.
+     * Otherwise, an AssertionError is thrown.
+     *
+     * @param exceptionType an exception type
+     * @param action an action
+     * @param <E> an exception type
+     * @return the exception
+     */
+    public static <E extends Throwable> E expect(Class<E> exceptionType, Action action) {
```
mikkokar: Please use Kotlintest instead of introducing yet another new class.

components/common/pom.xml
```diff
@@ -97,6 +102,42 @@
...
+<!--            <phase>generate-test-sources</phase>-->
+<!--            <goals>-->
+<!--              <goal>add-test-source</goal>-->
+<!--            </goals>-->
+<!--            <configuration>-->
+<!--              <sources>-->
+<!--                <source>src/test/kotlin</source>-->
+<!--              </sources>-->
+<!--            </configuration>-->
+<!--          </execution>-->
```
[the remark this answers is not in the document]
kvosper: I will delete the commented stuff.

components/proxy/src/main/java/com/hotels/styx/StyxConfig.java
```diff
@@ -64,10 +66,9 @@ public StyxConfig(String yaml) {
     public StyxConfig(StartupConfig startupConfig, Configuration configuration) {
         this.startupConfig = requireNonNull(startupConfig);
         this.configuration = requireNonNull(configuration);
-        this.adminServerConfig = get("admin", AdminServerConfig.class).orElseGet(AdminServerConfig::new);
-        this.proxyServerConfig = get("proxy", ProxyServerConfig.class).orElseGet(ProxyServerConfig::new);
-
-        this.styxHeaderConfig = get("styxHeaders", StyxHeaderConfig.class).orElseGet(StyxHeaderConfig::new);
+        this.proxyServerConfig = lazy(() -> get("proxy", ProxyServerConfig.class).orElseGet(ProxyServerConfig::new));
+        this.adminServerConfig = lazy(() -> get("admin", AdminServerConfig.class).orElseGet(AdminServerConfig::new));
+        this.styxHeaderConfig = lazy(() -> get("styxHeaders", StyxHeaderConfig.class).orElseGet(StyxHeaderConfig::new));
```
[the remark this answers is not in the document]
kvosper: This is not an optimisation, the code doesn't work without this.
This is because validation is applied to the `StyxConfig` object after construction.
We should not try to deserialise objects like `proxy`, `admin`, etc. before validation (in case of invalid config, that results in the Exceptions we want to avoid).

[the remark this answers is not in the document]
mikkokar: Okay. I will have another look.

components/proxy/src/main/kotlin/com/hotels/styx/StyxConfigValidation.kt
```diff
@@ -0,0 +1,54 @@
...
+
+val LOG: Logger = getLogger("com.hotels.styx")
+
+/**
+ * Used to perform validation on StyxConfig.
+ * StyxConfig can only be validated if using YamlConfiguration underneath.
+ */
+// TODO it would be nice if we could implement this without violating encapsulation by knowing whether YAML is used internally
+// TODO but that should be designed carefully so that our class model still makes sense.
+fun validate(config: StyxConfig) {
```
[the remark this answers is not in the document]
mikkokar: Here is how StyxServer is built programmatically:

```
    val styxServer = StyxServer(
            StyxServerComponents.Builder()
                    .styxConfig(StyxConfig.fromYaml(yamlText))
                    .build())
```

A `StyxConfig` object is created, and that injected to `StyxServerComponents`. As you say @dvlato it seems sensible to validate the server configuration right there in the `StyxConfig` class.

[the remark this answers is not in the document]
dvlato: I totally agree it seems the StyxConfig construction itself could take care of the config validation.  That might help us unify the two different 'validation points'.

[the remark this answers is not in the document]
kvosper: There is a down side here, though. Currently, is is possible to enable or disable config validation programmatically in `StyxServerComponents`. If we move it into `StyxConfig` we will need to either add more overloaded constructors for `disableConfigValidation` or have it always check the system property, even in tests, which could lead to tests affecting each other, since system properties live for the lifetime of the JVM.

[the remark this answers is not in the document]
mikkokar: >There is a down side here, though. Currently, is is possible to enable or disable config validation programmatically in StyxServerComponents. If we move it into StyxConfig we 

Not on `master` though. 

> will need to either add more overloaded constructors for disableConfigValidation or 

As a starting point, try to remove `StartupConfig` out of `StyxConfig`. It is not really necessary there, but this will bring down the number of constructors. AFAICS the `StartupConfig` is necessary for the admin interface. Just pass it there separately.

Also consider removing either `StyxConfig` default constructor, or `defaultConfig` static factory method. Do we really need both?

Similarly `StyxConfig(String yaml)` should be removed in favour of `fromYaml` static factory method. You can then perform the config validation in this method before constructing the object. You can also provide an overload offering an option to disable validation:

```
public static StyxConfig fromYaml(String yamlText) // always validates
public static StyxConfig fromYaml(String yamlText, boolean validate)
```

Obviously there is no need to do anything about `StyxConfig(Configuration)` constructor because we cannot apply validation to `Configuration` objects.

components/proxy/src/main/java/com/hotels/styx/StyxConfig.java
```diff
@@ -19,62 +19,43 @@
...
-    private final ProxyServerConfig proxyServerConfig;
-    private final AdminServerConfig adminServerConfig;
-    private final StyxHeaderConfig styxHeaderConfig;
-
     public StyxConfig() {
-        this(defaultStartupConfig(), EMPTY_CONFIGURATION);
+        this(EMPTY_CONFIGURATION);
     }
 
     public StyxConfig(Configuration configuration) {
```
dvlato: I understood we were going to hide these constructors so we can make sure that configuration is validated if necessary.

components/proxy/src/main/java/com/hotels/styx/StyxConfig.java
```diff
@@ -93,40 +74,30 @@ private static Configuration loadYamlConfiguration(String yaml) {
     }
 
     public StyxHeaderConfig styxHeaderConfig() {
-        return styxHeaderConfig;
+        return get("styxHeaders", StyxHeaderConfig.class).orElseGet(StyxHeaderConfig::new);
```
dvlato: Is this still needed?

components/proxy/src/main/java/com/hotels/styx/StyxConfig.java
kvosper: No we are just deleting some constructors that are no longer needed since we moved YAML parsing to a static method, and extracted `StartupConfig` to be handled separately.

components/proxy/src/main/java/com/hotels/styx/StyxConfig.java
kvosper: Not sure what you mean, but the change here is to simplify the way this object is constructed, which is still a valid refactoring.

dvlato: I don't see why we'd want to generate a new StyxHeaderConfig/Proxyconfig etc. object anytime we need to read it (that would be the case if the "configuration" object did not include the block). Isn't it simpler/preferable if it our StyxConfig object is immutable?

components/proxy/src/main/java/com/hotels/styx/StyxConfig.java
dvlato: Ok... I would still prefer a model in which we can validate any "Configuration" (if the validation is implemented!) but since we don't have any such validation it might be a bit redundant. Setting this to resolved.

components/proxy/src/main/java/com/hotels/styx/StyxConfig.java
kvosper: These methods are rarely called - they are only needed at startup, so putting the parsing in the constructor was a premature optimisation.

Please note that these methods do not mutate StyxConfig in any way, so the class remains immutable.

dvlato: Of course, if you don't create fields in StyxConfig it is not mutable...  Should I have said I would rather have a snapshot of the whole config instead of having to load on demand? 

The "only needed at startup" assertion requires quite some code reading to verify. Specially in the case of ProxyToBackend.Factory.build() and StyxPipelineFactory.create(), and is probably contradicted by BackendServicesRouter.onChange() ... 

I don't think that a change that makes code simpler to reason about is an optimization but a good practice.

kvosper: I agree that it makes it easier to reason about. I will see if I can just revert that part.

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
```diff
@@ -262,7 +278,14 @@ public StyxServerComponents build() {
...
         };
-        LoggingSetUp FROM_CONFIG = environment -> setUpLogging(environment.configuration().logConfigLocation());
+        LoggingSetUp FROM_CONFIG = environment -> setUpLogging(
+                Paths.get(environment
+                        .startupConfig()
+                        .logConfigLocation()
+                        .url()
+                        .getFile())
+                        .toString()
+        );
```
mikkokar: I believe this is the only reason to add the `startupConfig` into `Environment`. We should look into moving the `FROM_CONFIG` out of this file. Perhaps it could be just a lambda in the `StyxServer` class.

If so, then `startupConfig` could be removed from environment. This would be a big plus as it would narrow down `startupConfig` usage to just admin interface where it is used merely to display the config file locations.

mikkokar: Approving regardless of the above comment.

PR #437, opened 2019-06-18
Fix connection pool error.

mikkokar:
`Publisher<Connection>` was not emitted in some corner cases, leading to unnecessary timeouts from `borrowConnection`.

Also, `connectionsInEstablishment` was not decremented in some corner cases, resulting in a slow drift up that could end up preventing new connections from being established.

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
+++ b/components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
@@ -76,26 +76,17 @@ public class SimpleConnectionPool implements ConnectionPool, Connection.Listener
         return Mono.<Connection>create(sink -> {
             Connection connection = dequeue();
             if (connection != null) {
-                if (borrowedCount.getAndIncrement() < poolSettings.maxConnectionsPerHost()) {
-                    sink.success(connection);
-                } else {
-                    borrowedCount.decrementAndGet();
-                    queueNewConnection(connection);
-                }
+                borrowedCount.incrementAndGet();
+                sink.success(connection);
             } else {
-                if (waitingSubscribers.size() >= poolSettings.maxPendingConnectionsPerHost()) {
+                if (waitingSubscribers.size() < poolSettings.maxPendingConnectionsPerHost()) {
+                    this.waitingSubscribers.add(sink.onDispose(() -> waitingSubscribers.remove(sink)));
+                    newConnection();
+                } else {
                     sink.error(new MaxPendingConnectionsExceededException(
                             origin,
                             poolSettings.maxPendingConnectionsPerHost(),
                             poolSettings.maxPendingConnectionsPerHost()));
-                } else {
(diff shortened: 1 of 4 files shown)


components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -76,26 +76,17 @@ public Origin getOrigin() {
...
-                } else {
-                    borrowedCount.decrementAndGet();
-                    queueNewConnection(connection);
-                }
+                borrowedCount.incrementAndGet();
+                sink.success(connection);
             } else {
-                if (waitingSubscribers.size() >= poolSettings.maxPendingConnectionsPerHost()) {
+                if (waitingSubscribers.size() < poolSettings.maxPendingConnectionsPerHost()) {
+                    this.waitingSubscribers.add(sink.onDispose(() -> waitingSubscribers.remove(sink)));
```
dvlato: Should we use  doOn operators (eg doFinally() for this one and for the connectionsInEstablishment variable as well) for these  side-effects?

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -104,6 +95,14 @@ public Origin getOrigin() {
     }
 
     private void newConnection() {
+        int borrowed = borrowedCount.get();
+        int inEstablishment = connectionsInEstablishment.getAndIncrement();
```
dvlato: Are we incrementing this value here just to reduce the number of calls to the AtomicInteger? It looks a bit odd that we increment it to decrement it straightaway in some cases, since we don't do the same for the borrowedCount.   I would consider alternative implementations.

dvlato:
An additional small change: do you think we should rename "activeConnections" to "availableConnections"?

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
mikkokar: Please clarify. Where would we use the `doOn/doFinally` operator?

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
mikkokar: No. This is to avoid a race condition in relation to `connectionsInEstablishment`. So if we'd do this:

    if (borrowed + connectionsInEstablishment.get() >= max) {
        return;
    } 
    connectionsInEstablishment.increment();
    

Then it is possible (I would say even likely) that several other threads will simultaneously call `get` and pass the test, when they should have failed.

But a call to `getAndIncrement` ensures that each thread sees an updated value and only one (or some) of them will pass the test while others will fail.

However this is not necessary for `borrowCount` because it is never modified in this method. It is modified (atomically) elsewhere whenever a connection is handed out to a borrower, or a connection is returned back to the pool.

PR #438, opened 2019-06-28
Health check monitor

mikkokar:
Adds a health check monitor for the routing objects.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/StyxServer.java
+++ b/components/proxy/src/main/java/com/hotels/styx/StyxServer.java
@@ -173,6 +173,16 @@ public final class StyxServer extends AbstractService {
         this.proxyServer = proxyServerSetUp.createProxyServer(components);
         this.adminServer = createAdminServer(components);
 
+        components.servicesDatabase()
+                .entrySet()
+                .forEach(entry -> {
+                    StyxService service = entry.getValue().getStyxService();
+                    service.start()
+                            .thenAccept(ignore -> {
+                                LOG.debug("Service '{}/{}' started", entry.getValue().getType(), entry.getKey());
+                            });
+                });
+
         this.serviceManager = new ServiceManager(new ArrayList<Service>() {
             {
                 add(proxyServer);
--- a/components/proxy/src/main/java/com/hotels/styx/StyxServer.java
+++ b/components/proxy/src/main/java/com/hotels/styx/StyxServer.java
@@ -42,7 +42,9 @@ import java.io.Reader;
 import java.net.InetSocketAddress;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import static com.hotels.styx.infrastructure.logging.LOGBackConfigurer.initLogging;
 import static com.hotels.styx.infrastructure.logging.LOGBackConfigurer.shutdownLogging;
@@ -179,11 +181,8 @@ public final class StyxServer extends AbstractService {
                 add(proxyServer);
                 add(adminServer);
 
-                ImmutableList.<StyxService>builder()
-                    .add(new ServiceProviderMonitor("Styx-Service-Monitor", components.servicesDatabase()))
-                    .addAll(components.services().values())
-                    .build()
-                    .stream()
+                concat(components.services().values(),
+                        new ServiceProviderMonitor("Styx-Service-Monitor", components.servicesDatabase()))
                     .map(StyxServer::toGuavaService)
                     .forEach(this::add);
             }
@@ -305,4 +304,13 @@ public final class StyxServer extends AbstractService {
             styxServer.notifyStopped();
         }
     }
+
+    private static Stream<StyxService> concat(Collection<StyxService> services, StyxService service) {
+        return ImmutableList.<StyxService>builder()
+                .add(service)
+                .addAll(services)
+                .build()
+                .stream();
+    }
+
 }
--- a/components/proxy/src/main/java/com/hotels/styx/StyxServer.java
+++ b/components/proxy/src/main/java/com/hotels/styx/StyxServer.java
@@ -16,6 +16,7 @@
 package com.hotels.styx;
 
 import com.google.common.base.Stopwatch;
+import com.google.common.collect.ImmutableList;
 import com.google.common.io.CharStreams;
 import com.google.common.util.concurrent.AbstractService;
 import com.google.common.util.concurrent.Service;
@@ -173,23 +174,18 @@ public final class StyxServer extends AbstractService {
         this.proxyServer = proxyServerSetUp.createProxyServer(components);
         this.adminServer = createAdminServer(components);
 
-        components.servicesDatabase()
-                .entrySet()
-                .forEach(entry -> {
-                    StyxService service = entry.getValue().getStyxService();
-                    service.start()
-                            .thenAccept(ignore -> {
-                                LOG.debug("Service '{}/{}' started", entry.getValue().getType(), entry.getKey());
-                            });
-                });
-
         this.serviceManager = new ServiceManager(new ArrayList<Service>() {
             {
                 add(proxyServer);
                 add(adminServer);
-                components.services().values().stream()
-                        .map(StyxServer::toGuavaService)
-                        .forEach(this::add);
+
+                ImmutableList.<StyxService>builder()
+                    .add(new ServiceProviderMonitor("Styx-Service-Monitor", components.servicesDatabase()))
+                    .addAll(components.services().values())
+                    .build()
+                    .stream()
+                    .map(StyxServer::toGuavaService)
+                    .forEach(this::add);
             }
         });
     }
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/RoutingObjectHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/RoutingObjectHandler.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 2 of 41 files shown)


components/proxy/src/test/kotlin/com/hotels/styx/services/HealthChecksTest.kt
```diff
@@ -0,0 +1,186 @@
...
+        scenario("A successful probe resets inactive count") {
+            check(ObjectActive(2), true) shouldBe ObjectActive(0)
+        }
+
+        scenario("A successful probe doesn't affect negative probes in Active state") {
+            check(ObjectActive(0), true) shouldBe ObjectActive(0)
+        }
+    }
+
+    feature("CheckOriginState") {
```
kvosper: I can't see where this is asserting/verifying anything. Is there some way it can be refactored to make it easier to understand?

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/providers/HealthCheckProviderSpec.kt
```diff
@@ -0,0 +1,265 @@
...
+                  healthyThreshold: 3
+                  unhealthyThreshold: 2
+
+            httpPipeline:
+                type: LoadBalancingGroup
+                config:
+                  origins: aaa
+            """.trimIndent())
+
+    val httpClient = StyxHttpClient.Builder().build()
```
kvosper:
```suggestion

```
Unused field

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -161,9 +162,14 @@ public StyxServer(StyxServerComponents components, Stopwatch stopwatch) {
             {
                 add(proxyServer);
                 add(adminServer);
-                components.services().values().stream()
-                        .map(StyxServer::toGuavaService)
-                        .forEach(this::add);
+
+                ImmutableList.<StyxService>builder()
```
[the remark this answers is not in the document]
mikkokar: Okay.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/RoutingObjectHandler.java
```diff
@@ -117,12 +118,12 @@ public RoutingObjectHandler(StyxObjectStore<RoutingObjectRecord> routeDatabase,
                 .build();
     }
 
-    private static String serialise(String name, RoutingObjectRecord app) {
+    private static String serialise(String name, RoutingObjectRecord record) {
         JsonNode node = YAML_MAPPER
-                .addMixIn(RoutingObjectDefinition.class, RoutingObjectDefMixin.class)
-                .valueToTree(new RoutingObjectDefinition(name, app.getType(), emptyList(), app.getConfig()));
+                .addMixIn(StyxObjectDefinition.class, RoutingObjectDefMixin.class)
+                .valueToTree(new StyxObjectDefinition(name, record.getType(), ImmutableList.copyOf(record.getTags()), record.getConfig()));
```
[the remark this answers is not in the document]
mikkokar: It already takes a defensive copy. The `ImmutableList.copyOf` is just converting a type from a Set to a List. 

The tags are stored in `record` as Set to disallow duplicates. They are stored as List in StyxObjectDefinition due to YAML serialisation.

But perhaps they should use the same types.

I'll see if I can somehow improve this.

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
```diff
@@ -106,22 +122,42 @@ private StyxServerComponents(Builder builder) {
...
+                .map(StyxServerComponents::readComponents)
                 .orElse(ImmutableMap.of())
-                .forEach((name, record) -> {
-                    RoutingMetadataDecorator adapter = new RoutingMetadataDecorator(Builtins.build(ImmutableList.of(name), routingObjectContext, record));
-                    routeObjectStore.insert(name, new RoutingObjectRecord(record.type(), ImmutableSet.copyOf(record.tags()), record.config(), adapter))
+                .forEach((name, definition) -> {
+                    RoutingObject routingObject = Builtins.build(ImmutableList.of(name), routingObjectContext, definition);
+                    RoutingMetadataDecorator adapter = new RoutingMetadataDecorator(routingObject);
+
+                    routeObjectStore.insert(name, new RoutingObjectRecord(definition.type(), ImmutableSet.copyOf(definition.tags()), definition.config(), adapter))
```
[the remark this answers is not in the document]
mikkokar: This is not necessary because `RoutingObjectRecord` already uses Kotlin immutable Set.

components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
```diff
@@ -0,0 +1,195 @@
...
+import java.time.Duration
+import java.util.Optional
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.ScheduledThreadPoolExecutor
+import java.util.concurrent.TimeUnit.MILLISECONDS
+import java.util.concurrent.atomic.AtomicReference
+
+internal class HealthCheckMonitoringService(
```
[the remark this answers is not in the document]
mikkokar: The argument being val (a property) or not visible to consumers instantiating this class. But now I do realise that `val` creates public properties. I should actually mark them as private.

The arguments are in a logical order that follows object configuration roughly in an order of importance: `path` first as it is mandatory, followed by timeouts, and then thresholds.  

Also, if an argument later on needs to be converted to a property (with `val`), or if it no longer needs to be a property (remove `val`), would you then reshuffle the argument list respectively?

components/proxy/src/test/kotlin/com/hotels/styx/routing/Support.kt
```diff
@@ -59,21 +62,38 @@ fun HttpHandler.handle(request: HttpRequest, count: Int = 10000) = this.handle(r
...
-    val refLookup = HashMap<RoutingObjectReference, RoutingObject>()
+fun routeLookup(block: HashMap<StyxObjectReference, RoutingObject>.() -> Unit): RouteRefLookup {
+    val refLookup = HashMap<StyxObjectReference, RoutingObject>()
     refLookup.block()
     return RouteRefLookup { refLookup[it] }
 }
 
-fun mockObject(content: String = "") = mockk<RoutingObject> {
-    every { handle(any(), any()) } returns Eventual.of(HttpResponse.response(OK).body(content, UTF_8).build().stream())
+fun mockObject(content: String = "", capture: ArgumentCapture = DontCapture) = mockk<RoutingObject> {
```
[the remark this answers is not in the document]
mikkokar: I am in agreement. This stuff needs to be documented in this file.

Will do.

components/proxy/src/test/kotlin/com/hotels/styx/services/HealthChecksTest.kt
mikkokar: This is some old throwaway code that I forgot to remove.
(Removed now).

components/proxy/src/main/java/com/hotels/styx/routing/config/Builtins.java
```diff
@@ -87,35 +97,42 @@
     private Builtins() {
     }
 
-    public static RoutingObject build(List<String> parents, RoutingObjectFactory.Context context, RoutingObjectConfiguration configNode) {
-        if (configNode instanceof RoutingObjectDefinition) {
-            RoutingObjectDefinition configBlock = (RoutingObjectDefinition) configNode;
+    public static RoutingObject build(List<String> parents, RoutingObjectFactory.Context context, StyxObjectConfiguration configNode) {
```
kvosper: We have three public methods here named `build` but with different arguments. I think some comments/documentation are in order here to explain what they are for and the differences between them.

components/proxy/src/main/java/com/hotels/styx/serviceproviders/ServiceProviderFactory.java
```diff
@@ -0,0 +1,39 @@
...
+import com.hotels.styx.routing.db.StyxObjectStore;
+
+/**
+ * A generic factory that can be implemented in order to facilitate creating objects whose type is not known
+ * until read from configuration.
+ *
+ */
+public interface ServiceProviderFactory {
+    /**
+     * Create an instance of the product.
```
kvosper:
```suggestion
     * Create an instance of the service provider.
```

components/proxy/src/main/java/com/hotels/styx/serviceproviders/ServiceProviderFactory.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @param environment          environment
+     * @param serviceConfiguration configuration specific to the factory product
```
kvosper:
```suggestion
     * @param serviceConfiguration configuration specific to the service provider
```

components/proxy/src/main/java/com/hotels/styx/serviceproviders/ServiceProviderFactory.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @return product instance
```
kvosper:
```suggestion
     * @return service provider instance
```

components/proxy/src/main/java/com/hotels/styx/routing/config/Builtins.java
mikkokar: Okay. The return type gives a clue what they are for. But will add documentation regardless.

components/proxy/src/main/java/com/hotels/styx/routing/config/Builtins.java
```diff
@@ -87,35 +97,69 @@
     private Builtins() {
     }
 
-    public static RoutingObject build(List<String> parents, RoutingObjectFactory.Context context, RoutingObjectConfiguration configNode) {
-        if (configNode instanceof RoutingObjectDefinition) {
-            RoutingObjectDefinition configBlock = (RoutingObjectDefinition) configNode;
+    /**
+     * Buiulds a routing object.
```
kvosper:
```suggestion
     * Builds a routing object.
```

PR #443, opened 2019-07-04
Configurable docker base image version.

mikkokar:
Add a build time parameter to choose a Docker base image tag.

This allows to choose between the JRE only (default) base, and full JDK that has `jps`, `jcmd` & other tools handy for development purposes.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-ARG TAG=11-jre-slim
+ARG TAG=11-jdk
 FROM openjdk:${TAG}
 
 ARG STYX_VERSION=""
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM openjdk:11-jre-slim
+ARG TAG=11-jre-slim
```
dvlato: Do we want to use jre by default? Bundling the JDK tools might prove useful in several environments.

mikkokar: I am not opinionated about this. My intention was to default on the "production" image. What do you think?

mikkokar: I have changed to JDK version. Please have another look.

PR #444, opened 2019-07-08
Print config invalid error, add testing

kvosper:


components/proxy/src/test/java/com/hotels/styx/StyxServerTest.java
```diff
@@ -177,6 +182,38 @@ public void serverDoesNotStartIfServiceFails() {
...
+            setProperty("CONFIG_FILE_LOCATION", Paths.get(fixturesHome()).resolve("conf/invalid.yml").toString());
+
+            Runtime runtime = captureSystemExit(() -> StyxServer.main(new String[0]));
+            verify(runtime).exit(2);
+
+            eventually(() -> assertThat(log.log(), hasItem(loggingEvent(ERROR,
+                    "Styx server failed to start due to configuration error in file .+: Missing a mandatory field 'proxy'"
+            ))));
+        } finally {
+            clearProperty("STYX_HOME");
```
dvlato: should we also clear the "UNIT_TESTING_MODE"  property?

kvosper: We can, but we're not going to be running the server for real inside the same JVM as the unit tests. Do you think we should do it anyway for consistency? Maybe we could do it in set-up and tear-down methods?

dvlato: I would either set the property for the whole test suite or do the cleanup there. In general this "it's an unit test" setting makes a bit uneasy; I've done it before in other projects but I think it can become a bit ugly to manage. In this case we could probably set up the logging differently so we don't need those loggingSetUp loggingSetUp.

components/proxy/src/test/resources/conf/default.yml
```diff
@@ -0,0 +1,97 @@
...
+#plugins:
+#  active: demo
+#  all:
+#    demo:
+#      factory:
+#        class: "com.hotels.styx.startup.extensions.DemoPlugin$Factory"
+#        classPath: "${STYX_HOME:}"
+#      config:
+#        adminText: "this is the admin page"
+#        responseHeaderValue: "this is the response header value"
```
mikkokar: I would remove the commented configuration.

components/proxy/src/test/resources/conf/invalid.yml
```diff
@@ -0,0 +1,104 @@
...
+#  active: demo
+#  all:
+#    demo:
+#      factory:
+#        class: "com.hotels.styx.startup.extensions.DemoPlugin$Factory"
+#        classPath: "${STYX_HOME:}"
+#      config:
+#        adminText: "this is the admin page"
+#        responseHeaderValue: "this is the response header value"
+
```
mikkokar: Please remove the commented configuration.

PR #446, opened 2019-07-10
Connection pool utilisation metric

kvosper:
Add connection pool utilisation metric

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
+++ b/components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
@@ -184,8 +184,8 @@ public class SimpleConnectionPool implements ConnectionPool, Connection.Listener
     }
 
     private double utilisation() {
-        int usage = borrowedCount.get() + waitingSubscribers.size();
-        int limit = poolSettings.maxConnectionsPerHost() + poolSettings.maxPendingConnectionsPerHost();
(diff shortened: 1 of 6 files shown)


components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -187,6 +183,13 @@ public void connectionClosed(Connection connection) {
         return this.stats;
     }
 
+    private double utilisation() {
+        double usage = borrowedCount.get() + waitingSubscribers.size();
+        double limit = poolSettings.maxConnectionsPerHost() + poolSettings.maxPendingConnectionsPerHost();
```
dvlato: is this correct? I would think the limit is "maxConnectionsPerHost", so if we allow 100 connections and we have 100 established connections, the pool is completely used.

Usage would probably be also borrowedCount? I don't know if it makes sense that we say that we have a pool usage of 120%. Our usage is 100% and then there's a few connections waiting to be served, those will fail if  borrowedCount does not go down...  

In general, is this utilisation metric required if we are publishing all the others?

mikkokar: It would be more sensible to publish the pool size (max active connections, max pending subscribers) as metrics, and calculate the utilisation in the dashboards.

PR #449, opened 2019-07-25
Consume request body on early return

kvosper:
These code changes add consumption of request upon early return to many classes.
This is done in two ways:
1. Some classes are switched from `HttpHandler` to `WebServiceHandler`
2. Some classes use convenience methods from the new `EarlyReturn` class

Option 1 is preferred (and most such conversions have already been done in previous PRs), but in some cases, such as when branching can lead to propagation of `LiveHttpRequest` instead of early-return, option 2 has been used.

You may also see some unrelated files have automatically had their copyright notices updated. I advocate just allowing these changes in, since the build process is going to keep making the changes on each build and it wastes time removing them again after every end-to-end test run.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/LiveHttpMessage.java
+++ b/components/api/src/main/java/com/hotels/styx/api/LiveHttpMessage.java
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2018 Expedia Inc.
+  Copyright (C) 2013-2019 Expedia Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -14,6 +14,8 @@
   limitations under the License.
  */
 package com.hotels.styx.api;
+
+import org.reactivestreams.Publisher;
 
 import java.util.List;
 import java.util.Optional;
@@ -102,6 +104,7 @@ interface LiveHttpMessage {
      * <p>
      */
     default void consume() {
+        // TODO is there any point to this aggregate call? Can't we just do body().drop().subscribe(...) or something?
         body().drop().aggregate(1)
                 .thenApply(buffer -> {
                     buffer.delegate().release();
@@ -109,5 +112,8 @@ interface LiveHttpMessage {
                 });
     }
 
-
+    default <T> Eventual<T> consume2() {
+        // Note: since the ByteStream will have zero elements, we can safely ignoring the original element type of body().drop().
+        return new Eventual<T>((Publisher) body().drop());
+    }
 }
--- a/components/api/src/main/java/com/hotels/styx/api/LiveHttpMessage.java
+++ b/components/api/src/main/java/com/hotels/styx/api/LiveHttpMessage.java
@@ -16,12 +16,16 @@
 package com.hotels.styx.api;
 
 import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
 
 import java.util.List;
 import java.util.Optional;
 
 import static com.hotels.styx.api.HttpHeaderNames.CONTENT_LENGTH;
 import static com.hotels.styx.api.HttpHeaderNames.CONTENT_TYPE;
+import static java.lang.Long.MAX_VALUE;
+import static org.slf4j.LoggerFactory.getLogger;
 
 /**
  * All behaviour common to both streaming requests and streaming responses.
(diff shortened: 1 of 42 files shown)


components/common/src/test/java/com/hotels/styx/api/EarlyReturnTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import static com.hotels.styx.api.HttpResponseStatus.OK;
+import static com.hotels.styx.api.LiveHttpRequest.post;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+public class EarlyReturnTest {
+    @Test
+    public void bodyIsConsumedWhenReturningEarlyWithError() {
```
dvlato: Do we have a test that shows the opposite behaviour?  Body is not consumed when not returning early.

kvosper: Not sure what that would be? These tests test what happens when convenience methods are called. The opposite would be to test that nothing happens when you don't call any methods. 

Perhaps the test names are unclear? Please let me know.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/RouteRefLookup.java
```diff
@@ -53,15 +53,13 @@ public RoutingObject apply(StyxObjectReference route) {
...
-                        liveRequest.consume();
-
-                        return Eventual.of(response(NOT_FOUND)
-                                .body("Not found: " + route.name(), UTF_8)
-                                .build()
-                                .stream()
-                        );
-                    });
+                    .orElse((liveRequest, na) ->
+                            returnEarlyWithResponse(liveRequest,
```
dvlato: I would think consume() was just discarding the body instead of storing it. Is the change ok?

kvosper: Good point. `consume` simply discards each element as it is received, whereas `aggregate` will keep the content in memory until it is complete, or the byte limit is reached.

The idea to use `aggregate` only was to ensure we had a byte limit, however it does use memory unnecessarily. I will look into how to avoid this. I originally discussed with @mikkokar, but this detail was missed. Any opinions Mikko?

components/api/src/main/java/com/hotels/styx/api/LiveHttpMessage.java
```diff
@@ -100,14 +105,43 @@ default boolean chunked() {
...
+            }
+
+            @Override
+            public void onNext(Object o) {
+                // ignore
+            }
+
+            @Override
+            public void onError(Throwable t) {
+                // this will get the implementation class
```
dvlato: is this comment needed?

components/api/src/main/java/com/hotels/styx/api/LiveHttpMessage.java
```diff
@@ -100,14 +105,43 @@ default boolean chunked() {
...
+     *
+     * The consumption begins immediately, running in the background.
      */
-    default void consume() {
-        body().drop().aggregate(1)
-                .thenApply(buffer -> {
-                    buffer.delegate().release();
-                    return null;
-                });
+    default void consumeInBackground() {
```
mikkokar: I was hoping to get rid of this method. It is susceptible for denial of service: It keeps consuming as long as remote client keeps sending traffic.

Also this particular implementation looks to leak memory, as objects are not down referenced at `onNext`.
I'm sure this is not the case as you have deliberately left a comment (`// ignore`), but looking at this I'm still clueless as how it is supposed to work?

components/api/src/main/java/com/hotels/styx/api/LiveHttpRequest.java
```diff
@@ -383,6 +383,11 @@ public Transformer newBuilder() {
                 .findFirst();
     }
 
+    @Override
+    public Eventual<LiveHttpRequest> consume() {
+        return MessageBodyConsumption.consume(this, LiveHttpRequest.class);
```
mikkokar: HTTP message body is the same for both requests and responses. Therefore you shouldn't have to pass in a class as a  parameter.

components/api/src/main/java/com/hotels/styx/api/MessageBodyConsumption.java
```diff
@@ -0,0 +1,38 @@
...
+package com.hotels.styx.api;
+
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Mono;
+
+final class MessageBodyConsumption {
+    private MessageBodyConsumption() {
+    }
+
+    static <T extends LiveHttpMessage> Eventual<T> consume(T message, Class<T> type) {
```
mikkokar: Please avoid overly clever code reuse. Follow the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) instead.

There isn't any need to pass in class parameters etc because the message body stream is the same for both request and response types.

mikkokar:
I would suggest you take `ByteStreamAggregator` class as a basis, to create a new class for discarding a body up to and until max N bytes. Then use that class to implement `consume` method just like we do for `aggregate`.

Have you explored this option?

components/api/src/main/java/com/hotels/styx/api/LiveHttpMessage.java
kvosper: Please note that the objects from `consume()` will be requests or responses that have already had the body consumed, so there is no further down-referencing necessary at this point. The comment says `// ignore` because it is not necessary to do anything with the request/response when called from this method.

I also wanted to get rid of it. However, I got a test failure in `HealthChecks.kt` when replacing ```                   .map {
                        it.consumeInBackground()
                        it.status().code() < 400
                    }``` 
with
```.flatMap {
  it.consume()
}
.map {
  it.status().code() < 400
}```

mikkokar:
Keeping this open for Kyle ;-)

PR #455, opened 2019-08-22
Optimise response event listener

mikkokar:
## Problem 

ResponseEventListener adds unnecessary pressure on heap:

* It constructs a `StateMachine` for the response that is being monitored. The underlying `StateMachine` loads of `Map.Entry` and `Key` objects.
* It uses a separate `LiveHttpResponse` builder for each *ContentError*, *ContentEnd* and *ContentCancelled* events. 

# Fix

* Replace `StateMachine` object with nested `switch/case` and `if/else` statements. 
* Collapse *ContentError*, *ContentEnd*, and *ContentCancelled* event response body content callbacks in a single `LiveHttpResponse` builder to avoid excessive HTTP header copying.

# Impact

On routing object based application router:
* HeaderEntry: ~ 40% less objects
* HeaderEntry[]: ~ 33% less objects
* Object[]: ~ 37% less objects
And so on.

### DIFF
--- a/components/common/src/main/java/com/hotels/styx/api/ResponseEventListener.java
+++ b/components/common/src/main/java/com/hotels/styx/api/ResponseEventListener.java
@@ -20,8 +20,10 @@ import com.hotels.styx.common.QueueDrainingEventProcessor;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 
+import java.util.Optional;
 import java.util.function.Consumer;
 
+import static com.hotels.styx.api.ResponseEventListener.State.INITIAL;
 import static com.hotels.styx.api.ResponseEventListener.State.COMPLETED;
 import static com.hotels.styx.api.ResponseEventListener.State.STREAMING;
 import static com.hotels.styx.api.ResponseEventListener.State.TERMINATED;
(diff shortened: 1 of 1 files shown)


components/common/src/main/java/com/hotels/styx/api/ResponseEventListener.java
```diff
@@ -119,19 +83,59 @@ public ResponseEventListener whenFinished(Runnable action) {
...
+                    switch (state) {
+                        case INITIAL:
+                            if (event instanceof MessageHeaders) {
+                                state = STREAMING;
+                            } else if (event instanceof MessageCancelled) {
+                                cancelAction.run();
+                                whenFinishedAction.run();
+                                state = TERMINATED;
+                            } else if (event instanceof MessageCompleted) {
+                                // TODO: Add custom exception type?
```
dvlato: Should we deal with this TODO / remove it?  What do you think of dating TODO comments?

mikkokar: Hi David. This is an old TODO that appears as a "new" feature in the diffs.

A date could be helpful for code reviewers to spot old TODO items that have merely moved around. But I'm not opinionated about this.

Issue #454, opened 2019-08-21
Sanitise HTTP message logs


mikkokar:
# Problem Statement

Styx logs HTTP messages for access logs, errors, warnings and so on. The messages are constructed by `toString` method, which includes all HTTP fields in verbatim, including sensitive information like security tokens.

Styx needs a configurable mechanism to sanitise the HTTP messages before logging them.

# Acceptance Criteria

- Remove `toString` method from `HttpRequest`, `HttpResponse`, `LiveHttpRequest`, `LiveHttpResponse`, and any possible base classes. This is to prevent developers from logging them accidentally.

- Add a "removeCookies" option, that is a list of cookie names, in styx server under `request-logging` section. Any cookie name in this list gets removed from the logged message output.
 
```
    request-logging:
      inbound:
          ...
      outbound:
          ...
      removeCookies:
         - sessionToken
         - anotherToken
```

- TODO: do we need `removeHeaders` option. Similar to `removeCookies` above.

- The new config is documented in styx end user manual.

- The new config option is added to styx config validator, so that it can pick up any invalid configuration.

mikkokar:
Hi @fantayeneh and @taer. Please check this bug. 

@OwenLindsell could you look into this please?

fantayeneh:
I would add an option to redact headers and cookies. That will make debugging easy.

OwenLindsell:
> I would add an option to redact headers and cookies. That will make debugging easy.

Definitely something we will consider

OwenLindsell:
Almost done on this story.  Will hopefully submit a PR tomorrow.

OwenLindsell:
Fixed

PR #457, opened 2019-09-06
GitHub#454 Sanitise HTTP message logs

OwenLindsell:
Added two new configuration options to hide headers and cookies in request logging:

```
    request-logging:
      hideHeaders:
          - Content-Type
      hideCookies:
          - sessionID
          - samlToken
```
The hideHeaders and hideCookies options take a list of header or cookie names. Any header or cookie in these lists will be obfuscated in the logged message output e.g.
```
headers=[Content-Type:****, Cookie:sessionID=****;samlToken=****]
```


components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -158,7 +159,7 @@ private static Origin originFromRequest(LiveHttpRequest request, Boolean isHttps
                 .orElseGet(() -> {
                     checkArgument(request.url().isAbsolute(), "host header is not set for request=%s", request);
                     return request.url().authority().map(Url.Authority::hostAndPort)
-                            .orElseThrow(() -> new IllegalArgumentException("Cannot send request " + request + " as URL is not absolute and no HOST header is present"));
+                            .orElseThrow(() -> new IllegalArgumentException("Cannot send request " + formatRequest(request) + " as URL is not absolute and no HOST header is present"));
```
mikkokar: Hi. StyxHttpClient doesn't really need to format the headers. Just the bare `.toString` is enough. 

The exception will eventually be processed by `HttpErrorStatusCauseLogger`. Let's leave the full message formatting there.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/HttpRequestOperation.java
```diff
@@ -311,7 +312,7 @@ public void onNext(ByteBuf chunk) {
                         } else {
                             String channelIdentifier = String.format("%s -> %s", nettyConnection.channel().localAddress(), nettyConnection.channel().remoteAddress());
                             LOGGER.error(format("Failed to send request body data. origin=%s connection=%s request=%s",
-                                    nettyConnection.getOrigin(), channelIdentifier, request), future.cause());
+                                    nettyConnection.getOrigin(), channelIdentifier, formatRequest(request)), future.cause());
```
mikkokar: No need to log the headers here. The exception will eventually propagate to `HttpErrorStatusCauseLogger` which alone can format the request headers.

components/common/src/main/java/com/hotels/styx/common/format/HttpHeaderFormatter.java
```diff
@@ -0,0 +1,89 @@
...
+        headerValuesToHide.replaceAll(String::toLowerCase);
+        cookieValuesToHide.replaceAll(String::toLowerCase);
+        instance = new HttpHeaderFormatter(headerValuesToHide, cookieValuesToHide);
+    }
+
+    public static HttpHeaderFormatter instance() {
+        return instance == null
+                ? new HttpHeaderFormatter(emptyList(), emptyList())
+                : instance;
+    }
```
mikkokar: As discussed, let's remove the singleton. It is going to be limiting in its applicability. 

Also since we will remove the header formatting everywhere else but `HttpErrorStatusCauseLogger` and `LoggingInterceptor`, it should be easy enough to inject the formatter into them.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
dvlato: I slightly disagree here, as if we don't format the request it won't show any headers (so we cannot really verify that there's no HOST header). It doesn't seem so important now, but In case we implemented HTTP/2, the :authority header could convey the same information so it would be good to print all 'displayable' headers.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -161,6 +164,8 @@ public StyxServer(StyxServerComponents config) {
     public StyxServer(StyxServerComponents components, Stopwatch stopwatch) {
         this.stopwatch = stopwatch;
 
+        initialiseHttpHeaderSanitiser(components.environment().configuration());
```
dvlato: I wouldn't initialise the HeaderSanitiser here as this method would probably not be invoked by users starting Styx programmatically.

components/server/src/test/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoderTest.java
```diff
@@ -332,7 +333,9 @@ private FullHttpResponse send(HttpRequest request) {
     }
 
     private void assertThatHttpHeadersAreSame(Iterable<HttpHeader> headers, HttpHeaders headers1) {
-        assertThat(newArrayList(headers).toString(), is(newArrayList(headers1).toString()));
+        for (HttpHeader header : headers) {
```
dvlato: This doesn't verify that headers1 is the same size as headers.

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
```diff
@@ -349,8 +349,7 @@ public String toString() {
         return toStringHelper(this)
                 .add("version", version)
                 .add("method", method)
-                .add("uri", url)
-                .add("headers", headers)
+                .add("url", url)
```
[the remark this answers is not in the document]
OwenLindsell: It was renamed because we convert it from a URI to a URL in the code (line 421 of HttpRequest) and the underlying object is a URL object.  Still as David and I discussed, it is a fairly trivial issue.  I've no problem with changing it back.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -189,6 +194,13 @@ public StyxServer(StyxServerComponents components, Stopwatch stopwatch) {
         });
     }
 
+    private void initialiseHttpHeaderSanitiser(StyxConfig config) {
+        List<String> headerValuesToHide = config.get("request-logging.hideHeaders", List.class).orElse(Collections.emptyList());
+        List<String> cookieValuesToHide = config.get("request-logging.hideCookies", List.class).orElse(Collections.emptyList());
```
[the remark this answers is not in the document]
OwenLindsell: fixed

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
OwenLindsell: fixed - removed formatting from all places apart from HttpErrorStatusCauseLogger and HttpRequestMessageLogger

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/HttpRequestOperation.java
OwenLindsell: fixed

components/common/src/main/java/com/hotels/styx/common/format/HttpHeaderFormatter.java
OwenLindsell: fixed

components/api/src/main/java/com/hotels/styx/api/HttpHeader.java
```diff
@@ -98,6 +98,6 @@ public boolean equals(Object o) {
 
     @Override
     public String toString() {
-        return name + "=" + HEADER_VALUES_JOINER.join(values);
+        return name + ":" + HEADER_VALUES_JOINER.join(values);
```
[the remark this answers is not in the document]
OwenLindsell: yes, I didn't consider that. I've no problem with changing it back if needed.

components/api/src/main/java/com/hotels/styx/api/LiveHttpRequest.java
```diff
@@ -388,8 +388,7 @@ public String toString() {
         return toStringHelper(this)
                 .add("version", version)
                 .add("method", method)
-                .add("uri", url)
-                .add("headers", headers)
+                .add("url", url)
```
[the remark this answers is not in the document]
OwenLindsell: see my reply above

components/api/src/test/java/com/hotels/styx/api/HttpHeaderTest.java
```diff
@@ -31,15 +31,15 @@ public void rejectsEmptyValuesList() {
...
+        assertThat(header.toString(), is("name:value"));
         assertThat(header.value(), is("value"));
         assertThat(header.values(), contains("value"));
     }
 
     @Test
     public void createsMultipleValueHeader() {
         HttpHeader header = header("name", "value1", "value2");
-        assertThat(header.toString(), is("name=value1, value2"));
+        assertThat(header.toString(), is("name:value1, value2"));
```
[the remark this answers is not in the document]
OwenLindsell: see my reply above

components/api/src/main/java/com/hotels/styx/api/HttpHeader.java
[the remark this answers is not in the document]
OwenLindsell: the aim was to avoid this...
Cookie=cookie=value;
HTTP protocol writes it like this:
Cookie: cookie=value;

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
OwenLindsell: fixed, it's now in Environment

components/server/src/test/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoderTest.java
OwenLindsell: fixed

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -158,7 +159,7 @@ private static Origin originFromRequest(LiveHttpRequest request, Boolean isHttps
                 .orElseGet(() -> {
                     checkArgument(request.url().isAbsolute(), "host header is not set for request=%s", request);
```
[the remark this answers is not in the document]
OwenLindsell: I agree with mikko that we should only format in the error logger and the request logger

mikkokar:
Fixes #454.

components/proxy/src/main/java/com/hotels/styx/Environment.java
```diff
@@ -125,6 +134,11 @@ public Builder eventBus(EventBus eventBus) {
             return this;
         }
 
+        public Builder httpMessageFormatter(HttpMessageFormatter httpMessageFormatter) {
+            this.httpMessageFormatter = httpMessageFormatter;
```
mikkokar: Please add a null check.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -97,8 +97,10 @@ private static StyxServer createStyxServer(String[] args) {
         LOG.info("Styx configFileLocation={}", startupConfig.configFileLocation());
         LOG.info("Styx logConfigLocation={}", startupConfig.logConfigLocation());
 
+        StyxConfig styxConfig = parseConfiguration(startupConfig);
+
         StyxServerComponents components = new StyxServerComponents.Builder()
-                .styxConfig(parseConfiguration(startupConfig))
+                .styxConfig(styxConfig)
                 .startupConfig(startupConfig)
```
mikkokar: Is this just a cosmetic change? Please consider backing this off to keep the change history clean.

components/proxy/src/main/java/com/hotels/styx/proxy/BackendServicesRouter.java
```diff
@@ -178,10 +180,11 @@ private OriginHealthStatusMonitor healthStatusMonitor(BackendService backendServ
                                 healthCheckClient);
     }
 
-    private StyxHttpClient healthCheckClient(BackendService backendService) {
+    private StyxHttpClient healthCheckClient(BackendService backendService, HttpMessageFormatter httpMessageFormatter) {
         StyxHttpClient.Builder builder = new StyxHttpClient.Builder()
                 .connectTimeout(backendService.connectionPoolConfig().connectTimeoutMillis(), MILLISECONDS)
-                .userAgent("Styx/" + environment.buildInfo().releaseVersion());
+                .userAgent("Styx/" + environment.buildInfo().releaseVersion())
+                .httpMessageFormatter(httpMessageFormatter);
```
mikkokar: Health check client (functionality) doesn't really need message formatting. Of course there could be a technical reason why this is passed in here. I didn't check.

components/proxy/src/main/java/com/hotels/styx/proxy/HttpErrorStatusCauseLogger.java
```diff
@@ -29,7 +30,13 @@
  * Wrapper for {@link HttpErrorStatusListener} that also logs {@link Throwable}s.
  */
 public class HttpErrorStatusCauseLogger implements HttpErrorStatusListener {
+
     private static final Logger LOG = getLogger(HttpErrorStatusCauseLogger.class);
+    private HttpMessageFormatter formatter;
+
+    public HttpErrorStatusCauseLogger(HttpMessageFormatter formatter) {
+        this.formatter = formatter;
+    }
```
mikkokar: Formatter could be `final`, and the constructor would need a null check.

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
```diff
@@ -262,6 +274,11 @@ public Builder loggingSetUp(LoggingSetUp loggingSetUp) {
             return this;
         }
 
+        public Builder httpMessageFormatter(HttpMessageFormatter httpMessageFormatter) {
+            this.httpMessageFormatter = httpMessageFormatter;
+            return this;
+        }
```
mikkokar: This presumably is unused and needs to be removed?

components/proxy/src/main/java/com/hotels/styx/proxy/BackendServicesRouter.java
OwenLindsell: This is needed because the health check client constructs an instance of HttpRequestOperation which in turn constructs an instance of HttpRequestMessageLogger.  This is old code, but I think we should consider injecting HttpRequestMessageLogger instead of constructing it in two places.  Ultimately all 'new' operations should be performed early and we should try(where possible) to construct our objects all in one place.

components/client/src/main/java/com/hotels/styx/client/StyxHttpClient.java
```diff
@@ -310,6 +312,11 @@ Builder copy() {
             return new Builder(this);
         }
 
+        public Builder httpMessageFormatter(HttpMessageFormatter httpMessageFormatter) {
+            this.httpMessageFormatter = httpMessageFormatter;
+            return this;
+        }
```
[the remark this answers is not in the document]
OwenLindsell: Fixed by providing a default in HttpRequestOperationFactoryBuilder

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/HttpRequestOperation.java
```diff
@@ -91,12 +92,12 @@ public HttpRequestOperation(LiveHttpRequest request, OriginStatsFactory originSt
...
      */
     public HttpRequestOperation(LiveHttpRequest request, OriginStatsFactory originStatsFactory,
-                                int responseTimeoutMillis, boolean requestLoggingEnabled, boolean longFormat) {
+                                int responseTimeoutMillis, boolean requestLoggingEnabled, boolean longFormat, HttpMessageFormatter httpMessageFormatter) {
         this.request = requireNonNull(request);
         this.originStatsFactory = Optional.ofNullable(originStatsFactory);
         this.responseTimeoutMillis = responseTimeoutMillis;
         this.requestLoggingEnabled = requestLoggingEnabled;
-        this.httpRequestMessageLogger = new HttpRequestMessageLogger("com.hotels.styx.http-messages.outbound", longFormat);
+        this.httpRequestMessageLogger = new HttpRequestMessageLogger("com.hotels.styx.http-messages.outbound", longFormat, httpMessageFormatter);
```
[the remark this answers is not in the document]
OwenLindsell: 👍

components/proxy/src/main/java/com/hotels/styx/proxy/HttpErrorStatusCauseLogger.java
OwenLindsell: fixed

components/api/src/main/java/com/hotels/styx/api/HttpRequest.java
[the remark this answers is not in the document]
OwenLindsell: reverted to url

components/api/src/main/java/com/hotels/styx/api/HttpHeader.java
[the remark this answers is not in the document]
OwenLindsell: reverted to use =

components/common/src/main/java/com/hotels/styx/common/format/HttpHeaderFormatter.java
```diff
@@ -0,0 +1,87 @@
...
+        return shouldHideCookie(cookie)
+                ? cookie.name() + "=****"
+                : cookie.toString();
+    }
+
+    private boolean shouldHideCookie(RequestCookie cookie) {
+        return cookieValuesToHide.stream()
+                .anyMatch(h -> h.equalsIgnoreCase(cookie.name()));
+    }
+
```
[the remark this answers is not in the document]
OwenLindsell: fixed

components/api/src/main/java/com/hotels/styx/api/LiveHttpRequest.java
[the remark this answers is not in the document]
OwenLindsell: reverted to uri

components/api/src/test/java/com/hotels/styx/api/HttpHeaderTest.java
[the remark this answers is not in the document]
OwenLindsell: reverted to =

components/api/src/main/java/com/hotels/styx/api/LiveHttpRequest.java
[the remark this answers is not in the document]
OwenLindsell: reverted to uri

components/common/src/main/java/com/hotels/styx/common/format/HttpHeaderFormatter.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.stream.StreamSupport;
+
+import static java.util.Objects.requireNonNull;
+
+public class HttpHeaderFormatter {
+
+    private static final List<String> COOKIE_HEADER_NAMES = Arrays.asList("cookie", "set-cookie");
+
+    private List<String> headerValuesToHide;
+    private List<String> cookieValuesToHide;
```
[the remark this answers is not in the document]
OwenLindsell: fixed

components/proxy/src/main/java/com/hotels/styx/Environment.java
OwenLindsell: fixed

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
OwenLindsell: reverted

components/proxy/src/main/java/com/hotels/styx/proxy/BackendServicesRouter.java
OwenLindsell: reverted as no longer needed as we have a default now

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
OwenLindsell: removed

components/common/src/main/java/com/hotels/styx/common/format/HttpHeaderFormatter.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private String hideOrFormatCookie(RequestCookie cookie) {
+        return shouldHideCookie(cookie)
+                ? cookie.name() + "=****"
+                : cookie.toString();
+    }
+
+    private boolean shouldHideCookie(RequestCookie cookie) {
+        return cookieValuesToHide.stream()
+                .anyMatch(h -> h.equalsIgnoreCase(cookie.name()));
```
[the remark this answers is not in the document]
OwenLindsell: fixed

components/common/src/main/java/com/hotels/styx/common/format/HttpMessageFormatter.java
```diff
@@ -0,0 +1,92 @@
...
+import com.hotels.styx.api.HttpResponse;
+import com.hotels.styx.api.HttpResponseStatus;
+import com.hotels.styx.api.HttpVersion;
+import com.hotels.styx.api.LiveHttpRequest;
+import com.hotels.styx.api.LiveHttpResponse;
+import com.hotels.styx.api.Url;
+
+public class HttpMessageFormatter {
+    
+    private HttpHeaderFormatter httpHeaderFormatter;
```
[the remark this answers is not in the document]
OwenLindsell: fixed

components/common/src/main/java/com/hotels/styx/common/format/HttpMessageFormatter.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public HttpMessageFormatter(HttpHeaderFormatter httpHeaderFormatter) {
+        this.httpHeaderFormatter = httpHeaderFormatter;
```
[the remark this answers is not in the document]
OwenLindsell: fixed

components/common/src/main/java/com/hotels/styx/common/logging/HttpRequestMessageLogger.java
```diff
@@ -26,93 +27,59 @@
...
+
     private final Logger logger;
     private final boolean longFormatEnabled;
+    private final HttpMessageFormatter httpMessageFormatter;
 
-    public HttpRequestMessageLogger(String name, boolean longFormatEnabled) {
+    public HttpRequestMessageLogger(String name, boolean longFormatEnabled, HttpMessageFormatter httpMessageFormatter) {
         this.longFormatEnabled = longFormatEnabled;
+        this.httpMessageFormatter = httpMessageFormatter;
         logger = getLogger(name);
```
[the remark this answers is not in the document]
OwenLindsell: fixed

components/common/src/test/java/com/hotels/styx/common/format/HttpMessageFormatterTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.testng.Assert.assertTrue;
+
+public class HttpMessageFormatterTest {
+
+    private static final HttpRequest httpRequest = new HttpRequest.Builder()
+            .version(HTTP_1_1)
+            .header("HeaderName", "HeaderValue")
+            .build();
+
+    private static final HttpResponse httpResponse = new HttpResponse.Builder()
```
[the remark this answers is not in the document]
OwenLindsell: fixed

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/HttpOutboundMessageLoggingSpec.scala
```diff
@@ -112,10 +112,10 @@ class HttpOutboundMessageLoggingSpec extends FunSpec
         assertThat(logger.log.size(), is(2))
 
         assertThat(logger.log(), hasItem(loggingEvent(INFO,
-          "requestId=[-a-z0-9]+, request=\\{method=GET, uri=http://localhost:[0-9]+/foobar, origin=\"localhost:[0-9]+\", headers=\\[.*\\]}")))
+          "requestId=[-a-z0-9]+, request=LiveHttpRequest\\{version=HTTP/1.1, method=GET, uri=http://localhost:[0-9]+/foobar, headers=\\[.*\\], id=[-a-z0-9]+}, origin=appOne:generic-app-01:localhost:[0-9]+")))
```
mikkokar: I see we are now exposing class names (implementation detail) in our access logs (a documented feature). 

Let's remove them from the access logs. The class names are okay for error logs etc but not here. As a documented feature we need to consider the access logs like any other external interface with possible backwards compatibility implications. 

Therefore please remove the class name from this message.

Also, I'd move the `origin` either in its original place, or in front of the request message. My reasoning is:

1) Trying to keep the message format as similar as possible 

2) Headers field is often really large, and therefore should appear after all "fixed size" attributes. This is because it is difficult to find and see the fixed size attributes after a large body of header text.

components/proxy/src/main/java/com/hotels/styx/proxy/BackendServicesRouter.java
```diff
@@ -131,30 +131,24 @@ public void onChange(Registry.Changes<BackendService> changes) {
...
 
-            StyxHttpClient healthCheckClient = healthCheckClient(backendService);
-
-            OriginHealthStatusMonitor healthStatusMonitor = healthStatusMonitor(backendService, healthCheckClient);
-
-            StyxHostHttpClient.Factory hostClientFactory = (ConnectionPool connectionPool) -> {
-                StyxHeaderConfig headerConfig = environment.styxConfig().styxHeaderConfig();
-                return StyxHostHttpClient.create(connectionPool);
-            };
+            OriginHealthStatusMonitor healthStatusMonitor = healthStatusMonitor(backendService);
```
mikkokar: Good refactoring here in this file. Thanks :-D

components/common/src/main/java/com/hotels/styx/common/format/DefaultHttpMessageFormatter.java
```diff
@@ -0,0 +1,44 @@
...
+ * Provides default formatting for requests and responses.
+ */
+public class DefaultHttpMessageFormatter implements HttpMessageFormatter {
+
+    public String formatRequest(HttpRequest request) {
+        return request == null ? null : request.toString();
+    }
+
+    public String formatRequest(LiveHttpRequest request) {
+        return request == null ? null : request.toString();
```
mikkokar: This code is okay for this PR. 

However it might be safer to return an empty string, or "NA" etc instead of bare `null` when request is not provided. What do you think?

components/proxy/src/main/java/com/hotels/styx/proxy/BackendServicesRouter.java
OwenLindsell: np

components/common/src/main/java/com/hotels/styx/common/format/DefaultHttpMessageFormatter.java
OwenLindsell: I wasn't sure what circumstances this would occur in (if ever?), so I left the behaviour as already was.  It's no problem changing it though even in this PR.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/HttpOutboundMessageLoggingSpec.scala
OwenLindsell: The origin isn't part of the request, so I moved it out.  Good point about long headers field.  I'll move it to in front.

components/common/src/main/java/com/hotels/styx/common/format/DefaultHttpMessageFormatter.java
mikkokar: Ok

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/HttpOutboundMessageLoggingSpec.scala
mikkokar: > The origin isn't part of the request, so I moved it out.

Yeah, it was a right thing to do.

components/common/src/main/java/com/hotels/styx/common/format/DefaultHttpMessageFormatter.java
OwenLindsell: changed to N/A

system-tests/e2e-suite/src/test/scala/com/hotels/styx/proxy/HttpOutboundMessageLoggingSpec.scala
OwenLindsell: moved to front

components/common/src/main/java/com/hotels/styx/common/format/DefaultHttpMessageFormatter.java
OwenLindsell: Now changed to string "null" instead of "N/A".  It's more meaningful and is consistent with what we have in HttpRequestMessageLogger.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/StaticResponseHandler.java
```diff
@@ -83,7 +111,20 @@ public RoutingObject build(List<String> fullName, Context context, StyxObjectDef
             StaticResponseConfig config = new JsonNodeConfig(configBlock.config())
                     .as(StaticResponseConfig.class);
 
-            return new StaticResponseHandler(config.status, config.response);
+            HttpHeaders nettyHeaders = buildNettyHeaders(config);
+            return new StaticResponseHandler(config.status, config.response, nettyHeaders);
+        }
+
+        private HttpHeaders buildNettyHeaders(StaticResponseConfig config) {
```
mikkokar: A small comment: "Netty headers".

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/logging/HttpMessageLoggingSpec.kt
```diff
@@ -0,0 +1,118 @@
...
+import io.kotlintest.specs.FeatureSpec
+import reactor.core.publisher.toMono
+
+class HttpMessageLoggingSpec : FeatureSpec() {
+
+    init {
+        feature("Styx request/response logging") {
+
+            scenario("Logger should hide cookies and headers") {
+                val proxyHost = "${styxServer.proxyHttpAddress().hostName}:${styxServer.proxyHttpAddress().port}"
```
mikkokar: Use `StyxServer.proxyHttpHostHeader ` extension method from support/styxServerProvider.kt to simplify `proxyHost`.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/logging/HttpMessageLoggingSpec.kt
```diff
@@ -0,0 +1,118 @@
...
+
+                client.send(HttpRequest.get("/a/path")
+                        .header(HttpHeaderNames.HOST, proxyHost)
```
mikkokar: Static imports x2 ^^^.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/logging/HttpMessageLoggingSpec.kt
```diff
@@ -0,0 +1,118 @@
...
+                  value: "h2" 
+                - name: "cookie"
+                  value: "cookie1=c1;cookie2=c2"
+
+        httpPipeline: root
+      """.trimIndent()
+
+    val styxServer = StyxServer(StyxServerComponents.Builder()
+            .styxConfig(StyxConfig.fromYaml(yamlText))
+            .build())
```
mikkokar: There is a class called `StyxServerProvider` in support/StyxServerProvider.kt. It provides a unique reference point to a Styx server while allowing it to be restarted, and thus cleared from any acquired state`, between the tests.

You don't really have to use it in this test, but could make this more consistent with others.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/logging/HttpMessageLoggingSpec.kt
```diff
@@ -0,0 +1,118 @@
...
+                val proxyHost = "${styxServer.proxyHttpAddress().hostName}:${styxServer.proxyHttpAddress().port}"
+
+                client.send(HttpRequest.get("/a/path")
+                        .header(HttpHeaderNames.HOST, proxyHost)
+                        .header("header1", "h1")
+                        .header("header2", "h2")
+                        .header("cookie", "cookie1=c1;cookie2=c2")
+                        .build())
+                        .toMono()
+                        .block()!!
```
mikkokar: There is also a `wait` extension method that combines `.toMono().block()!!`.

PR #474, opened 2019-10-08
Make admin interfaces streamable

taer:
This should allow plugins to stream requests larger than 1M


components/server/src/main/java/com/hotels/styx/server/AdminHttpRouter.java
```diff
@@ -0,0 +1,54 @@
...
+import com.hotels.styx.api.WebServiceHandler;
+import com.hotels.styx.common.http.handler.HttpAggregator;
+
+import static com.hotels.styx.api.HttpResponse.response;
+import static com.hotels.styx.api.HttpResponseStatus.NOT_FOUND;
+
+/**
+ * Simple Http Router.
+ */
+public class AdminHttpRouter implements HttpHandler {
```
mikkokar: Hi @taer Thank you for doing this! Very useful. But one comment. 

Please move this class to `styx-proxy` module. This class is called *Admin*HttpRouter, and you are adding it to the `styx-server` module which doesn't have an admin interface. The admin interface is in the `styx-proxy`. Therefore it should live in the proxy module. But otherwise it good to go :-D.

taer: done. I think the test failure is a timeout based test

PR #475, opened 2019-10-10
Issue 466: Expose plugins for HttpInterceptorPipeline

mikkokar:
## Description
This PR tweaks plugins loading so that the plugins will be available for `InterceptorPipeline` object.

Specifically, when started as a *stand-alone application*:

- `plugins.active` server configuration attribute is now optional. It can be omitted when styx is started with the routing object engine.
- `httpPipeline` attribute activates the new routing object engine over the current application router. In this case all plugins are loaded from `plugins.all` section.

When started *programmatically*:

- Can be started only in the old application router mode.
- All plugin factories passed to `StyxServerComponents.Builder.pluginFactories()` are loaded into the pipeline, in the order of appearance.


system-tests/docker-toxiproxy/README.md
```diff
@@ -90,3 +90,11 @@ To enable/disable origins:
 ```bash
 $ toxiproxy-cli toggle httpd-01
 ```
+
+
+## Debugging
```
dvlato: I thought these changes were merged here: https://github.com/HotelsDotCom/styx/pull/470/files . Do you need to rebase the PR?

mikkokar: Looks like I've accidentally based this work on that PR's branch. But git claims it can merge this PR regardless.

PR #478, opened 2019-10-11
Remove hamcrest dependency from styx common production code.

mikkokar:
- Reimplement MorePreconditions without Hamcrest.
- Remove TransformingMatcher.
- Update Jackson databank used in plugin/examples to 2.9.10.

PR #480, opened 2019-10-11
Remove use of hamcrest in production (non-test) code.

dvlato:


support/testsupport/src/main/java/com/hotels/styx/support/matchers/TransformingMatcher.java
```diff
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2018 Expedia Inc.
```
mikkokar: Let's get rid of this class. You can just copy the changes from my closed PR: https://github.com/HotelsDotCom/styx/pull/478.

PR #481, opened 2019-10-11
Update jackson/databind module in plugin/examples to 2.9.10.

mikkokar:
As per security review.


dvlato:
This PR is not needed anymore - plugin examples is inheriting the versions from styx-bom .

PR #492, opened 2019-10-21
Move backendServiceFactory from default configuration into styx config.

mikkokar:
The old backend services factory is backed in the `default.yaml` in Styx Docker base image. 

This PR moves it down to `docker-test-env/styx-config/styxconf.yml` that is mapped from the host, making it easier to swap between the old application router and the new routing object engine.

### DIFF
--- a/system-tests/docker-test-env/styx-config/styxconf.yml
+++ b/system-tests/docker-test-env/styx-config/styxconf.yml
@@ -4,5 +4,8 @@ include: /styx/default-config/default.yml
 services:
   factories:
     backendServiceRegistry:
+      class: "com.hotels.styx.proxy.backends.file.FileBackedBackendServicesRegistry$Factory"
       config:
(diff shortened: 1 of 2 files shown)


system-tests/docker-test-env/styx-config/styxconf.yml
```diff
@@ -4,5 +4,8 @@ include: /styx/default-config/default.yml
 services:
   factories:
     backendServiceRegistry:
+      class: "com.hotels.styx.proxy.backends.file.FileBackedBackendServicesRegistry$Factory"
       config:
-        originsFile: "/styx/config/origins.yml"
+        originsFile: "/styx/default-config/origins.yml"
```
chrisgresty: Should this be /styx/config/origins.yml as before?

mikkokar: Fixed.

PR #499, opened 2019-10-23
Styx docker environment: add NGINX origins

mikkokar:
Replaces Apache httpd with NGINX based origins.


system-tests/docker-test-env/docker-compose.yml
```diff
@@ -34,9 +35,6 @@ services:
     volumes:
       - ./styx-config:/styx/config
       - ./styx-logs:/styx/logs
-    links:
```
dvlato: About time :)

system-tests/docker-test-env/docker-compose.yml
```diff
@@ -1,17 +1,18 @@
 version: '3'
 services:
-
-  httpd-01:
-    image: httpd:2.4
+  origin:
+    image: nginx:stable-alpine
     ports:
-      - "8081:80"
+      - "9091:8080"
```
dvlato: I would rather start the nginx listener in the same ports we are going to expose externally.

system-tests/docker-test-env/nginx/testCredentials.crt
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
dvlato: This certification expired on 2015.
Why are we declaring a HTTPS port for Nginx if are not using it in our origin configuration?

system-tests/docker-test-env/docker-compose.yml
mikkokar: Hi @dvlato 

I'd like to keep all http services behind the same port (8080). It is easier to configure all of styx, toxi-proxy, and nginx etc when the port is always the same. The ports won't conflict inside docker network: each service will have its own host name.

Therefore the ports are exposed only for the debugging purposes, and on the host hey need to be unique.

Therefore I'd keep it as it is.

system-tests/docker-test-env/nginx/testCredentials.crt
mikkokar: For now this only servers as a configuration example for enabling TLS on NGINX. I couldn't get this to work without a certificate, and this old, self-signed test certificate appears to do just fine.

I could add this to the origins configuration if that helps?

mikkokar: I have now added a secure origin.

dvlato: Maybe we should add toxyproxy config for that secure origin as well.

PR #503, opened 2019-10-24
Origins dashboard, initial design

mikkokar:
## Description

Implement a dashboard that shows, for each application:

- The path prefix mapping, and the application name
- Information for each origin:
   - Availability (active/inactive)
   - TLS encrypted or not (http/https)
   - Origin name and hostname
   - The remaining tags, excluding the app membership, and status tags. The app membership and the status can be inferred without showing tags.

This PR also:
- Adds a kotlin.html dependency
- Bundles **materialize** css library in styx (Licensed under MIT license).
- ClassPathResourceHandler: configurable asset path and request prefix.
- Modify admin endpoint handlers for YamlFileConfigurationService:
  - `assets/`. - for serving static assets
  - `configuration/` - for origins file
  - `origins/` - this new dashboard

## Example page

screenshot


components/proxy/src/main/kotlin/com/hotels/styx/ObjectTags.kt
```diff
@@ -16,3 +16,10 @@
 package com.hotels.styx
 
 fun lbGroupTag(name: String) = "lbGroup=$name"
+
+fun matchLbGroupTag(tag: String): String? = "lbGroup=(.+)".toRegex()
```
chrisgresty: Maybe call this `lbGroupTagValue`, as that's what it is returning? (regex matching is just the mechanism).

components/server/src/main/java/com/hotels/styx/server/handlers/ClassPathResourceHandler.java
```diff
@@ -33,32 +34,54 @@
...
     private final String root;
 
     public ClassPathResourceHandler(String root) {
-        this.root = ensureRootEndsInSlash(root);
+        this(root, root);
     }
 
-    private static String ensureRootEndsInSlash(String root) {
-        return root.endsWith("/") ? root : root + "/";
+    public ClassPathResourceHandler(String allowedPrefix, String root) {
```
chrisgresty: Perhaps call these parameters `requestPrefix` and `resourceRoot`? To make it a little clearer what the strings get applied to/checked against.

components/api/src/main/java/com/hotels/styx/api/extension/service/spi/StyxService.java
```diff
@@ -66,11 +66,30 @@
...
+
+    /**
+     * An admin interface hook for the service implementation.
+     * The HTTP handler endpoints are prefixed with {@code baseUrlPath}.
+     *
+     * @param baseUrlPath A base path for the admin interface handlers.
+     *
+     * @return Returns a list of named admin interface handlers.
+     */
+    default Map<String, HttpHandler> adminInterfaceHandlers(String baseUrlPath) {
```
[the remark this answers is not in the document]
mikkokar: Please have another look. 

The description feels even more baffling now, but perhaps we can together work out something better?

PR #504, opened 2019-10-25
GitHub#486 - Fix connection pool leaks

OwenLindsell:


components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -77,10 +77,12 @@ public Origin getOrigin() {
             Connection connection = dequeue();
             if (connection != null) {
                 borrowedCount.incrementAndGet();
+                sink.onCancel(() -> availableConnections.add(connection));
```
dvlato: Shouldn't we also reduce the borrowedCount as well?

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -77,10 +77,12 @@ public Origin getOrigin() {
...
                 sink.success(connection);
             } else {
                 if (waitingSubscribers.size() < poolSettings.maxPendingConnectionsPerHost()) {
-                    this.waitingSubscribers.add(sink.onDispose(() -> waitingSubscribers.remove(sink)));
+                    this.waitingSubscribers.add(sink);
+                    sink.onDispose(() -> waitingSubscribers.remove(sink));
```
dvlato: I think this change can be reverted. However, I thought we had an issue with the waitingSubscribers list and we haven't changed anything in that code path... Don't we need to change anything here?

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -143,6 +145,7 @@ private void queueNewConnection(Connection connection) {
             availableConnections.add(connection);
         } else {
             borrowedCount.incrementAndGet();
+            subscriber.onCancel(() -> availableConnections.add(connection));
```
dvlato: Same comment here about borrowedCount.

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
OwenLindsell: Yes, very true.  I will fix!

OwenLindsell: Fixed

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
OwenLindsell: Fixed

Issue #494, opened 2019-10-23
Expose styx service providers in admin interface


mikkokar:
## Motivation

As a styx user, I'd like inspect configured styx service providers via admin interface, so that I verify the health check provider configuration in styx backwards compatibility mode.

## Background

Styx backwards compatibility mode for origins file (PR #458) automatically creates health check services for the balancing groups. 

Although the service configuration strictly follows the origins file, it is still important to expose the actual service provider state via the admin interface:
 
* To check the running, non-configured state
* To ensure that correct configuration is applied
* To help investigating faults

## Implementation Considerations

All styx service providers go into `providerObjectStore` database in `StyxServerComponents`. Don't restrict the admin endpoints to the health check providers alone. Show any/all `StyxService` stored there.

## Acceptance Criteria

* `/admin/service/providers` -endpoint lists all service providers stored in the provider object database (see above)
* `/admin/service/provider/<name>` returns the details for the named service provider only.

mikkokar:
Fixed by #505

PR #505, opened 2019-10-25
Expose styx service providers in admin interface

chrisgresty:
See Issue #494


components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+        StyxObjectStore<ProviderObjectRecord> store = createTestStore();
+        ServiceProviderHandler handler = new ServiceProviderHandler(store);
+        HttpRequest request = HttpRequest.get("/admin/service/provider/object2").build();
+        HttpResponse response = Mono.from(handler.handle(request, HttpInterceptorContext.create())).block();
+
+        assertThat(response.status(), equalTo(HttpResponseStatus.OK));
+
+        StyxObjectDefinition actualProvider = deserialiseProvider(response.bodyAs(StandardCharsets.UTF_8));
+        assertThat(actualProvider.name(), equalTo("object2"));
+        assertThat(actualProvider, notNullValue());
```
dvlato: should we verify it's not null before calling .name() instead of the other way around?

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+        ServiceProviderHandler handler = new ServiceProviderHandler(empty);
+        HttpRequest request = HttpRequest.get("/not/my/url").build();
+        HttpResponse response = Mono.from(handler.handle(request, HttpInterceptorContext.create())).block();
+
+        assertThat(response.status(), equalTo(HttpResponseStatus.NOT_FOUND));
+    }
+
+    private StyxObjectStore<ProviderObjectRecord> createTestStore() throws IOException {
+        ObjectMapper mapper =new ObjectMapper();
+        StyxService mockService = Mockito.mock(StyxService.class);
```
dvlato: you can probably add a static import for this.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+        StyxObjectStore<ProviderObjectRecord> empty = new StyxObjectStore<>();
+        ServiceProviderHandler handler = new ServiceProviderHandler(empty);
+        HttpRequest request = HttpRequest.get("/not/my/url").build();
+        HttpResponse response = Mono.from(handler.handle(request, HttpInterceptorContext.create())).block();
+
+        assertThat(response.status(), equalTo(HttpResponseStatus.NOT_FOUND));
+    }
+
+    private StyxObjectStore<ProviderObjectRecord> createTestStore() throws IOException {
+        ObjectMapper mapper =new ObjectMapper();
```
dvlato: I don't know if this is going to pass stylistic checks in checkstyle or sonar....  Add a whitespace before new?

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/admin/OriginsFileCompatibilitySpec.kt
```diff
@@ -484,8 +534,30 @@ class OriginsFileCompatibilitySpec : FunSpec() {
                             }
                 }
             }
+
+            test("Health checking service no longer appears in the /admin/service/providers endpoint") {
```
dvlato: Probably we don't need /admin/service/providers (and /admin/service/provider/appB below and above) in the test name but something a bit more meaningful for the reader.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+    private List<StyxObjectDefinition> extractProviders(String responseBody) throws IOException {
+        return Arrays.stream(responseBody.split("---\n"))
+                .filter(StringUtils::isNotBlank)
+                .map(this::deserialiseProvider)
+                .collect(Collectors.toList());
+    }
+
+    private void validateProvider(StyxObjectDefinition actual, ProviderObjectRecord expected) {
+        assertThat(actual.type(), equalTo(expected.getType()));
+        assertThat(actual.tags(), containsInAnyOrder(expected.getTags().toArray()));
```
dvlato: I would also check that it doesn't contain any other elements (e.g., verify the length).

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ServiceProviderHandler.java
```diff
@@ -0,0 +1,127 @@
...
+                            .build());
+
+                })
+                .get("/admin/service/provider/:providerName", (request, context) -> {
+                    String name = placeholders(context).get("providerName");
+
+                    try {
+                        String object = providerDatabase.get(name)
+                                .map(record -> serialise(name, record))
+                                .orElseThrow(ResourceNotFoundException::new);
```
dvlato: I see we have several of these ResourceNotFoundException classes.   Should we return the response from the map() call instead of having to throw an exception and capture it?

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ServiceProviderHandler.java
```diff
@@ -0,0 +1,127 @@
...
+        return urlRouter.handle(request, context);
+    }
+
+    public static ObjectMapper yamlMapper() {
+        return YAML_MAPPER.copy();
+    }
+
+    private static String serialise(String name, ProviderObjectRecord record) {
+        JsonNode node = YAML_MAPPER
+                .valueToTree(new StyxObjectDefinition(name, record.getType(), ImmutableList.copyOf(record.getTags()), record.getConfig()));
```
dvlato: I don't know if the length of this line is going to be a problem anywhere...

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ServiceProviderHandler.java
```diff
@@ -0,0 +1,127 @@
...
+                .get("/admin/service/provider/:providerName", (request, context) -> {
+                    String name = placeholders(context).get("providerName");
+
+                    try {
+                        String object = providerDatabase.get(name)
+                                .map(record -> serialise(name, record))
+                                .orElseThrow(ResourceNotFoundException::new);
+
+                        return Eventual.of(response(OK).body(object, UTF_8).build());
+                    } catch (ResourceNotFoundException e) {
```
dvlato: This could throw a RuntimeException/ClassCastException as well. I don't think an Internal server error is the best error if they are requesting an object that it's not a provider.  I would say that's a 4xx error.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @Test
+    public void returnsNoContentStatusWhenNoProvidersAvailable() {
+        StyxObjectStore<ProviderObjectRecord> empty = new StyxObjectStore<>();
+        ServiceProviderHandler handler = new ServiceProviderHandler(empty);
+        HttpRequest request = HttpRequest.get("/admin/service/providers").build();
+        HttpResponse response = Mono.from(handler.handle(request, HttpInterceptorContext.create())).block();
+
+        assertThat(response.status(), equalTo(HttpResponseStatus.NO_CONTENT));
+        assertThat(response.contentLength().get(), equalTo(0L));
```
dvlato: I don't particularly like the Exception message when we call get() on an optional. Maybe we could have a nicer assertion message.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
chrisgresty: Yes, indeed.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
chrisgresty: OK

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
chrisgresty: Even if it does, it's still ugly. Thx.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
chrisgresty: I think `containsInAnyOrder` also requires the length to be equal.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
chrisgresty: Perhaps assert `isPresent()` beforehand?

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/admin/OriginsFileCompatibilitySpec.kt
chrisgresty: OK

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ServiceProviderHandler.java
chrisgresty: Good point. Thanks.
I'll log it here too, so the stack trace is captured.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ServiceProviderHandler.java
chrisgresty: There's a bit much going on anyway. Will rearrange.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ServiceProviderHandler.java
chrisgresty: Yes, much nicer.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
chrisgresty: Ah, actually it seems a NO_CONTENT response should not have a content-length header at all, even one set to 0. So I'll fix that, and check for its absence.
https://tools.ietf.org/html/rfc7230#section-3.3.2

> A server MUST NOT send a Content-Length header field in any response
   with a status code of 1xx (Informational) or 204 (No Content).

dvlato: I think we intentionally let it have  "Content-Length: 0" here as it seems otherwise the client was getting  'Transfer-Encoding: chunked' otherwise and that broke everything apart.  If I recall correctly, we thought the header was added by Netty but we didn't really verify it.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
dvlato: Oh right. The name always gets me confused.

components/proxy/src/test/java/com/hotels/styx/admin/handlers/ServiceProviderHandlerTest.java
chrisgresty: In a quick check of the running server, I didn't get a `Transfer-Encoding: chunked` header when a zero-content response was returned. Maybe the admin server's netty is set up differently to the main proxy server?

PR #513, opened 2019-11-01
Fix kotlin unit tests

chrisgresty:
Kotlin tests are run by a JUnit 5 test runner. The styx-proxy component has Kotlin unit tests as well as TestNG unit tests, but the Maven Surefire plugin does not, by default, run both JUnit and TestNG tests in the same project.

The surefire plugin needs to be configured with two executions - one each for JUnit5 and TestNG - and the default execution disabled. 

Surefire determines which test framework to use by examining the test classpath. If it finds the appropriate dependency ("junit:junit", "org.junit.platform:junit-platform-engine" or "org.testng:testng" for JUnit4, JUnit5 and TestNG respectively) it executes that framework.

The two custom executions defined here disable the unwanted frameworks for each case, leaving only one that will run. It does this by changing the name of the dependency artifacts that the surefire plugin is looking for. (i.e. by setting the TestNG artifact name to "none:none", surefire will not find that on the test classpath and so will not try to run TestNG tests in that execution.)

The config parameters used are `testNgArtifactName`, `junitArtifactName` and `junitPlatformArtifactName` - as listed here: http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html

The original inspiration for this approach came from here: https://issues.apache.org/jira/browse/SUREFIRE-377

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/StyxHeaderConfig.java
+++ b/components/client/src/main/java/com/hotels/styx/client/StyxHeaderConfig.java
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2018 Expedia Inc.
+  Copyright (C) 2013-2019 Expedia Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


chrisgresty:
> Please add a quick comment in the POM file explaining how the TestNG and JUnit5 activation works.

I added details in the PR

Issue #496, opened 2019-10-23
Expose Styx origins configuration file content in backwards compatibility mode


mikkokar:
The origins configuration (file content) is not exposed via the admin interface in the backwards compatibility mode (PR #458).

The contents are logged when the file is reloaded, but it would be better expose the actual configuration via the admin interface.

The file content should be exposed via this admin interface endpoint: `/admin/configuration/origins`

chrisgresty:
Addressed with PR #520

PR #520, opened 2019-11-06
Expose styx origins config file in Admin interface (Issue #496)

chrisgresty:
Issue #496 
Modify the YamlFileConfigurationService to publish an endpoint to the Admin interface, that returns the current origins configuration file.

### DIFF
--- a/components/proxy/src/main/kotlin/com/hotels/styx/services/YamlFileConfigurationService.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/services/YamlFileConfigurationService.kt
@@ -83,7 +83,7 @@ internal class YamlFileConfigurationService(
             }
 
     override fun adminInterfaceHandlers() =
-        mutableMapOf(Pair("origins", TextHttpHandler { originsConfig }))
+        mapOf("origins" to TextHttpHandler { originsConfig })
 
 
     fun reloadAction(content: String): Unit {
@@ -99,6 +99,7 @@ internal class YamlFileConfigurationService(
         }.mapCatching { (healthMonitors, routingObjectDefs) ->
             updateRoutingObjects(routingObjectDefs)
             updateHealthCheckServices(serviceDb, healthMonitors)
+        }.onSuccess {
             originsConfig = content
             initialised.countDown()
         }.onFailure {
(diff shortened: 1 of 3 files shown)


components/proxy/src/main/kotlin/com/hotels/styx/services/YamlFileConfigurationService.kt
```diff
@@ -78,6 +82,10 @@ internal class YamlFileConfigurationService(
                 LOGGER.info("service stopped")
             }
 
+    override fun adminInterfaceHandlers() =
+        mutableMapOf(Pair("origins", TextHttpHandler { originsConfig }))
```
mikkokar: I believe you could just:

`        mapOf("origins" to TextHttpHandler { originsConfig })`

components/proxy/src/main/kotlin/com/hotels/styx/services/YamlFileConfigurationService.kt
```diff
@@ -91,6 +99,7 @@ internal class YamlFileConfigurationService(
         }.mapCatching { (healthMonitors, routingObjectDefs) ->
             updateRoutingObjects(routingObjectDefs)
             updateHealthCheckServices(serviceDb, healthMonitors)
+            originsConfig = content
```
mikkokar: It would be better to set the variables in `onSuccess` block like so:

```
        }.onSuccess {
            originsConfig = content
            initialised.countDown()
        }.onFailure {
            LOGGER.error("Failed to reload new configuration. cause='{}'", it.message, it)
        }
```

components/proxy/src/main/kotlin/com/hotels/styx/services/YamlFileConfigurationService.kt
chrisgresty: Thanks. That's a nice construct.
I thought I was having problems with mutable vs immutable earlier, but perhaps it was actually something else.

components/proxy/src/main/kotlin/com/hotels/styx/services/YamlFileConfigurationService.kt
chrisgresty: Ah, makes sense. Thanks.

PR #522, opened 2019-11-08
Add a "lbGroup" prefix to load balancing group membership tag.

mikkokar:
Add `lbGroup` prefix to load balancing group membership tags. This makes them self-identifying. As a result the tag can be recognised without having to know the load balancing group configuration.

Also add a `ObjectTags.kt` to serve as a formal specification for tag names.

### DIFF
--- a/components/proxy/src/main/kotlin/com/hotels/styx/ObjectTags.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/ObjectTags.kt
@@ -1,0 +1,18 @@
+/*
+  Copyright (C) 2013-2019 Expedia Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package com.hotels.styx
+
+fun lbGroupTag(name: String) = "lbGroup=$name"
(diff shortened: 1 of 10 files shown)


system-tests/ft-suite/src/test/kotlin/com/hotels/styx/routing/LoadBalancingGroupSpec.kt
```diff
@@ -149,30 +149,30 @@ class LoadBalancingGroupSpec : FeatureSpec() {
...
                                       - state:inactive
                                     config:
                                       host: localhost:${appA01.port()}
 
                                   app-A-02:
                                     type: HostProxy
                                     tags:
-                                      - App-A
+                                      - lbGroup=App-A
                                       - state:inactive:3
```
OwenLindsell: shouldn't this use `:` instead of `=` to be consistent with the state:inactive tag?

components/proxy/src/main/kotlin/com/hotels/styx/ObjectTags.kt
```diff
@@ -0,0 +1,18 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package com.hotels.styx
+
+fun lbGroupTag(name: String) = "lbGroup=$name"
```
chrisgresty: I guess you don't need it now (YAGNI, etc.) but do you also see adding functions here to extract the RHS from a tag ? e.g. `fun lbGroupTag(tags: Set<String>) : String`

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/routing/LoadBalancingGroupSpec.kt
chrisgresty: Or are we going to standardize on `=`?

mikkokar: Hi. I thought about this and chose ’=‘ because ‘:’ already appears in creation time tag values.

Therefore would be best to standardise on ‘=‘. My gut feeling is that it will be less likely to occur in the value parts.

components/proxy/src/main/kotlin/com/hotels/styx/ObjectTags.kt
mikkokar: Hi. Yes. Right now it is unnecessary. But will add if or when necessary.

Issue #529, opened 2019-11-14
Resolve routing object name clashes


mikkokar:
We recently added support for multiple backend service providers for the routing object engine (https://github.com/HotelsDotCom/styx/issues/519).

However, the resulting routing objects should be prefixed with the provider name to avoid possible routing object name clashes. Consider the following two origin files:

origins.yml:
```
- id: "origin"
  path: "/"
  healthCheck:
    uri: /hello.txt
  origins:
    - { id: "origin-01", host: "origin:8080" }
```

origins-green.yml:
```
- id: "origin"
  path: "/"
  healthCheck:
    uri: /hello.txt
  origins:
    - { id: "origin-01", host: "otherhost:8080" }
```

They would be translated to `LoadBalancingGroup` and `HostProxy` objects with clashing names. To rectify this the object names shall be prefixed with the provider name.

mikkokar:
This issue is addressed by #531. Instead of prefixing object names, we chose to reject all configurations introducing name clashes.

PR #531, opened 2019-11-15
YamlConfigurationService: Reject origins file if it introduces name clashes.

mikkokar:
Reject origins file reload if it would result in name clashes between routing objects, and
log a meaningful error message.


mikkokar:
Addresses #529. But in a more restrictive manner.

Issue #525, opened 2019-11-13
Improve StyxObjectStore notification mechanism


chrisgresty:
The current mechanism is vulnerable to out-of-order notifications, whereby older database states can be emitted after newer ones. Given that it is not essential for all notifications to be emitted, only that the notifications that are emitted are in the correct order, it is possible to replace the queue implementation with one that:
* associates version indices with each new database state
* collapses consecutive notifications
* maintains a limited version history

## Associated test failure
_From Bamboo build [#1140](https://bamboo.hcom/browse/STYX-STYX-1140):_ `StyxObjectStoreTest` - `Compute/Replace an existing value` can result in the watcher being added early to the DB, and so receiving an unexpected early update which fails the test. This test should be re-assessed.

mikkokar:
## Issues
### Out-of-order Notifications

The out-of-order issue affects the initial watch notification in particular. The state change notifications are queued with the modified snapshot. But the initial watch notification is queued as a lambda function that uses the most recent version at the time of its evaluation.

As a result, the initial watch can be followed by other notifications for earlier database snapshots.

### Watchers added too early

Watcher is added to the `watchers` list too early. As a result it might start receiving database change notifications before its initial watch notification.


## Acceptance Criteria

### A snapshot identifier
* Associate a unique identifier for each database snapshot.
* Is an integer (or long).
* Is strictly increasing, starting at zero, and increases by one for each new snapshot.
* A higher identifier implies more recent snapshot.

### Notifications

* Change notifications are collapsed to the most recent snapshot.
* A following "queue invariant" holds true: the notification queue contains at most 1 notification change event.

Suppose there are a burst of modifications to the database, resulting queued notifications for: `[snapshot(n), .. , snapshot(7), snapshot(6)]`, while the database is emitting notifications for `snapshot(5)`. Once the `snapshot(5)` has been processed, the database can skip all intermediate snapshots `6, 7, ...` and emit the most recent snapshot `n`. Therefore there is no need to store more than one watch notification.

* The change notifications and initial watch notifications must preserve their relative order. 

An initial watch notification is sent to a new watch subscriber *only*. Therefore each initial watch notification should be queued as a separate event. The relative ordering among initial watch notifications, and among the database change event must be preserved.

* Initial watch notification should always emit the most recently published snapshot. This is to prevent the initial watch notification from being followed by a database change notification for an earlier snapshot.

mikkokar:
Fixed by: #537

PR #537, opened 2019-11-21
Fix issue: #525: A race condition in styx object store.

mikkokar:
Fixes issue #525.

### DIFF
--- a/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
@@ -1,3 +1,18 @@
+/*
+  Copyright (C) 2013-2019 Expedia Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
 package com.hotels.styx.routing.db
 
 import com.hotels.styx.api.configuration.ObjectStore
--- a/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
@@ -1,0 +1,107 @@
+package com.hotels.styx.routing.db
+
+import com.hotels.styx.api.configuration.ObjectStore
+import org.pcollections.HashTreePMap
+import org.pcollections.PMap
+import reactor.core.publisher.FluxSink
+import java.util.Optional
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CopyOnWriteArrayList
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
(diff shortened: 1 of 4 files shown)


components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
```diff
@@ -0,0 +1,122 @@
...
+            pendingSnapshot = snapshot
+
+            pendingChangeNotification.getAndSet(true)
+        }
+
+        if (!inQueue) {
+            executor.submit {
+                pendingChangeNotification.set(false)
+
+                issuedSnapshot = pendingSnapshot
```
chrisgresty: Should these be protected by the lock? To prevent the possibility of a double notification of the same snapshot. Not a critical issue, admittedly.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
```diff
@@ -0,0 +1,122 @@
...
+                watchers.forEach { watcher ->
+                    watcher.invoke(newSnapshot(pendingSnapshot))
```
chrisgresty: Should probably be notifying with the issuedSnapshot rather than the pendingSnapshot.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
```diff
@@ -0,0 +1,122 @@
...
+
+internal class NotificationQueue<T>(val watchers: CopyOnWriteArrayList<ChangeWatcher<T>>, val executor: ExecutorService) {
+    @Volatile
+    private var pendingSnapshot = IndexedSnapshot<T>(0, HashTreePMap.empty())
+    @Volatile
+    private var issuedSnapshot = IndexedSnapshot<T>(0, HashTreePMap.empty())
+    private val pendingChangeNotification = AtomicBoolean(false)
+    private val lock = ReentrantLock()
+
+    private val listeners = ConcurrentHashMap<String, DispatchListener<T>>()
```
chrisgresty: Are the `listeners` only for use in testing, or is there another purpose?

dvlato: This code does not seem clear enough to me.  For instance,  can't we use more meaningful names than watchers and listeners? 

We might need to review the segmentation in functions to see if the sequence of function names help making the intent of the code clearer...

mikkokar: Yes. They are only for testing purposes.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
mikkokar: The event dispatch always runs serially: They run in the single threaded executor. 

However locking is necessary to guarantee the following invariant holds:

* `PendingChangeNotification == true` only when `issuedSnapshot < pendingSnapshot`
* `PendingChangeNotification == false` when `issuedSnapshot >= pendingSnapshot`

Clearly this invariant is not consistent in the event dispatch thread.

I will fix this.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
mikkokar: Oh, definitely!
 Thanks for spotting this.

components/proxy/src/main/kotlin/com/hotels/styx/routing/db/NotificationQueue.kt
mikkokar: Hi David. `Watcher` is the documented term in the object database API.

`Listener` is just an event hook for the debugging purposes. Perhaps I'll rename it to `eventHook`? But `listener` is more common in Java world. Or perhaps I'll clarify this in comments?

Issue #535, opened 2019-11-20
Add uptime admin endpoint


mikkokar:
Add admin endpoint to show server uptime when routing object engine is used.

Or add this information to some existing admin endpoint.

bestokes:
Fixed in PR #539.

PR #539, opened 2019-11-25
Issue #535: Add admin endpoint for uptime

mikkokar:
Adds an `/admin/uptime` endpoint to show Styx uptime.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/UptimeHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/UptimeHandler.java
@@ -1,0 +1,50 @@
+/*
+  Copyright (C) 2013-2019 Expedia Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package com.hotels.styx.admin.handlers;
+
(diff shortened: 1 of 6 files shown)


components/proxy/src/main/java/com/hotels/styx/admin/handlers/UptimeHandler.java
```diff
@@ -0,0 +1,50 @@
...
+public class UptimeHandler implements WebServiceHandler {
+    private final MetricRegistry metricRegistry;
+
+    public UptimeHandler(MetricRegistry metricRegistry) {
+        this.metricRegistry = metricRegistry;
+    }
+
+    @Override
+    public Eventual<HttpResponse> handle(HttpRequest request, HttpInterceptor.Context context) {
+        Object uptime = metricRegistry.getGauges().get("jvm.uptime.formatted").getValue();
```
dvlato: Don't we need to check for null values here? I don't know if there are possible race conditions.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UptimeHandler.java
```diff
@@ -0,0 +1,50 @@
...
+
+        return Eventual.of(HttpResponse.response(OK)
+                .disableCaching()
+                .addHeader(CONTENT_TYPE, "application/json")
+                .body(format("\"%s\"", uptime), UTF_8)
```
dvlato: Is this json content?

mikkokar: Hi. @dvlato, @chrisgresty said it is ;-)

chrisgresty: :)
JSON used to insist that only a map or an array was a valid document, but a few years ago this was changed to allow other value types (string, integer, etc). You could consider using `objectMapper.writeValueAsString(uptime)` instead.

dvlato: I guess technically we can say returning a quoted string is returning valid JSON... :)

PR #542, opened 2019-11-27
Build Styx Docker image using fabric8 docker maven plugin 

mikkokar:
This PR adds `fabric8` docker maven plugin to build Styx Core docker image.

This is to simplify and "abstract away" the details how to build Styx Docker image automatically from other projects.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,4 +133,4 @@ changelog:
 # Default configuration file: /styx/default-config/default.yml
 #
 docker-image: clean
-	mvn clean verify -Prelease,linux,docker -Dmaven.test.skip=true
+	mvn clean install -Prelease,linux,docker -Dmaven.test.skip=true
--- a/Makefile
+++ b/Makefile
@@ -133,4 +133,4 @@ changelog:
 # Default configuration file: /styx/default-config/default.yml
 #
 docker-image: clean
-	mvn clean install -Prelease,linux,docker -Dmaven.test.skip=true
+	mvn install -Prelease,linux,docker -Dmaven.test.skip=true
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -133,8 +133,4 @@ changelog:
 # Default configuration file: /styx/default-config/default.yml
 #
 docker-image: clean
-	mvn install -Prelease,linux -Dmaven.test.skip=true
-	mkdir -p ${DOCKER_CONTEXT}
-	cp `find  distribution/target -maxdepth 1 -name "styx*linux-x86_64.zip"` ${DOCKER_CONTEXT}/styx.zip
-	cp docker/styx-image/* ${DOCKER_CONTEXT}
-	docker build -t styxcore:latest --build-arg STYX_IMAGE=styx.zip ${DOCKER_CONTEXT}/.
+	mvn clean install -Prelease,linux,docker -Dmaven.test.skip=true
```
dvlato: do we need mvn clean even though there's a 'clean' above?

distribution/pom.xml
```diff
@@ -133,4 +137,58 @@
...
+              </execution>
+            </executions>
+
+            <configuration>
+              <images>
+                <image>
+                  <alias>styxcore</alias>
+                  <name>${styxcore.docker.image}</name>
+                  <build>
+                    <dockerFile>${project.basedir}/../docker/styx-image/Dockerfile</dockerFile>
```
dvlato: we probably can reference the parent project here instead of using  ${project.basedir}/../

Makefile
mikkokar: No. Thanks for spotting 👍

distribution/pom.xml
mikkokar: I have changed this to `${project.parent.basedir}`.

PR #549, opened 2019-12-02
Common tags

mikkokar:
## Description

Introduces a `CommonTag` class to simplify and harmonise Styx object tags API. Now each Tag will be represented as a CommonTag instance, with provided `name`, `encode` and `decode` functions.

This imposes a common format: `name = tag-value`. 

Each tag is associated with value type `T`. The `encode` and `decode` functions convert between values of `T` and encoded tag value strings.

All common tag operations, like creating, searching, matching, and removing tags are now implemented as higher order functions in the common base class.

## Alternative API

Some of the common functions could be implemented as Kotlin extension methods. Which method do you prefer? Let me know and we can shape the code accordingly.

An alternative to CommonTags find method would be:
```
fun <T> Set<String>.valueOf(tag: CommonValueTag<T>): T? = this.firstOrNull { tag.match(it) }
        ?.let { tag.valuePart(it) }
        ?.let { tag.decode(it) }
```

Similarly, an alternative to `match` would be:
```
fun String.isA(tag: CommonValueTag<*>) = tag.match(this)
```

An alternative to `valueOf` method would be:
```
fun <T> String.match(tag: CommonValueTag<T>) = tag.valuePart(this)
         ?.let { tag.decode(it) }
```

### DIFF
--- a/components/proxy/src/main/kotlin/com/hotels/styx/CommonTags.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/CommonTags.kt
@@ -15,36 +15,85 @@
  */
 package com.hotels.styx
 
-/*
- * Common Tag Operations:
+/**
+ * A common tag representation.
+ *
+ * Enforces a common tag format and provides a set of higher order functionality
+ * that operate on provided `name`, `encode` and `decode` properties.
+ *
+ * A tag string is a name-value pair separated by equal ('=') sign:
+ *
+ *     tag-string = name "=" value-part
+ *
+ * `encode` and `decode` are functions to encode a T to a value-part string,
+ * and to decode the value-part string back to T.
+ *
+ * @property name a tag name
+ * @property encode a function that encodes a value of T as a tag value string
+ * @property decode a function that decodes a tag value string as a T
  */
-
 sealed class CommonValueTag<T>(
         val name: String,
         val encode: (T) -> String?,
         val decode: (String) -> T?) {
 
+    /**
+     * Extracts value part (the right hand side) from a tag string.
+     *
+     * @param tag a tag string
+     * @return the tag value
+     */
     fun valuePart(tag: String) = if (tag.startsWith("$name=")) {
         tag.removePrefix("$name=")
     } else {
         null
     }
 
+    /**
+     * Tests if a given string matches this tag. A match is positive when the string
+     * starts with `name` followed by `=`.
+     *
+     * @param a tag string
+     * @return True if this string is possibly a matching tag. Otherwise return false.
+     */
     fun match(tag: String) = tag.startsWith("$name=")
 
+    /**
+     * Decodes given tag string to its typed value.
+     *
+     * @param tag a tag string
+     * @return a decoded tag value, or null if decoding failed
+     */
     fun valueOf(tag: String) = valuePart(tag)
             ?.let { decode(it) }
 
+    /**
+     * Find this tag from a set of tag strings. If found, decode the tag value. Return null if
+     * tag was not found, or if decoding failed.
+     *
+     * @param tags a set of tag strings
+     * @return A decoded value if tag was found. Otherwise return null.
+     */
     fun find(tags: Set<String>) = tags.firstOrNull { this.match(it) }
             ?.let { valuePart(it) }
             ?.let { this.decode(it) }
 
+    /**
+     * Removes all instances of this tag from a set of tag strings.
+     * Return a new Set<String> with all instances of this tag removed.
+     *
+     * @param tags a set of tag strings
+     * @return A new set of strings without this tag.
+     */
     fun remove(tags: Set<String>) = tags
             .filterNot { this.match(it) }
             .toSet()
 }
 
-
+/**
+ * A NullableValueTag invoke method returns null when value cannot be encoded to string.
+ * The API consumer must handle this situation.
+ */
 class NullableValueTag<T>(
         name: String,
         encode: (T) -> String?,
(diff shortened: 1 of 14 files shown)


components/proxy/src/main/kotlin/com/hotels/styx/CommonTags.kt
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Find this tag from a set of tag strings. If found, decode the tag value. Return null if
+     * tag was not found, or if decoding failed.
+     *
+     * @param tags a set of tag strings
+     * @return A decoded value if tag was found. Otherwise return null.
+     */
+    fun find(tags: Set<String>) = tags.firstOrNull { this.match(it) }
+            ?.let { valuePart(it) }
+            ?.let { this.decode(it) }
```
chrisgresty: Perhaps use `valueOf(it)` here

components/proxy/src/main/kotlin/com/hotels/styx/CommonTags.kt
```diff
@@ -0,0 +1,122 @@
...
+ * and to decode the value-part string back to T.
+ *
+ * @property name a tag name
+ * @property encode a function that encodes a value of T as a tag value string
+ * @property decode a function that decodes a tag value string as a T
+ */
+sealed class CommonValueTag<T>(
+        val name: String,
+        val encode: (T) -> String?,
+        val decode: (String) -> T?) {
```
chrisgresty: Is it worth providing defaults for encode and decode as `{ it }` ?

components/proxy/src/main/kotlin/com/hotels/styx/ObjectTags.kt
```diff
@@ -15,58 +15,64 @@
...
+        { it },
+        { it })
+
+/*
+ * TAG: healthCheck
+ * healthCheck=on
+ * healthCheck=on;probes-OK:2
+ * healthCheck=on;probes-FAIL:1
+ */
 private const val HEALTHCHECK = "healthCheck"
```
chrisgresty: No longer used.

components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
```diff
@@ -85,12 +85,32 @@ internal class HealthCheckMonitoringService(
...
                 .filter(::containsRelevantStateTag)
-                .forEach { (name, _) ->
-                    markObject(objectStore, name, ObjectActive(0, false))
+                .forEach { (name, record) ->
+                    objectStore.get(name).ifPresent {
+                        try {
+                            objectStore.compute(name) { previous ->
+                                if (previous == null) throw ObjectDisappearedException()
+
+                                val newTags = record.tags
```
chrisgresty: Should this be `previous.tags` instead of `record.tags`?

components/proxy/src/main/kotlin/com/hotels/styx/services/OriginsConfigConverter.kt
```diff
@@ -151,7 +151,7 @@ internal class OriginsConfigConverter(
         }
 
         internal fun hostProxy(app: BackendService, origin: Origin) : StyxObjectDefinition {
-            val healthCheckTag :String = if (isHealthCheckConfigured(app)) stateTag(STATE_UNREACHABLE) else stateTag(STATE_ACTIVE);
+            val healthCheckTag :String = if (isHealthCheckConfigured(app)) stateTag(STATE_UNREACHABLE)!! else stateTag(STATE_ACTIVE)!!
```
chrisgresty: No need for `!!`

components/proxy/src/test/kotlin/com/hotels/styx/ObjectTagsTest.kt
```diff
@@ -45,6 +45,7 @@ class ObjectTagsKtTest : BehaviorSpec({
         }
         `when`("the label is blank") {
             then("null is returned") {
+                // Will now throw:
```
chrisgresty: Is this comment still current?

components/proxy/src/main/kotlin/com/hotels/styx/CommonTags.kt
mikkokar: Is this possible without knowing the type T?

components/proxy/src/main/kotlin/com/hotels/styx/CommonTags.kt
mikkokar: ok

components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
mikkokar: Either way. `previous` and `record` are the same object. But using `previous` feels more correct sine it may have been updated concurrently. Will fix this.

components/proxy/src/main/kotlin/com/hotels/styx/services/OriginsConfigConverter.kt
mikkokar: 👍

components/proxy/src/test/kotlin/com/hotels/styx/ObjectTagsTest.kt
mikkokar: No. I will remove this.

PR #551, opened 2019-12-02
Add close() method to connectionPool so connections are closed when origins are reloaded

dvlato:
We were missing a ConnectionPool.close() method that will close open connections when an origins file is reloaded.

We are just invoking Connection::close() without modifying the metrics as, in most cases (except when the origin is removed), they would be overridden by the new SimpleConnectionPool for that origin server. We can discuss this to see what's the best way to implement it.

We can also discuss what's the best meaningful test we can implement. I guess we could add a global counter for the tota number of connections opened / closed (not per origin but global!). We might also have to locate all the code paths that should trigger a pool closure  and add assertions for those in unit tests. 

Of course, this PR does not prevent us from leaving connections pool that hadn't been returned to the pool (they are not in "availableConnections") at the moment the pool was closed.

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
+++ b/components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
@@ -75,25 +75,29 @@ public class SimpleConnectionPool implements ConnectionPool, Connection.Listener
 
     @Override
     public Publisher<Connection> borrowConnection() {
-        return Mono.<Connection>create(sink -> {
-            Connection connection = dequeue();
-            if (connection != null) {
-                attemptBorrowConnection(sink, connection);
-            } else {
-                if (waitingSubscribers.size() < poolSettings.maxPendingConnectionsPerHost()) {
-                    this.waitingSubscribers.add(sink);
-                    sink.onDispose(() -> waitingSubscribers.remove(sink));
-                    newConnection();
+        if (active) {
+            return Mono.<Connection>create(sink -> {
+                Connection connection = dequeue();
+                if (connection != null) {
+                    attemptBorrowConnection(sink, connection);
                 } else {
-                    sink.error(new MaxPendingConnectionsExceededException(
-                            origin,
-                            poolSettings.maxPendingConnectionsPerHost(),
-                            poolSettings.maxPendingConnectionsPerHost()));
+                    if (waitingSubscribers.size() < poolSettings.maxPendingConnectionsPerHost()) {
+                        this.waitingSubscribers.add(sink);
+                        sink.onDispose(() -> waitingSubscribers.remove(sink));
+                        newConnection();
+                    } else {
+                        sink.error(new MaxPendingConnectionsExceededException(
+                                origin,
+                                poolSettings.maxPendingConnectionsPerHost(),
+                                poolSettings.maxPendingConnectionsPerHost()));
+                    }
                 }
-            }
-        }).timeout(
-                Duration.ofMillis(poolSettings.pendingConnectionTimeoutMillis()),
-                Mono.error(() -> new MaxPendingConnectionTimeoutException(origin, connectionSettings.connectTimeoutMillis())));
+            }).timeout(
+                    Duration.ofMillis(poolSettings.pendingConnectionTimeoutMillis()),
+                    Mono.error(() -> new MaxPendingConnectionTimeoutException(origin, connectionSettings.connectTimeoutMillis())));
+        } else {
+            throw new IllegalStateException("Pool is closed");
+        }
     }
 
     private void newConnection() {
@@ -158,23 +162,25 @@ public class SimpleConnectionPool implements ConnectionPool, Connection.Listener
     }
 
     public boolean returnConnection(Connection connection) {
-        if (!active) {
-            connection.close();
-        } else {
+        if (active) {
             borrowedCount.decrementAndGet();
             if (connection.isConnected()) {
                 queueNewConnection(connection);
             }
+        } else {
+            connection.close();
         }
         return false;
     }
 
     public boolean closeConnection(Connection connection) {
         connection.close();
-        borrowedCount.decrementAndGet();
-        closedConnections.incrementAndGet();
+        if (active) {
+            borrowedCount.decrementAndGet();
+            closedConnections.incrementAndGet();
 
-        newConnection();
+            newConnection();
+        }
         return true;
     }
 
@@ -193,8 +199,10 @@ public class SimpleConnectionPool implements ConnectionPool, Connection.Listener
 
     @Override
     public void connectionClosed(Connection connection) {
-        terminatedConnections.incrementAndGet();
-        availableConnections.remove(connection);
+        if (active) {
+            terminatedConnections.incrementAndGet();
+            availableConnections.remove(connection);
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


mikkokar:
Hi @dvlato this PR doesn't have any effect unless the pool consumer calls the `close()` method.

We need to check the following pool consumers:

 1. BackendServicesProxy & OriginsInventory
 2. HostProxy
 3. ProxyToBackend class

Going forward, the `HostProxy` is the only class that will survive our refactored Styx Core. The other two/three will disappear. Hopefully soon.

mikkokar:
> Of course, this PR does not prevent us from leaving connections pool that hadn't been returned to the pool (they are not in "availableConnections") at the moment the pool was closed.

I suggest: Add a `closed` flag in the connection pool. Set it `true` when the pool is closed. Then in the `returnConnection` check the flag and close the returned connection when the pool is closed:

```
    public boolean returnConnection(Connection connection) {
        borrowedCount.decrementAndGet();

        // This is the new bit:
        if (closed) {
           connection.close();
        } 

        if (connection.isConnected()) {
            queueNewConnection(connection);
        }
        return false;
    }
```

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -73,25 +75,29 @@ public Origin getOrigin() {
...
                 }
-            }
-        }).timeout(
-                Duration.ofMillis(poolSettings.pendingConnectionTimeoutMillis()),
-                Mono.error(() -> new MaxPendingConnectionTimeoutException(origin, connectionSettings.connectTimeoutMillis())));
+            }).timeout(
+                    Duration.ofMillis(poolSettings.pendingConnectionTimeoutMillis()),
+                    Mono.error(() -> new MaxPendingConnectionTimeoutException(origin, connectionSettings.connectTimeoutMillis())));
+        } else {
+            throw new IllegalStateException("Pool is closed");
```
mikkokar: Hi @dvlato . Might be better to do a sink.error() instead.

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -156,19 +162,25 @@ private void attemptBorrowConnection(MonoSink<Connection> sink, Connection conne
...
         connection.close();
-        borrowedCount.decrementAndGet();
-        closedConnections.incrementAndGet();
+        if (active) {
+            borrowedCount.decrementAndGet();
+            closedConnections.incrementAndGet();
 
-        newConnection();
+            newConnection();
+        }
```
mikkokar: I think we should decrement the metrics regardless, and trigger a `newConnection()` only if the pool is `active`. Like so:

```
         connection.close();
         borrowedCount.decrementAndGet();
         closedConnections.incrementAndGet();
         if (active) {
             newConnection();
         }
```

The metrics might still be made available even for the closed pool (perhaps as a separate enhancement. Therefore we should keep them consistent with the reality.

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -191,6 +206,18 @@ public void connectionClosed(Connection connection) {
         availableConnections.remove(connection);
     }
 
+    @Override
+    public void close() {
+        active = false;
+        Connection con;
+        while ((con = availableConnections.poll()) != null) {
+            borrowedCount.decrementAndGet();
```
mikkokar: Decrementing `borrowedCount` seems wrong. The connections in `availableConnections` are sitting idle in the pool, and not being borrowed.

components/client/src/test/unit/java/com/hotels/styx/client/connectionpool/SimpleConnectionPoolTest.java
```diff
@@ -921,4 +922,98 @@ public void ensureInEstablishmentCountIsDecrementedInError() throws InterruptedE
...
+
+        when(connectionFactory.createConnection(any(Origin.class), any(ConnectionSettings.class)))
+                .thenReturn(Mono.just(connection1));
+
+        ConnectionPoolSettings poolSettings = new ConnectionPoolSettings.Builder()
+                .maxConnectionsPerHost(2)
+                .maxPendingConnectionsPerHost(2)
+                .build();
+
+        SimpleConnectionPool pool = new SimpleConnectionPool(origin, poolSettings, connectionFactory);
```
mikkokar: Consider inlining the `poolSettings` variable. It is only used once.

components/client/src/test/unit/java/com/hotels/styx/client/connectionpool/SimpleConnectionPoolTest.java
```diff
@@ -921,4 +922,98 @@ public void ensureInEstablishmentCountIsDecrementedInError() throws InterruptedE
...
+            return true;
+        }).when(connection1).close();
+        pool.returnConnection(connection1);
+        assertEquals(pool.stats().connectionAttempts(), 1);
+        assertEquals(pool.stats().busyConnectionCount(), 0);
+        assertEquals(1, pool.stats(). closedConnections());
+    }
+
+    @Test
+    public void closesConnectionAndUpdatesMetricsWhenClosingConnectionFromClosedPool() {
```
mikkokar: This test is to show that a non-active pool doesn't substitute a closed TCP connection. Therefore I'd consider either renaming this test to something like `closedPoolDoesNotReplaceClosedTcpConnection`.

Or perhaps even better clarify this in a comment?

components/client/src/test/unit/java/com/hotels/styx/client/connectionpool/SimpleConnectionPoolTest.java
```diff
@@ -921,4 +922,98 @@ public void ensureInEstablishmentCountIsDecrementedInError() throws InterruptedE
...
+        assertEquals(pool.stats().connectionAttempts(), 1);
+        assertEquals(pool.stats().pendingConnectionCount(), 0);
+        assertEquals(pool.stats().busyConnectionCount(), 0);
+        assertEquals(pool.stats().availableConnectionCount(), 0);
+
+        assertEquals(pool.stats().terminatedConnections(), 1);
+        assertEquals(pool.stats().closedConnections(), 0);
+    }
+
+
```
mikkokar: Good test coverage. However one scenario is omitted: that of a `close()` method closes all active, non-borrowed TCP connections.

components/client/src/test/unit/java/com/hotels/styx/client/connectionpool/SimpleConnectionPoolTest.java
dvlato: I think I just copied one of your tests (closesConnection I think) as to keep the code base uniform.  Let me update all the tests that do not require special configuration...

components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
dvlato: oh you are right @mikkokar ,  I added that by mistake...  Fixed.

components/client/src/test/unit/java/com/hotels/styx/client/connectionpool/SimpleConnectionPoolTest.java
dvlato: Let's try with a clearer name.

components/client/src/test/unit/java/com/hotels/styx/client/connectionpool/SimpleConnectionPoolTest.java
dvlato: Added.

mikkokar:
Travis fails due to check style error: 

`[[1;31mERROR[m] src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java:[99,29] (whitespace) ParenPad: '(' is followed by whitespace.`

PR #554, opened 2019-12-05
Refactor http servers

mikkokar:
## Rationale

Simplify Styx HTTP server implementation:
- Allow only one "connector" for each HTTP server (port).
- Remove unnecessary builders.
- Move plugin loading and initialisation into a separate class. 
- Move data plane handler creation and move it into Styx Server.
- Create separate Netty HTTP servers for HTTP and HTTPS endpoints.


components/api/src/main/java/com/hotels/styx/api/extension/service/spi/AbstractStyxService.java
```diff
@@ -67,12 +72,17 @@ public StyxServiceStatus status() {
...
         if (changed) {
             return startService()
                     .exceptionally(failWithMessage("Service failed to start."))
-                    .thenAccept(na -> status.compareAndSet(STARTING, RUNNING));
+                    .thenAccept(na -> {
+                        status.compareAndSet(STARTING, RUNNING);
+                        LOGGER.info("Started {}", name);
+                    });
         } else {
+            LOGGER.info("Starting {} failed", name);
```
dvlato: do we need a log line next to an exception?

components/api/src/main/java/com/hotels/styx/api/extension/service/spi/AbstractStyxService.java
```diff
@@ -81,11 +91,16 @@ public StyxServiceStatus status() {
...
         if (changed) {
             return stopService()
                     .exceptionally(failWithMessage("Service failed to stop."))
-                    .thenAccept(na -> status.compareAndSet(STOPPING, STOPPED));
+                    .thenAccept(na -> {
+                        status.compareAndSet(STOPPING, STOPPED);
+                        LOGGER.info("Stopped {}", name);
+                    });
         } else {
+            LOGGER.info("Stoppig {} failed", name);
```
dvlato: Same here. Maybe we can unify the stopping failed message with the exception.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -163,45 +167,86 @@ public StyxServer(StyxServerComponents components, Stopwatch stopwatch) {
...
+        HttpHandler httpHandler = new StyxPipelineFactory(
+                components.routingObjectFactoryContext(),
+                components.environment(),
+                components.services(),
+                components.plugins(),
+                components.eventLoopGroup(),
+                components.nettySocketChannelClass())
+                .create();
+
+        // Startup phase 1: start plugins, control plane providers, and other services:
```
dvlato: Can we break up this method?

components/proxy/src/main/java/com/hotels/styx/proxy/ProxyServerBuilder.java
```diff
@@ -19,57 +19,83 @@
...
+        return builder;
     }
 
     private LiveHttpResponse.Transformer addInfoHeader(LiveHttpResponse.Transformer responseBuilder, LiveHttpRequest request) {
         return responseBuilder.header(styxInfoHeaderName, responseInfoFormat.format(request));
     }
 
-    public ProxyServerBuilder handlerFactory(Supplier<HttpHandler> handlerFactory) {
-        this.handlerFactory = handlerFactory;
+    public ProxyServerBuilder handler(HttpHandler handlerFactory) {
```
dvlato: can we rename handlerFactory if it is not a factory anymore?

components/server/src/main/java/com/hotels/styx/server/netty/NettyServer.java
```diff
@@ -78,53 +73,34 @@
...
-            return new InetSocketAddress(host, httpsServerSocketBinder.port());
-        } else {
-            return null;
-        }
+        return Optional.ofNullable(protocolSocketBinder)
+                .map(it -> {
+                    try {
+                        return new InetSocketAddress(host, it.port());
+                    } catch (IllegalArgumentException e) {
+                        return null;
```
dvlato: Can we at least log anything here?

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -220,8 +265,14 @@ private static StartupConfig parseStartupConfig(String[] args) {
...
-        this.serviceManager.addListener(new ServerStartListener(this));
-        this.serviceManager.startAsync();
+        CompletableFuture.runAsync(() -> {
+            // doStart should return quicly. Therefore offload waiting on a separate thread:
+            this.phase1Services.addListener(new Phase1ServerStartListener(this));
+            this.phase1Services.startAsync().awaitHealthy();
+
+            this.phase2Services.addListener(new ServerStartListener(this));
+            this.phase2Services.startAsync();
+        });
```
chrisgresty: If anything in here fails, is the exception being logged?

components/api/src/main/java/com/hotels/styx/api/extension/service/spi/AbstractStyxService.java
mikkokar: Well not necessarily. But let me add the service name to the exception message.

components/server/src/main/java/com/hotels/styx/server/netty/NettyServer.java
mikkokar: No.

This is not necessarily an erroneous situation. `NettyServer` is a stateful object. Returning `null` implies the server is shut (or not started). Whether or not this is acceptable depends on caller's use case. Logging a message therefore needs to take place at call site.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
mikkokar: Not really, because this is in a constructor, and assigning to final variables. 
But I will be returning to this soon. Expect more changes, but in separate increments.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
mikkokar: I have made some changes to log them from StyxServer logger.

Sadly the `shutdownLogging` mechanism (see in StyxServer.java) prevents the tests from working correctly. I need to look into removing this in near future.

PR #560, opened 2019-12-10
Remove RxJava - Part 4

OwenLindsell:
Request body now follows reactive streams back pressure protocol.


components/server/src/main/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoder.java
```diff
@@ -80,10 +76,13 @@ protected void decode(ChannelHandlerContext ctx, HttpObject httpObject, List<Obj
...
             if (httpObject instanceof HttpRequest) {
-                this.producer = new FlowControllingHttpContentProducer(ctx, this.flowControlEnabled);
-                Observable<ByteBuf> contentObservable = Observable.create(contentSubscriber -> {
-                    contentSubscriber.setProducer(this.producer);
-                    this.producer.subscriptionStart(contentSubscriber);
+                this.producer = new FlowControllingHttpContentProducer(ctx);
+                Flux<ByteBuf> contentObservable = Flux.create(sink -> {
+                    ctx.channel().config().setAutoRead(false);
+                    sink.onRequest(n -> {
+                        ctx.channel().read();
```
dvlato: We can use a lambda expression instead of a statement here

components/server/src/main/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoder.java
```diff
@@ -184,22 +183,17 @@ private HttpVersion toStyxVersion(io.netty.handler.codec.http.HttpVersion httpVe
...
         private boolean completed;
-        private Subscriber<? super ByteBuf> contentSubscriber;
+        private FluxSink<? super ByteBuf> contentSubscriber;
 
-        FlowControllingHttpContentProducer(ChannelHandlerContext ctx, boolean flowControlEnabled) {
+        FlowControllingHttpContentProducer(ChannelHandlerContext ctx) {
             this.ctx = ctx;
-
-            if (flowControlEnabled) {
-                this.ctx.channel().config().setAutoRead(false);
```
dvlato: Why don't we keep the setAutoRead here?  (without the flag )

components/server/src/main/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoder.java
```diff
@@ -80,10 +76,11 @@ protected void decode(ChannelHandlerContext ctx, HttpObject httpObject, List<Obj
 
         try {
             if (httpObject instanceof HttpRequest) {
-                this.producer = new FlowControllingHttpContentProducer(ctx, this.flowControlEnabled);
-                Observable<ByteBuf> contentObservable = Observable.create(contentSubscriber -> {
-                    contentSubscriber.setProducer(this.producer);
-                    this.producer.subscriptionStart(contentSubscriber);
+                this.producer = new FlowControllingHttpContentProducer(ctx);
+                Flux<ByteBuf> contentObservable = Flux.create(sink -> {
+                    ctx.channel().config().setAutoRead(false);
```
mikkokar: Hi @OwenLindsell We need to set `autoRead(false)` straight away after receiving the headers. That is, move it out of the subscription. Otherwise the flow control only kicks later when the Styx "client side" subscribes to the request content. This might be a lot later if should there be any asynchronous interceptors in the forwarding path.

OwenLindsell: good spot.  I've moved it 👍

components/common/src/main/java/com/hotels/styx/common/content/BufferSubscriber.java
```diff
@@ -0,0 +1,54 @@
...
+import com.hotels.styx.api.Buffer;
+import com.hotels.styx.api.Buffers;
+import io.netty.buffer.ByteBuf;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+/**
+ * A Subscriber that converts from ByteBuf to Buffer.
+ */
+public class BufferSubscriber implements Subscriber<ByteBuf> {
```
mikkokar: AFAICS this could be a private static inner class in ContentPublisher.

components/common/src/main/java/com/hotels/styx/common/content/ContentPublisher.java
```diff
@@ -0,0 +1,53 @@
...
+import com.hotels.styx.api.Buffer;
+import io.netty.channel.EventLoop;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+/**
+ * A publisher to wrap the FlowControllingHttpContentProducer FSM and perform subscription operations via a Netty EventLoop.
+ */
+public final class ContentPublisher implements Publisher<Buffer> {
```
mikkokar: Perhaps we should rename this to `ByteStreamPublisher`?  Because it is the ByteStream publisher for which this class produces content. That might make it easier for other devs make the connection.

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
```diff
@@ -33,17 +32,20 @@
...
+import static com.hotels.styx.common.content.FlowControllingHttpContentProducer.ProducerState.EMITTING_BUFFERED_CONTENT;
+import static com.hotels.styx.common.content.FlowControllingHttpContentProducer.ProducerState.STREAMING;
+import static com.hotels.styx.common.content.FlowControllingHttpContentProducer.ProducerState.TERMINATED;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 import static reactor.core.publisher.Operators.addCap;
 
-class FlowControllingHttpContentProducer {
+/**
+ * A FSM that controls the flow of content chunks in accordance with the Reactive Streams specification.
```
mikkokar:
```
 * A FSM that controls the flow of content chunks in accordance with the Reactive Streams specification.
 * Specifically, it:
 *   1. Enforces the ordering of reactive events.
 *   2. Handles incoming content stream from the network 
 *   3. Handles reactive subscription, cancel, and back pressure
 *   4. Is not thread safe.
 *   5. Relies on external scheduler to serialise events.
```

components/server/src/main/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoder.java
```diff
@@ -61,80 +62,108 @@
...
+                () -> scheduleResourcesTearDown(ctx),
+                format("%s, %s", loggingPrefix, ""),
+                null);
+    }
+
+    private void scheduleResourcesTearDown(ChannelHandlerContext ctx) {
+        ctx.channel().eventLoop().schedule(
+                () -> producer.ifPresent(FlowControllingHttpContentProducer::tearDownResources),
+                TEARDOWN_DELAY,
+                MILLISECONDS);
```
mikkokar: Now looking back retrospectively, it seems bit funny that we tear down resources by generating a `ResponseTimeoutException` with "channelClosed" reason phrase. Let's think about improving this in the near future.

mikkokar: By the same token, I now realise that the `scheduleTeardown` trigger should be different for uplink traffic.

The FSM was developed for downlink traffic. In downlink direction we cannot have two back-to-back responses without subscriber fully consuming the response body. Therefore in downlink we can use channel idleness to drive the `scheduleTeardown` event.

  * The FSM must go to the COMPLETED state before the channel can be reused for the next request. Because origin cannot respond anything before we send the another request, the cannel (TCP connection) would remain idle if the content subscriber never consumes the content. 

But for uplink, I believe, this assumption is not valid. It is possible for an origin to respond without consuming the request body at all. This response would allows a remote a sender to send a new request even if Styx never consumed the original request body. Therefore we cannot rely on channel idleness to clean up resources.

We need to think of something else that is guaranteed to trigger a clean up.

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
OwenLindsell: Thanks, I'll just paste that in! :)

components/common/src/main/java/com/hotels/styx/common/content/BufferSubscriber.java
OwenLindsell: Good point - it is now!

components/common/src/main/java/com/hotels/styx/common/content/ContentPublisher.java
OwenLindsell: That couples it deeply to the ByteStream class, which it knows nothing about.  It only knows about the FlowControllingHttpContentProducer and even that doesn't know about ByteStream.  However, I get your point, the name is not very meaningful, so I have renamed it to FlowControllingPublisher.  Is that more helpful?

components/server/src/main/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoder.java
```diff
@@ -61,80 +62,108 @@
...
-                HttpRequest request = (HttpRequest) httpObject;
-                LiveHttpRequest styxRequest = toStyxRequest(request, contentObservable);
+            if (msg instanceof HttpRequest) {
+                ctx.channel().config().setAutoRead(false);
+                ctx.channel().read();
+
+                Publisher<Buffer> contentPublisher = new ContentPublisher(ctx.channel().eventLoop(), getNewContentProducer(ctx));
+
+                HttpRequest request = (HttpRequest) msg;
+                LiveHttpRequest styxRequest = toStyxRequest(request, contentPublisher);
```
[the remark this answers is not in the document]
OwenLindsell: definitely a matter of taste here.  Inlining is ok, but I try to avoid multiple nested inlining.  I just find it easier to read like this.

components/server/src/main/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoder.java
```diff
@@ -61,80 +62,108 @@
...
-                    this.producer.subscriptionStart(contentSubscriber);
-                });
-
-                HttpRequest request = (HttpRequest) httpObject;
-                LiveHttpRequest styxRequest = toStyxRequest(request, contentObservable);
+            if (msg instanceof HttpRequest) {
+                ctx.channel().config().setAutoRead(false);
+                ctx.channel().read();
+
+                Publisher<Buffer> contentPublisher = new ContentPublisher(ctx.channel().eventLoop(), getNewContentProducer(ctx));
```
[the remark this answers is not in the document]
OwenLindsell: seems about the same to me, so I've changed it to your preferred way

components/server/src/main/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoder.java
```diff
@@ -61,80 +62,108 @@
...
+                    getContentProducer(ctx).lastHttpContent();
                 }
+
             }
         } catch (BadRequestException ex) {
             throw ex;
         } catch (Exception ex) {
-            throw new BadRequestException(ex.getMessage() + " in " + httpObject, ex);
+            throw new BadRequestException(ex.getMessage() + " in " + msg, ex);
         }
```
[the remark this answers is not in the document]
OwenLindsell: I considered that.  I think the exception will be handled by Netty, which will pass it onto the exceptionCaught method, which calls the FSM and then propagates it to the HttpPipelineHandler.  Or have I not understood it correctly?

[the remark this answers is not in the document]
mikkokar: Hi yes. The exception will propagate. And ideally the pipeline handler responds by closing the connection. It is still better to be defensive, and move the FSM to TERMINATED. That guarantees that any extra pending content chunk events will be correctly dealt with. Also the FSM state would be consistent with the error, in case someone ever needs to debug it.

components/common/src/main/java/com/hotels/styx/common/content/ContentPublisher.java
mikkokar: Hi. It was just a suggestion. I'm not too opinionated about this. But between `FlowControllingPublisher` and `ContentPublisher`, I would prefer the original because it is shorter and simpler :-)

components/server/src/main/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoder.java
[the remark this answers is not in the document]
OwenLindsell: The exceptionCaught method makes the same call to the FSM, so it will move it to terminated.

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
```diff
@@ -70,6 +83,9 @@
 
     private volatile Subscriber<? super ByteBuf> contentSubscriber;
 
+    private final HashedWheelTimer timer = new HashedWheelTimer();
```
dvlato: From HashedWheelTimer: 

```
Do not create many instances.
HashedWheelTimer creates a new thread whenever it is instantiated and started. Therefore, you should make sure to create only one instance and share it across your application. One of the common mistakes, that makes your application unresponsive, is to create a new instance for every connection.

```
However, it seems we are creating a Timer for every http request we receive.

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
```diff
@@ -13,47 +13,60 @@
...
 import com.google.common.annotations.VisibleForTesting;
 import com.hotels.styx.api.exceptions.ResponseTimeoutException;
 import com.hotels.styx.api.extension.Origin;
-import com.hotels.styx.client.netty.ConsumerDisconnectedException;
 import com.hotels.styx.common.StateMachine;
 import io.netty.buffer.ByteBuf;
+import io.netty.util.HashedWheelTimer;
 import io.netty.util.ReferenceCountUtil;
+import io.netty.util.Timeout;
+import io.netty.util.Timer;
```
dvlato: Import seems unnecessary.

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
```diff
@@ -13,47 +13,60 @@
...
+import io.netty.util.TimerTask;
 import org.reactivestreams.Subscriber;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.TimeUnit;
```
dvlato: Import seems unnecessary

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
```diff
@@ -79,39 +95,44 @@
...
+            long inactivityTimeoutMs) {
         this.askForMore = requireNonNull(askForMore);
         this.onCompleteAction = requireNonNull(onCompleteAction);
         this.onTerminateAction = requireNonNull(onTerminateAction);
-        this.delayedTearDownAction = requireNonNull(delayedTearDownAction);
-        this.origin = requireNonNull(origin);
+        this.loggingPrefix = loggingPrefix;
+        this.origin = origin;
 
+        TimerTask timerTask = timeout -> {
```
dvlato: we don't need curly braces here

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyToStyxResponsePropagator.java
```diff
@@ -95,9 +90,8 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
     private RuntimeException toStyxException(Throwable cause) {
         if (cause instanceof OutOfMemoryError) {
             return new StyxClientException("Styx Client out of memory. " + cause.getMessage(), cause);
-        } else {
-            return new BadHttpResponseException(origin, cause);
         }
+        return new BadHttpResponseException(origin, cause);
```
dvlato: I actually prefer the if {} else {} construct as it's a bit clearer - but you can argue it's more verbose for something that's obvious...

components/client/src/test/unit/resources/logback.xml
```diff
@@ -16,6 +16,5 @@
   <logger name="com.hotels.styx.http-messages.inbound" level="INFO"/>
   <logger name="com.hotels.styx.http-messages.outbound" level="INFO"/>
   <logger name="com.hotels.styx.client.OriginRestrictionLoadBalancingStrategy" level="INFO"/>
-  <logger name="com.hotels.styx.client.netty.connectionpool.FlowControllingHttpContentProducer" level="INFO"/>
```
dvlato: should we remove this or just rename with the appropriate package name?

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
```diff
@@ -13,47 +13,60 @@
...
+ *   2. Handles incoming content stream from the network
+ *   3. Handles reactive subscription, cancel, and back pressure
+ *   4. Is not thread safe.
+ *   5. Relies on external scheduler to serialise events.
+ */
+public class FlowControllingHttpContentProducer {
     private static final Logger LOGGER = LoggerFactory.getLogger(FlowControllingHttpContentProducer.class);
     private static final int MAX_DEPTH = 1;
 
-    private final StateMachine<ProducerState> stateMachine;
```
dvlato: This can be kept final

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
```diff
@@ -79,39 +95,44 @@
...
                 .transition(BUFFERING, ContentChunkEvent.class, this::contentChunkInBuffering)
+                .transition(BUFFERING, TearDownEvent.class, this::releaseAndTerminate)
                 .transition(BUFFERING, ChannelInactiveEvent.class, this::releaseAndTerminate)
                 .transition(BUFFERING, ChannelExceptionEvent.class, this::releaseAndTerminate)
                 .transition(BUFFERING, ContentSubscribedEvent.class, this::contentSubscribedInBuffering)
                 .transition(BUFFERING, ContentEndEvent.class, this::contentEndEventWhileBuffering)
 
                 .transition(BUFFERING_COMPLETED, RxBackpressureRequestEvent.class, this::rxBackpressureRequestInBufferingCompleted)
                 .transition(BUFFERING_COMPLETED, ContentChunkEvent.class, this::spuriousContentChunkEvent)
-                .transition(BUFFERING_COMPLETED, ChannelInactiveEvent.class, s-> scheduleTearDown(BUFFERING_COMPLETED))
```
dvlato: is it correct to expect the tearDown will be performed by the timeout?

components/server/src/main/java/com/hotels/styx/server/netty/codec/NettyToStyxRequestDecoder.java
```diff
@@ -63,85 +63,103 @@
...
+        return new FlowControllingHttpContentProducer(
+                () -> ctx.channel().read(),
+                () -> ctx.channel().config().setAutoRead(true),
+                cause -> { },
+                format("%s, %s", loggingPrefix, ""),
+                null,
+                inactivityTimeoutMs);
+    }
+
+    private Throwable toStyxException(Throwable cause) {
```
dvlato: can we use a more meaningful name here? It seems other kinds of Netty exceptions could be caught here but we are only dealing with TooLongFrameException (I don't know why!)

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
OwenLindsell: removed :)

components/client/src/test/unit/resources/logback.xml
OwenLindsell: remove it as the class has moved to a different module

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingPublisher.java
```diff
@@ -0,0 +1,53 @@
...
+import com.hotels.styx.api.Buffer;
+import io.netty.channel.EventLoop;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+/**
+ * A publisher to wrap the FlowControllingHttpContentProducer FSM and perform subscription operations via a Netty EventLoop.
+ */
+public final class FlowControllingPublisher implements Publisher<Buffer> {
```
mikkokar: Frankly, the class name is quite meaningless. Normally I try to avoid debating names, but in this case it is justified to rename this class to `NettyByteStream`. Reasoning below...

We use this class to adapt `FlowControllingHttpContentProducer` to `ByteStream`. This uses uses Netty to source the byte stream content from a TCP socket.

In this context, I'd suggest you to rename this to `NettyByteStreamPublisher`. But because `ByteStream` already is a publisher, therefore `NettyByteStream` is sufficient.

components/common/src/main/java/com/hotels/styx/common/content/QueueDrainingPublisher.java
```diff
@@ -0,0 +1,55 @@
...
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import java.util.concurrent.Executor;
+
+/**
+ * A publisher to wrap the FlowControllingHttpContentProducer FSM and perform subscription operations via a QueueDrainingExecutor.
+ */
+public final class QueueDrainingPublisher implements Publisher<Buffer> {
```
mikkokar: This class works with any executor. Apart from name,  there is nothing specific to queue draining event processor. Therefore you could call `queueDrainingExecutor` just `delegate`, etc. and perhaps rename this class to something more generic like `FlowControllingSubscriptionAdapter` etc? Or something better, actually ;-)

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingPublisher.java
```diff
@@ -0,0 +1,53 @@
...
+
```
mikkokar: Apart from the type it accepts, this class is identical to `QueueDrainingPublisher`.

You can eliminate this class and use the `QueueDrainingPublisher` if you wrap `EventLoop` behind an `Executor` interface.

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
```diff
@@ -78,39 +90,46 @@
...
+        if (uri.startsWith("_")) {
+            fsmHistories.put(uri, "Initial state:BUFFERING");
+        }
         this.stateMachine = new StateMachine.Builder<ProducerState>()
                 .initialState(BUFFERING)
 
                 .transition(BUFFERING, RxBackpressureRequestEvent.class, this::rxBackpressureRequestInBuffering)
                 .transition(BUFFERING, ContentChunkEvent.class, this::contentChunkInBuffering)
+                .transition(BUFFERING, TearDownEvent.class, this::releaseAndTerminate)
                 .transition(BUFFERING, ChannelInactiveEvent.class, this::releaseAndTerminate)
```
mikkokar: The timer is not cancelled in `releaseAndTerminate` transition. Is this correct? My gut feeling is that it should be cancelled (or stopped) in every transition to a terminal state.

components/common/src/main/java/com/hotels/styx/api/exceptions/ResponseTimeoutException.java
```diff
@@ -1,75 +0,0 @@
...
-package com.hotels.styx.api.exceptions;
-
-import com.hotels.styx.api.extension.Origin;
-
-import static java.lang.String.format;
-
-/**
- * An exception due to a response timeout.
- */
-public class ResponseTimeoutException extends TransportException implements IsDeadConnectionException, IsTimeoutException {
```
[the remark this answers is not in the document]
mikkokar: I'll propose that `tearDownResources` would take a `RuntimeException` or `Throwable` as a cause. Then we can pass a `ResponseTimeoutException` or a `ContentTimeoutException` accordingly.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyToStyxResponsePropagator.java
```diff
@@ -166,17 +155,8 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) throws Except
...
         if (evt instanceof IdleStateEvent) {
-            producer.channelInactive(
-                    new ResponseTimeoutException(
-                            origin,
-                            "idleStateEvent",
-                            producer.receivedBytes(),
-                            producer.receivedChunks(),
-                            producer.emittedBytes(),
-                            producer.emittedChunks()));
+            getContentProducer(ctx).tearDownResources("idle timeout");
```
mikkokar: It would better to just pass a cause to `tearDownResources`. Then it would be possible to pass in a `RequestTimeoutException` or `ContentTimeoutException` where appropriate.

components/common/src/main/java/com/hotels/styx/common/content/FlowControllerTimer.java
```diff
@@ -0,0 +1,70 @@
...
+    public void cancel() {
+        currentTimeout.cancel();
+    }
+
+    public void checkActivity() {
+        if (producer.state() != COMPLETED && producer.state() != TERMINATED) {
+            resetTimerOrTearDownFlowController();
+        }
+    }
+
```
mikkokar: We need separate `FlowControllingHttpContentProducer` FSM logic from the timer implementation. But I will create a separate issue for that.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -297,6 +298,7 @@ private static StartupConfig parseStartupConfig(String[] args) {
 
     @Override
     protected void doStart() {
+        RequestDebugger.add(components.environment().configuration().get("debug-request-id-prefix").orElse(null));
```
mikkokar: Should we remove this before merging?

I'm not sure how this would work because styx itself allocates a request ID and we should treat it as a pseudo random string.

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpPipelineHandler.java
```diff
@@ -106,7 +106,7 @@
                     ContentOverflowException.class
             )
             .add(SERVICE_UNAVAILABLE, ResourceExhaustedException.class)
-            .add(GATEWAY_TIMEOUT, ResponseTimeoutException.class)
+            .add(GATEWAY_TIMEOUT, ContentTimeoutException.class)
```
mikkokar: As commented elsewhere. We should keep the `ResponseTimeoutException` around.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/client/ExpiringConnectionSpec.scala
```diff
@@ -73,20 +75,36 @@ class ExpiringConnectionSpec extends FunSpec
   }
 
   it("Should expire connection after 1 second") {
-    val request = get(styxServer.routerURL("/app1")).build()
 
-    val response1 = Mono.from(pooledClient.sendRequest(request, requestContext())).block()
+    // Prime a connection:
+    val response1: HttpResponse = Mono.from(pooledClient.sendRequest(
+      get(styxServer.routerURL("/app1/1"))
+        .header(HttpHeaderNames.TRANSFER_ENCODING, HttpHeaderValues.CHUNKED)
```
mikkokar: Static inline.

system-tests/e2e-suite/src/test/scala/com/hotels/styx/client/ExpiringConnectionSpec.scala
```diff
@@ -73,20 +75,36 @@ class ExpiringConnectionSpec extends FunSpec
...
     }
 
     Thread.sleep(1000)
 
-    val response2 = Mono.from(pooledClient.sendRequest(request, requestContext())).block()
+    // Send a second request. The connection would have been expired after two seconds
+    // Therefore, the pool creates a new connection.
+    val response2: HttpResponse = Mono.from(pooledClient.sendRequest(
+      get(styxServer.routerURL("/app1/2"))
+        .header(HttpHeaderNames.TRANSFER_ENCODING, HttpHeaderValues.CHUNKED)
```
mikkokar: Static inline.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyToStyxResponsePropagator.java
```diff
@@ -156,7 +157,17 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) throws Except
     @Override
     public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
         if (evt instanceof IdleStateEvent) {
-            getContentProducer(ctx).tearDownResources("idle timeout");
+            FlowControllingHttpContentProducer producer = getContentProducer(ctx);
+            getContentProducer(ctx).tearDownResources(
+                new ResponseTimeoutException(
```
mikkokar, line 162: A strictly correct behaviour is to use a `ResponseTimeoutException` whenever the HTTP headers have yet not been received, and a `ContentTimeoutException` otherwise. We could use a presence of `producer` to determine this.

But this retains the existing behaviour, and therefore we can do this as a separate increment. Could you raise a GH issue for this please?

mikkokar:
HI @OwenLindsell there are check style related changes to the `pom.xml` files. They look bit unrelated. Could you add a quick comment to this PR to shed some light on them? 

Just a quick note will serve as a record for the future maintainers. This would be really helpful.

components/client/src/test/unit/java/com/hotels/styx/client/StyxHttpClientTest.java
```diff
@@ -90,9 +90,10 @@ public void tearDown() {
      */
     @Test
     public void cannotBeModifiedAfterCreated() throws ExecutionException, InterruptedException {
-        StyxHttpClient client = new StyxHttpClient.Builder()
-                .userAgent("v1")
-                .build();
+        StyxHttpClient.Builder builder = new StyxHttpClient.Builder().userAgent("v1");
+        StyxHttpClient client = builder.build();
+
+        builder.userAgent("v2");
```
mikkokar, line 96: This looks incorrect. I Looks like this change can be reverted back?

OwenLindsell: that change was reverted yesterday

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/NettyToStyxResponsePropagator.java
OwenLindsell: this is fixed in this branch

OwenLindsell: will do

OwenLindsell:
> HI @OwenLindsell there are check style related changes to the `pom.xml` files. They look bit unrelated. Could you add a quick comment to this PR to shed some light on them?
> 
> Just a quick note will serve as a record for the future maintainers. This would be really helpful.

pom changes were reverted yesterday

Issue #518, opened 2019-11-05
Administrative state for routing objects


mikkokar:
## Description

The origins toggle doesn't work in the backwards compatibility mode (https://github.com/HotelsDotCom/styx/pull/458). Implement a mechanism to substitute this functionality. The origins toggle is described here:

  * https://github.com/HotelsDotCom/styx/blob/master/docs/user-guide/admin-interface.md#origin-toggle

More fundamentally, the routing objects, especially the `HostProxy` objects, should have an administrative state:

* Enabled (administrator enabled)
* Disabled (administrator disabled)

Each `HostProxy` object, or any other object that needs to participate as a load balancing group member, will be tagged with a tag indicating the administrative state.

I propose the following naming convention:

* `state: enabled` - the object is administratively enabled
* `state: disabled` - the object is administratively disabled
* `status: active` - the object is marked active by the health checking function
* `status: inactive` - the object is marked inactive by the health checking function 

The load balancing group will include all objects that are tagged with both `state: enabled` and `status: active`. All other objects are excluded by the load balancer.

## Admin Interface

The origins state is administered via admin interface endpoint.

I propose the following

POST to `/admin/providers/<provider-name>/<app-id>/<origin-id>/state` with property `state` either `enabled` or `disabled` will enable/disable the object as requested.

GET to `/admin/providers/<provider-name>/<app-id>/<origin-id>/state` will return the current object state.

## Implementation considerations

This feature needs to be provided by `YamlFileConfigurationService` and it should expose it via `adminInterfaceHandlers` method that can be overridden from `StyxService` interface.

This functionality can be, but doesn't have to be bridged to the original `/admin/tasks/origins` endpoint.

## Acceptance Criteria

- `LoadBalancingGroup` chooses only those objects that are both administratively enabled, and `active.
- Health check monitor should not monitor disabled object.
- Health check monitor state shouldn't encode its internal state (pass/fail count) in the `state` tag. Instead, it can encode it to some other tag.
- Document the tagging scheme and the two new endpoints in Styx user documentation.

PR #561, opened 2019-12-10
Fix origin reactivation logic for admin interface

chrisgresty:
(See #518 )
Reactivation of an origin should follow the existing logic as described here for the Origin Toggle: https://github.com/HotelsDotCom/styx/blob/master/docs/user-guide/admin-interface.md

i.e. if health checks are enabled then a reactivated origin goes into an `unreachable` state, and it is the health check service that moves it into an `active` state.

### DIFF
--- a/components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
@@ -116,10 +116,7 @@ internal class HealthCheckMonitoringService(
         futureRef.get().cancel(false)
     }
 
-    fun isRunning() : Boolean {
-        val executor = futureRef.get() ?: return false
-        return !executor.isCancelled && !executor.isDone
-    }
+    fun isRunning() = futureRef.get()?.let { !it.isCancelled && !it.isDone } == true
 
     internal fun runChecks(application: String, objectStore: StyxObjectStore<RoutingObjectRecord>) {
         val monitoredObjects = objectStore.entrySet()
--- a/components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
@@ -116,6 +116,11 @@ internal class HealthCheckMonitoringService(
         futureRef.get().cancel(false)
     }
 
+    fun isRunning() : Boolean {
+        val executor = futureRef.get() ?: return false
+        return !executor.isCancelled && !executor.isDone
+    }
+
     internal fun runChecks(application: String, objectStore: StyxObjectStore<RoutingObjectRecord>) {
         val monitoredObjects = objectStore.entrySet()
                 .map { Pair(it.key, it.value) }
--- a/components/proxy/src/main/kotlin/com/hotels/styx/services/OriginsAdminHandler.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/services/OriginsAdminHandler.kt
@@ -135,15 +135,15 @@ internal class OriginsAdminHandler(
         }
     }
 
-    private fun hasActiveHealthCheck(origin: RoutingObjectRecord) : Boolean {
-        val appId = lbGroupTag.find(origin.tags) ?: return false
-        serviceDb.entrySet().firstOrNull { (_, provider) ->
-            provider.type == HEALTH_CHECK_MONITOR
-                    && provider.tags.contains(targetTag(appId))
-                    && (provider.styxService as HealthCheckMonitoringService).isRunning()
-        } ?: return false
-        return true
-    }
+    private fun hasActiveHealthCheck(origin: RoutingObjectRecord) =
+            lbGroupTag.find(origin.tags)?.let(::findActiveHealthCheckMonitor) != null
+
+    private fun findActiveHealthCheckMonitor(appId: String) =
+            serviceDb.entrySet().firstOrNull { (_, provider) ->
+                provider.type == HEALTH_CHECK_MONITOR
+                        && provider.tags.contains(targetTag(appId))
+                        && (provider.styxService as HealthCheckMonitoringService).isRunning()
+            }
 
     private fun errorResponse(status: HttpResponseStatus, message: String) =
             response(status)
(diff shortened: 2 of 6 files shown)


components/proxy/src/main/kotlin/com/hotels/styx/services/OriginsAdminHandler.kt
```diff
@@ -130,6 +135,16 @@ internal class OriginsAdminHandler(
...
 
+    private fun hasActiveHealthCheck(origin: RoutingObjectRecord) : Boolean {
+        val appId = lbGroupTag.find(origin.tags) ?: return false
+        serviceDb.entrySet().firstOrNull { (_, provider) ->
+            provider.type == HEALTH_CHECK_MONITOR
+                    && provider.tags.contains(targetTag(appId))
+                    && (provider.styxService as HealthCheckMonitoringService).isRunning()
+        } ?: return false
+        return true
+    }
```
mikkokar: I would use more "functional" approach:

* This eliminates three `return` statements
* Therefore might be bit easier to follow?
* Bit more idiomatic in relation to our existing Kotlin codebase?

```
private fun hasActiveHealthCheck2(origin: RoutingObjectRecord) = lbGroupTag
        .find(origin.tags)
        ?.let { appId ->
            serviceDb.entrySet()
                    .firstOrNull { (_, provider) ->
                        provider.type == HEALTH_CHECK_MONITOR
                                && provider.tags.contains(targetTag(appId))
                                && (provider.styxService as HealthCheckMonitoringService).isRunning()
                    }
        } == null
```

But this is just my personal preference.

mikkokar: Here is some more info about Kotlin scope functions: https://kotlinlang.org/docs/reference/scope-functions.html

components/proxy/src/test/kotlin/com/hotels/styx/services/OriginsAdminHandlerTest.kt
```diff
@@ -167,10 +171,21 @@ class OriginsAdminHandlerTest : FeatureSpec({
...
+        scenario("Activating an unhealthchecked, inactive origin results in an active state") {
             expectStateChange(STATE_INACTIVE, STATE_ACTIVE, STATE_ACTIVE, false)
         }
 
+        // TODO: Complete this test
+        scenario("Activating a healthchecked, inactive origin results in an unreachable state") {
+            expectStateChange(STATE_INACTIVE, STATE_ACTIVE, STATE_UNREACHABLE, false, "app.origin.hc.running")
+        }
+
+        // TODO: Complete this test
```
mikkokar: 3xTODOs here. Are these still work in progress? Or just old comments?

components/proxy/src/test/kotlin/com/hotels/styx/services/OriginsAdminHandlerTest.kt
```diff
@@ -167,10 +171,21 @@ class OriginsAdminHandlerTest : FeatureSpec({
...
+        scenario("Activating a healthchecked (but the healthchecker is stopped), inactive origin results in an unreachable state") {
+            expectStateChange(STATE_INACTIVE, STATE_ACTIVE, STATE_ACTIVE, false, "app.origin.hc.stopped")
+        }
```
mikkokar: Please remind me, when do we end up stopping a health checker (if that is active)?

components/proxy/src/test/kotlin/com/hotels/styx/services/OriginsAdminHandlerTest.kt
chrisgresty: Ah, these are old. Will remove.

components/proxy/src/test/kotlin/com/hotels/styx/services/OriginsAdminHandlerTest.kt
chrisgresty: In practice, only when the origins config is updated to remove healthchecking configuration for an origin altogether. Which does mean that this case is unlikely. But I was bothered by the fact that the service does support being stopped, and if it was, then there would be nothing to move the state of the origin from `unreachable` to `active`. So, straying into YAGNI territory a bit, I guess. Happy to excise this if you reckon it's not needed.

components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
```diff
@@ -116,6 +116,11 @@ internal class HealthCheckMonitoringService(
         futureRef.get().cancel(false)
     }
 
+    fun isRunning() : Boolean {
+        val executor = futureRef.get() ?: return false
+        return !executor.isCancelled && !executor.isDone
```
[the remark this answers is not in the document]
chrisgresty: A similar comment to below about taking a more functional approach can be made here, too, which removes the name of that `val` altogether.

components/proxy/src/test/kotlin/com/hotels/styx/services/OriginsAdminHandlerTest.kt
mikkokar: > So, straying into YAGNI territory a bit, I guess. Happy to excise this if you reckon it's not needed.

No this is an important consideration. Just that this scenario didn't immediately occur to me.

Based on your response, I understand there are two possible scenarios:

1) Health check configuration is removed from origins configuration.
2) Origins configuration remains unchanged, but somebody stops the service over the admin interface?

Case 1: Removal of health check configuration only:
  - Looking at code, in this case the host proxy is re-tagged with STATE_ACTIVE. Which seems correct. 
  - Looking at `updateRoutingObjects` in `YamlFileConfigurationService`, I do see a possible bug. The update function doesn't take the tags into account. Suppose the tags alone change, but object's YAML configuration remains unchanged. This code doesn't seem to update the underlying routing object record in this case: WDYT?

```
     objectDefs.forEach { objectDef ->
            routeDb.compute(objectDef.name()) { previous ->
                if (previous == null || changed(objectDef.config(), previous.config)) {
                    // Possible optimisation: check if the tags alone changed, or if the configuration changed too.
                    previous?.routingObject?.stop()
                    converter.routingObjectRecord(objectDef)
                } else {
                    previous
                }
            }
        }
```

Case 2: Looking at code, it removes the health check tag, and sets `state=active`.

Are there any other scenarios we'd have to take into account?

chrisgresty: > The update function doesn't take the tags into account. 
One side-effect of doing so is that, even if you haven't actually changed the tags in the config file, but the tags have changed at runtime, then you will reset the tags to the configured set each time you reload the origins file. So if you've deactivated an origin via the admin interface, you will reactivate it by reloading the origins file, even if you didn't change the tag in the origins file.

chrisgresty: The current implementation also resets the tags, but only if there has been a config change for that origin. Which seems a bit more reasonable, even if not perfect.

PR #563, opened 2019-12-11
Add support for compression of responses

dvlato:


system-tests/ft-suite/src/test/kotlin/com/hotels/styx/proxy/CompressionSpec.kt
```diff
@@ -0,0 +1,165 @@
...
+import io.kotlintest.specs.FeatureSpec
+import reactor.core.publisher.toMono
+import java.nio.charset.StandardCharsets.UTF_8
+import java.util.zip.GZIPInputStream
+
+class CompressionSpec : FeatureSpec() {
+
+    init {
+        feature("Content-type of the response is compressible") {
+            scenario("Compresses HTTP response if client requested gzip accept-encoding") {
```
chrisgresty: Is it worth testing that an unsupported accept-encoding scheme results in no compression?

PR #564, opened 2019-12-11
Remove RxJava - Part 5

OwenLindsell:


components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpPipelineHandler.java
```diff
@@ -265,25 +264,29 @@ private State onLegitimateRequest(LiveHttpRequest request, ChannelHandlerContext
...
-                                   }
-
-                                   @Override
-                                   public void onNext(LiveHttpResponse response) {
-                                       eventProcessor.submit(new ResponseReceivedEvent(response, ctx));
-                                   }
-                               }
-                    );
+            Eventual<LiveHttpResponse> responseEventual = httpPipeline.handle(v11Request, newInterceptorContext(ctx));
+            responseEventual.subscribe(new BaseSubscriber<LiveHttpResponse>() {
```
mikkokar: I would consider inlining `responseEventual`.

PR #565, opened 2019-12-11
Remove RxJava - Part 6

OwenLindsell:


components/common/src/main/java/com/hotels/styx/common/CompletableFutures.java
```diff
@@ -1,50 +0,0 @@
...
-
-import io.netty.util.concurrent.Future;
-import rx.Observable;
-
-import java.util.concurrent.CompletableFuture;
-
-/**
- * Helpers for working with CompletableFuture objects.
- */
-public final class CompletableFutures {
```
mikkokar: Good to see this class going away :-D

Issue #550, opened 2019-12-02
Metrics: ttfb


wbautistaEXPEDIA:
Please bring back the `ttfb` metric. It appears to have been removed with https://github.com/HotelsDotCom/styx/commit/8fda1eb27d05787f877bf6677507826a4012230d#diff-5f8166fc3d0368081911ac498da9da7d

mikkokar:
Fixed by #575.

PR #575, opened 2019-12-18
Reinstate time-to-first-byte-ms metric.

chrisgresty:
See #550


components/client/src/main/java/com/hotels/styx/client/connectionpool/SimpleConnectionPool.java
```diff
@@ -223,6 +229,11 @@ public void close() {
 
     @VisibleForTesting
     private class ConnectionPoolStats implements Stats {
+
+        final Histogram timeToFirstByteHistogram = new Histogram(new SlidingWindowReservoir(50));
+        private final Supplier<Long> ttfbSupplier = () -> (long) timeToFirstByteHistogram.getSnapshot().getMean();
+        private final Supplier<Long> memoizedTtfbSupplier = memoizeWithExpiration(ttfbSupplier::get, 5, MILLISECONDS)::get;
+
```
mikkokar: As a documented feature, we need to expose TTFB metric as a histogram. But the supplier here just gets the mean of the recorded values.

For TTFB to be useful as a documented feature, we need to report a full histogram. The mean alone does't really say anything. Histogram on the other hand shows accurately when and how the latencies are affected.

components/client/src/main/java/com/hotels/styx/client/connectionpool/StatsReportingConnectionPool.java
```diff
@@ -91,6 +91,7 @@ private static void registerMetrics(ConnectionPool hostConnectionPool, MetricReg
         scopedRegistry.register("connections-closed", (Gauge<Integer>) () -> (int) stats.closedConnections());
         scopedRegistry.register("connections-terminated", (Gauge<Integer>) () -> (int) stats.terminatedConnections());
         scopedRegistry.register("connections-in-establishment", (Gauge<Integer>) () -> (int) stats.connectionsInEstablishment());
+        scopedRegistry.register("time-to-first-byte-ms", (Gauge<Long>) () -> (long) stats.timeToFirstByteMs());
```
mikkokar: I believe CodaHale adds a time unit in the metric itself. Therefore I believe `-ms` suffix is unnecessary. But better to verify this.

components/client/src/main/java/com/hotels/styx/client/connectionpool/StatsReportingConnectionPool.java
```diff
@@ -107,7 +107,7 @@ private void removeMetrics() {
         MetricRegistry scopedRegistry = getMetricScope(connectionPool);
         asList("busy-connections", "pending-connections", "available-connections", "ttfb",
                 "connection-attempts", "connection-failures", "connections-closed", "connections-terminated",
-                "connections-in-establishment")
+                "connections-in-establishment", "time-to-first-byte-ms")
```
mikkokar: I think "time-to-first-byte-ms" can be removed here?

docs/user-guide/metrics-reference.md
```diff
@@ -191,6 +190,14 @@ The origin side metrics scopes are illustrated in a diagram below:
...
   byte is received.
 
+**origins.`<backend>`.requests.time-to-first-byte**  
+**origins.`<backend>`.`<origin>`.requests.time-to-first-byte**
+
+* A latency distribution of requests to origin.
+* Measured as time to first content byte.
+* Timer started when request is sent, and stopped when the first content
+  byte is received.
+
```
mikkokar: 👍

PR #578, opened 2019-12-23
Spike: Simplify styx servers and inject executors

mikkokar:
## Netty Servers Spike

Goal is to prepare core styx classes to accommodate http2 server implementation.

### Server Configuration

Utilise existing Styx Config Object model and its underpinning framework for configuring Styx servers. This involves adding a new `servers` configuration map like so:

```
servers:
  http:
    type: HttpServer
    config:
      port: 8081
      handler: hello2

  h2:
    type: Http2Server
    config:
      port: 8443
      handler: hello2

routingObjects:
  hello2:
    type: StaticResponseHandler
    config:
      status: 200
      content: "hello-2"
```

Styx http2 will eventually be implemented as a `Http2Server` server object.

Benefits:

* Styx is no longer limited to a pair of `http` or `https` endpoints. Users specify as many servers as they want.

* Simplify & reduce Styx server implementation. We will handle routing objects, control plane providers, and Styx servers with same underlying framework. Thus we can cut down any bespoke code that is specific to the styx servers alone.

* (PR #582) As a part of this work, the `ProviderObjectRecord` has merged into a new `StyxObject<T>` as follows :

```
internal data class StyxObjectRecord<T : StyxService>(
        val type: String,
        val tags: Set<String>,
        val config: JsonNode,
        val styxService: T)

internal typealias ProviderObjectRecord = StyxObjectRecord<StyxService>
```

### Netty Executors

(PR #583)

I have introduced new `NettyExecutor` class. It encapsulates a Netty event loop, and it corresponding client and server channel types. `NettyExecutor.create` attempts to use an Epoll when available, and it reverts to NIO when not available. This replaces the functionality that used to be scattered across `ClientEventLoopFactory`, `EpollClientEventLoopFactory`, `NioEventLoopGroupFactory`, and `PlatformAwareEventLoopGroupFactory` classes.

We can now read the thread pool configuration etc in the `main` method from the styx configuration, instantiate executors accordingly, and inject them in to the class hierarchy.

This reduces a need to pass styx configuration across several components.

### Other improvements

**ProxyServerbuilder**
* Removed. The proxy server can be instantiated almost as easily from NettyServer directly.

PR: #581

**NettyServer**
* Remove `ServerListener` and `ServerSocketBinder` inner classes. This was made possible by earlier work to restrict only one server socket binder per `NettyServer`.

PR: #580


components/proxy/src/main/java/com/hotels/styx/ResponseInfoFormat.java
```diff
@@ -13,20 +13,20 @@
...
+package com.hotels.styx;
 
 import static com.hotels.styx.StyxConfig.NO_JVM_ROUTE_SET;
 import com.hotels.styx.api.LiveHttpRequest;
 
 /**
  * Formats response info into a string.
  */
+
+// TODO: Should be package private, but it is shared by com.hotels.styx.servers.StyxHttpServer.kt
```
OwenLindsell: Not sure we need a TODO here. It's a trivial problem and the TODO will be there forever and will never get fixed. Better would be to fix it by moving StyxHttpServer to com.hotels.styx package or just remove comment.

components/proxy/src/main/kotlin/com/hotels/styx/servers/StyxHttpServer.kt
```diff
@@ -0,0 +1,143 @@
...
+import io.netty.handler.ssl.SslContext
+import org.slf4j.LoggerFactory
+
+object StyxHttpServer {
+    @JvmField
+    val SCHEMA = `object`(
+            field("port", integer()),
+            field("handler", string())
+
+//                optional("tlsSettings", `object`(
```
OwenLindsell: Did you mean to submit commented out code?

components/proxy/src/test/kotlin/com/hotels/styx/servers/StyxHttpServerTest.kt
```diff
@@ -0,0 +1,26 @@
...
+  limitations under the License.
+ */
+package com.hotels.styx.servers
+
+import io.kotlintest.specs.StringSpec
+import io.kotlintest.shouldBe
+
+class StyxHttpServerTest : StringSpec() {
+
+    init {
```
OwenLindsell: Did you mean to commit an empty test?

components/server/src/main/java/com/hotels/styx/IStyxServer.java
```diff
@@ -13,29 +13,21 @@
...
+ * A Styx Server is a StyxService with a server socket.
  */
-public interface ServerEventLoopFactory {
-    /**
-     * EventLoopGroup for establishing new channels.
-     *
-     * @return event loop group
-     */
-    EventLoopGroup newBossEventLoopGroup();
+public interface IStyxServer extends StyxService {
```
OwenLindsell: Why do we need to create this interface?

components/server/src/main/java/com/hotels/styx/server/netty/NettyServer.java
```diff
@@ -94,153 +87,90 @@ public InetSocketAddress httpAddress() {
...
-            try {
-                if (connectorStopper != null) {
-                    connectorStopper.call();
-                }
-            } catch (Exception e) {
-                throw propagate(e);
-            }
+        public Void call() {
+            channelGroup.close().awaitUninterruptibly();
+            // TODO: Mikko: check why the return value is never used:
```
OwenLindsell: Should your name be in the code? :)

PR #579, opened 2019-12-26
speed up kotlin integration tests by reusing the JVM fork.

dvlato:
Latest master build in Travis:  

`[INFO] Styx - System Tests - Functional Tests ............. SUCCESS [02:29 min]`

Travis build for this PR (https://travis-ci.org/HotelsDotCom/styx/builds/629635200?utm_source=github_status&utm_medium=notification) :

`[INFO] Styx - System Tests - Functional Tests ............. SUCCESS [ 32.290 s]`

### DIFF
--- a/system-tests/ft-suite/src/test/kotlin/com/hotels/styx/resiliency/OriginResourcesSpec.kt
+++ b/system-tests/ft-suite/src/test/kotlin/com/hotels/styx/resiliency/OriginResourcesSpec.kt
@@ -50,7 +50,7 @@ class OriginResourcesSpec : StringSpec() {
                     .withStatus(200)
                     .withBody("mock-server-01"))
 
-
+    val initialThreadcount = threadCount("Styx-Client-Worker")
     init {
         "Client thread pool configuration" {
             val clientThreadCount = run {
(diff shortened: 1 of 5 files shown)


system-tests/ft-suite/src/test/kotlin/com/hotels/styx/resiliency/OriginResourcesSpec.kt
```diff
@@ -67,7 +67,7 @@ class OriginResourcesSpec : StringSpec() {
             }
 
             // From the static configuration below
-            clientThreadCount shouldBe 3
+            clientThreadCount-initialThreadcount shouldBe 3
```
mikkokar: We should probably remove this test.
It doesn't look very stable when running in shared JVM in parallel with others.

dvlato: you created this test for a regression a few months ago .so I would hesitate to just remove it... We are not running tests in parallel now, but I guess that if we did we might want to have a similar test where we can inject the pool-name (the prefix) or maybe verify this in a different way (e.g, exposing some reference to the thread pool and verifying it's the same).

Anyway, if we agree to remove it, it's fine for me!

mikkokar: It would be nice to keep this test. But as it is at the moment it doesn't look stable. As you say, we would have to inject some kind of thread name prefix for this particular test. That could be a separate increment.

PR #585, opened 2020-01-09
Support compiling the project with Java 11.

dvlato:


mikkokar:
@dvlato said that the test code had to change due to differences between Java versions 8 and 11.

* Version 11 has a different default garbage collector, which shows up in JVM metrics.
* Time format has got more accurate.
* TLS handshake failure message has changed a bit.

Issue #573, opened 2019-12-18
Sanitise Netty HTTP message in logs 


mikkokar:
## Description

Sanitise selected headers/cookies in Netty HTTP message objects. Sensitive headers/cookies are still logged as a part of Netty exception messages.

Issue #454 sanitises selected headers/cookies from Styx logs. Specifically it masks the headers from Styx HTTP messages. But the header values are still exposed when exceptions from Netty are logged, for example in HttpErrorStatusCauseLogger.

This example is from `HttpErrorStatusCauseLogger`. When the throwable cause is from Netty it sometimes includes full decoded Netty HTTP message object, including possibly sensitive headers/cookies.

```
    @Override
    public void proxyErrorOccurred(Throwable cause) {
        LOG.error("Error occurred during proxying", cause);
    }
```

PR #590, opened 2020-01-14
Sanitize logged headers and cookies from Netty HTTP objects

chrisgresty:
See #573


components/common/src/main/java/com/hotels/styx/common/format/SanitisedHttpHeaderFormatter.java
```diff
@@ -37,8 +38,12 @@
     private final List<String> cookiesToHide;
 
     public SanitisedHttpHeaderFormatter(List<String> headersToHide, List<String> cookiesToHide) {
-        this.headersToHide = requireNonNull(headersToHide);
-        this.cookiesToHide = requireNonNull(cookiesToHide);
+        this.headersToHide = Collections.unmodifiableList(requireNonNull(headersToHide));
```
Noddy76: This doesn't stop the passed lists from being modified it just wraps them in a facade that stops modification methods. To protect this object's state you will need to copy the list contents.

components/common/src/main/java/com/hotels/styx/common/format/SanitisingThrowableFactory.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Wrap a {@link Throwable} in a dynamic proxy that sanitizes its message to hide sensitive cookie
+     * information. The supplied Throwable must be non-final, and must have a no-args constructor. If the proxy
+     * cannot be created for any reason (including that it is proxying a final class, or one without a no-args constructor),
+     * then a warning is logged and the unproxied Throwable is returned back to the caller.
+     * @param target the Throwable to be proxied
+     * @param formatter hides the sensitive cookies.
+     * @return the proxied Throwable, or the supplied Target if it cannot be proxied.
+     */
+    public Throwable create(Throwable target, SanitisedHttpHeaderFormatter formatter) {
```
dvlato: Target seems fine in the context of creating a Proxy, but I wonder if it would a bit clearer if we indicated it's the origin exception. Just nitpicking :)

components/common/src/main/java/com/hotels/styx/common/format/SanitisingThrowableInterceptor.java
```diff
@@ -0,0 +1,103 @@
...
+import java.lang.reflect.Method;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Intercepts methods of a {@link Throwable} so that the message can be modified to sanitize the values of any recognised cookies.
+ * Any cause is similarly wrapped before being returned.
+ */
+public class SanitisingThrowableInterceptor {
```
dvlato: This is a matter of taste I guess, but I think we are exposing too many details with this SanitisingThrowableFactory + SanitisingThrowableInterceptor . Probably I would try to have both classes in the same file  and use simpler names.

chrisgresty: Thanks - yes, a bit of a rethink, and I agree it looks much better.

Issue #408, opened 2019-05-07
Can the proxy node has multi conntecters?


douxiaofeng99:
Can the proxy node has multi conntecters with differenert listen port?

mikkokar:
It currently supports only one proxy connector at a time with http or https port (or both). I think we could remove this limitation if and when time permits.

mikkokar:
This is possible after: #591.

PR #591, opened 2020-01-15
Add HttpServer object

mikkokar:
## Description

Adds a `HttpServer` configuration object. 

Also addresses  #408.

- Now you can start as many servers as necessary, each with different configuration.
- They don't use that global `httpHandler` attribute. Each object have their own `handler` attribute.
- ~~The old `proxy` is still mandatory, and would have to be deprecated in a separate increment.~~
- The old `proxy` is no longer mandatory. It can still be used for setting the global attributes like client thread pool size. In this case it is not necessary to set any connectors.

An example usage:

```
    servers:
      myHttp:
        type: HttpServer
        config:
          port: 0
          handler: nonSecure
                        
      myHttps:
        type: HttpServer
        config:
          port: 0
          handler: secure
          tlsSettings:
            certificateFile: $crtFile
            certificateKeyFile: $keyFile
            sslProvider: JDK

    routingObjects:
      secure:
        type: StaticResponseHandler
        config:
          status: 200
          content: "secure"
                      
      nonSecure:
        type: StaticResponseHandler
        config:
          status: 200
          content: "non-secure"

```

## Configuration

    val SCHEMA = `object`(
            field("port", integer()),
            field("handler", string()),
            optional("tlsSettings", `object`(
                    optional("sslProvider", string()),
                    optional("certificateFile", string()),
                    optional("certificateKeyFile", string()),
                    optional("sessionTimeoutMillis", integer()),
                    optional("sessionCacheSize", integer()),
                    optional("cipherSuites", list(string())),
                    optional("protocols", list(string()))
            )),

            optional("maxInitialLength", integer()),
            optional("maxHeaderSize", integer()),
            optional("maxChunkSize", integer()),

            optional("requestTimeoutMillis", integer()),
            optional("keepAliveTimeoutMillis", integer()),
            optional("maxConnectionsCount", integer()),

            optional("bossThreadsCount", integer()),
            optional("workerThreadsCount", integer())
    )

## Future improvements

* Share the Netty executor with the server objects.
* Implement admin server as a server object.
* ~~Deprecate `proxy` and `httpHandler` configuration attributes.~~
* Default parameter values should be consistent with the rest of the system?


mikkokar:
Update: Pushed a new commit to deprecate `proxy` attribute. It still can be specified, but it is no longer mandatory.

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
```diff
@@ -70,7 +70,7 @@
 
         STYX_SERVER_CONFIGURATION_SCHEMA_BUILDER = newDocument()
                     .rootSchema(object(
-                            field("proxy", object(
+                            optional("proxy", object(
```
dvlato: I think we need to move at least the compressResponses attribute somewhere else if it is optional now. 
Don't we need any other of the attributes? factories, bossThreadsCount ... It seems bossThreadsCount was always  applied to each listener/ServerBooStrap individually (we aren't sharing the NioEventLoopGroup between different BootStraps).

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -181,7 +181,7 @@ public StyxServer(StyxServerComponents components, Stopwatch stopwatch) {
                 .create();
 
         // Startup phase 1: start plugins, control plane providers, and other services:
-       ArrayList<Service> services = new ArrayList<>();
```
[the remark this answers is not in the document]
mikkokar: I don't know.

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
```diff
@@ -115,6 +116,7 @@
                                     field("factories", map(object(opaque())))
                             )),
                             optional("providers", map(routingObject())),
+                            optional("servers", map(routingObject())),
```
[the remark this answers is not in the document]
mikkokar: We should think about the naming before freezing this. However `listeners` sound bit like our routing objects. In fact I was thinking to rename `routingObjects` to `listeners`. Then you would have servers that consume traffic from the network, and listeners to process the traffic.

WDYT?

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
mikkokar: 1. `compressResponses` attribute, and the other missing attributes would still work with the old `proxy` attribute. I will add them as a separate increment to the server object. I can do it straight after the PR is merged.

2. Regarding The executors & thread counts. This is a know limitation and I think mentioned in the PR description. We need to think about a specific mechanism of configuring and injecting event loops into various kinks of styx objects: servers, routing objects, and providers. I'm thinking to introduce a separate `executors` block where you declare named executors, which can be then called into action from the other styx objects. For example:

```
executors:
  ioClient:
    type: NettyEventloop
    config:
       threads: 4
       channelType: client

  ioServer:
    type: NettyEventloop
    config:
      thread: 3
      channelType: server

servers:
  myServer:
    type: HttpServer
    config:
      port: 8080
      handler: hostProxy
      eventLoop: ioServer

routintObjects:
  hostProxy:
    type: HostProxy
    config:
       host: abc:8080
       eventLoop: ioClient

```

WDYT?

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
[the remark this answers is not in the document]
dvlato: We've already discussed this and it seems we can keep using our current names.

components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java
dvlato: Sounds good to me.

mikkokar: Raised an issue for missing content compression: https://github.com/HotelsDotCom/styx/issues/593

PR #592, opened 2020-01-16
Replace RxJava - Part 10

OwenLindsell:
Remove RxJava from tests


system-tests/e2e-suite/src/test/scala/com/hotels/styx/plugins/AggregatingPluginSpec.scala
```diff
@@ -65,8 +63,9 @@ class AggregatingPluginSpec extends FunSpec
     }
 
     it("Gets response from aggregating plugin (with body)") {
+      val chunks : Flux[Buffer] = Flux.fromArray(Seq(chunk("a"), chunk("b"), chunk("c"), chunk("d"), chunk("e")).toArray)
       mockServer.stub("/body", responseSupplier(
-        () => response(OK).body(new ByteStream(toPublisher(Observable.from(Seq(chunk("a"), chunk("b"), chunk("c"), chunk("d"), chunk("e")).asJava)))).build()
+        () => response(OK).body(new ByteStream(chunks)).build()
```
dvlato: We would usually inline the object.

PR #594, opened 2020-01-16
Release artifacts from Travis

dvlato:


travis/deploy.sh
```diff
@@ -0,0 +1,57 @@
...
+
+echo "Deploying Release to sonatype and docker hub"
+#Ensure a correct version was configured in the pom files.
+mvn versions:set -DnewVersion=$TRAVIS_TAG
+#Deploy to sonatype
+mvn deploy --settings travis/mvn-settings.xml -B -U -P sonatype-oss-release,linux -DskipTests=true -Dmaven.test.skip=true
+#Deploy to dockerhub
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+mvn install -f distribution/pom.xml -B -U -P docker -Dstyxcore.docker.image=dlatorre/styx
+docker push dlatorre/styx
```
chrisgresty: Should these lines include 'dlatorre'?

.github/workflows/verifybuild.yml
```diff
@@ -0,0 +1,21 @@
+name: Java CI
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
```
mikkokar: Hi @dvlato. When searching the documentation, it is always checkout@v2 (https://github.com/actions/checkout) that comes out at the top. 

Therefore asking, how did you come about using checkout@v1? Do you know if the v1 and v2 are related in any way? Should be using the v2 since it is newer? Etc.

mikkokar: Okay I see, the examples for Setup-java are using checkout@v1. (https://github.com/actions/setup-java).

mikkokar: And here: https://github.com/actions/cache/blob/master/examples.md#java---maven

travis/deploy.sh
```diff
@@ -0,0 +1,57 @@
...
+echo $GPG_SECRET_KEYS | base64 --decode >>  ${GPG_DIR}/secring.gpg
+
+echo "Deploying Release to sonatype and docker hub"
+#Ensure a correct version was configured in the pom files.
+mvn versions:set -DnewVersion=$TRAVIS_TAG
+#Deploy to sonatype
+mvn deploy --settings travis/mvn-settings.xml -B -U -P sonatype-oss-release,linux -DskipTests=true -Dmaven.test.skip=true
+#Deploy to dockerhub
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+mvn install -f distribution/pom.xml -B -U -P docker -Dstyxcore.docker.image=dlatorre/styx
```
mikkokar: This still has your user name in it:

>  -Dstyxcore.docker.image=dlatorre/styx

Also below:

> docker push dlatorre/styx

Push them to: https://hub.docker.com/repository/docker/hotelsdotcom/styx

travis/mvn-settings.xml
```diff
@@ -1,4 +1,12 @@
 <settings>
+    <mirrors>
+        <mirror>
+            <id>maven-central-mirror</id>
+            <name>Maven central mirror</name>
+            <url>https://maven-central.storage-download.googleapis.com/maven2/</url>
+            <mirrorOf>central</mirrorOf>
```
mikkokar: Hi, what is this mirror for?

travis/deploy.sh
dvlato: I think before that I would need to have credentials to push to hotelsdotcom/styx  :) I think I mentioned quite a few days ago that we could not merge as is but should discuss the approach...

travis/mvn-settings.xml
dvlato: That's the default mirror travis uses if we don't overwrite the settings.xml file (you can see that in any travis logs). When I started using this mvn_settings.xml , some builds failed trying to transfer artifacts from maven central.

.github/workflows/verifybuild.yml
dvlato: see the 'maven guide' was updated from checkout@v1 to @v2 around Christmas time:   https://github.com/actions/starter-workflows/blob/master/ci/maven.yml

Issue #593, opened 2020-01-16
Add content compression support to StyxServer object


mikkokar:
A pull request #591 implemented `HttpServer` styx object. However content compression got forgotten, and should now be added.

A `StyxHttpServer` class implements the `HttpServer` object. It should accept a `compressResponses` configuration attribute. When enabled, it should compress responses sent back from styx.

The new schema would look like this:

    val SCHEMA = `object`(
            field("port", integer()),
            field("handler", string()),
            optional("compressResponses", boolean()),     // <--- HERE!
            optional("tlsSettings", `object`(
                    optional("sslProvider", string()),
                    optional("certificateFile", string()),
                    optional("certificateKeyFile", string()),
                    optional("sessionTimeoutMillis", integer()),
                    optional("sessionCacheSize", integer()),
                    optional("cipherSuites", list(string())),
                    optional("protocols", list(string()))
            )),

            optional("maxInitialLength", integer()),
            optional("maxHeaderSize", integer()),
            optional("maxChunkSize", integer()),

            optional("requestTimeoutMillis", integer()),
            optional("keepAliveTimeoutMillis", integer()),
            optional("maxConnectionsCount", integer()),

            optional("bossThreadsCount", integer()),
            optional("workerThreadsCount", integer())
    )

PR #595, opened 2020-01-18
Add content compression support to StyxServer object

chrisgresty:
See issue #593

### DIFF
--- a/components/proxy/src/main/kotlin/com/hotels/styx/servers/StyxHttpServer.kt
+++ b/components/proxy/src/main/kotlin/com/hotels/styx/servers/StyxHttpServer.kt
@@ -40,6 +40,7 @@ object StyxHttpServer {
     val SCHEMA = `object`(
             field("port", integer()),
             field("handler", string()),
+            optional("compressResponses", bool()),
             optional("tlsSettings", `object`(
                     optional("sslProvider", string()),
                     optional("certificateFile", string()),
@@ -79,6 +80,7 @@ private data class StyxHttpServerTlsSettings(
 private data class StyxHttpServerConfiguration(
         val port: Int,
         val handler: String,
+        val compressResponses: Boolean,
         val tlsSettings: StyxHttpServerTlsSettings?,
 
         val maxInitialLength: Int = 4096,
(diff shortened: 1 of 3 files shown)


chrisgresty:
> I see some of these tests might be equivalent to the ones in CompressionSpec. Do you think we need both?

There are now two different paths to the same compression logic - one via the current StyxServer, and one via the new StyxHttpServer. So both need testing. I don't think it's enough to just test that the correct config is set, the test should prove that the request gets through the server to the compression logic as well. Hence duplicating the tests (more or less).

Then it becomes a question of where. Perhaps it would have been better to extend the CompressionSpec test, but the existing tests for StyxHttpServer were already executing at the kind of level that are needed for this test, so it seemed to me that the expected place to find these tests would be there.

components/proxy/src/main/kotlin/com/hotels/styx/servers/StyxHttpServer.kt
```diff
@@ -79,6 +80,7 @@ private data class StyxHttpServerTlsSettings(
 private data class StyxHttpServerConfiguration(
         val port: Int,
         val handler: String,
+        val compressResponses: Boolean,
```
mikkokar: I believe this defaults to `false`?

components/proxy/src/test/kotlin/com/hotels/styx/servers/StyxHttpServerTest.kt
```diff
@@ -302,12 +344,25 @@ private fun createConnection(port: Int) = NettyConnectionFactory.Builder()
 
 private val response = HttpResponse.response(OK)
         .header("source", "secure")
+        .header("content-type", "text/plain")
         .body("Hello, test!", UTF_8)
         .build()
 
+private val compressedResponse = HttpResponse.response(OK)
```
mikkokar: Static import `response`.

mikkokar:
> I see some of these tests might be equivalent to the ones in CompressionSpec. Do you think we need both?

At the moment, yes. Because there are two separate ways to start servers. But we will, sooner or later (hope sooner), to eliminate the old way.

PR #596, opened 2020-01-21
Expose Netty proxy executor to the plugins.

fantayeneh:
Currently, each plugin is responsible for its own threading model.
The PR exposes the event loop that started handling the request to the plugin so it can processes events in a non-blocking fashion.


components/common/src/main/java/com/hotels/styx/server/HttpInterceptorContext.java
```diff
@@ -30,46 +32,23 @@
...
-     */
-    public HttpInterceptorContext() {
-        this(false, null);
+        this.executor = executor;
     }
 
     // TODO deprecate
     public static HttpInterceptor.Context create() {
-        return new HttpInterceptorContext();
+        return new HttpInterceptorContext(false, null, Executors.newSingleThreadExecutor());
```
mikkokar: An `urlProbe` function in `HealthCheck.kt` calls `create` to create an "empty" context to be passed in to the health checked object. There appears to be other callers too.

Therefore, this change introduces an executor leak unless something is changed. Instead of changing the call sites, I would consider passing in a re-used executor from a static context. Like so:

```
    private static final globalExecutor = Executors.newSingleThreadExecutor();

    //    .. snip ..

    public static HttpInterceptor.Context create() {
        return new HttpInterceptorContext(false, null, globalExecutor);
    }

```

components/proxy/src/test/java/com/hotels/styx/proxy/interceptors/HttpMessageLoggingInterceptorTest.java
```diff
@@ -32,6 +32,8 @@
 import org.mockito.MockitoAnnotations;
 import reactor.core.publisher.Mono;
 
+import java.util.concurrent.Executors;
```
mikkokar: Should we shut down all these executor services we create in the tests?

components/common/src/main/java/com/hotels/styx/server/HttpInterceptorContext.java
mikkokar: OTOH, after discussing with @fantayeneh, this `create()` is should only really be used as convenience method for testing purposes. As such it can be removed from the class, and we can provide a support method in the test module itself. At minimum it should be marked with `@VisibleForTesting` annotation.

For this reason I'll change the call sites accordingly.

mikkokar:
To sum up:

 * @fantayeneh injects plugins to StyxServerComponents. The plugins already have their own worker executors baked in.
* The context now exposes a Netty executor so that a plugin can switch back to it when calling `chain.proceed()`.

@dvlato @OwenLindsell @chrisgresty can some of you look as well, since I have made some modifications.

components/common/src/main/java/com/hotels/styx/server/HttpInterceptorContext.java
dvlato: I would get rid of this method altogether as we can just pass in a newSingleThreadExecutor without creating any thread leaks in our tests.

mikkokar: Hi @dvlato I have now removed all calls to `HttpInterceptorContext.create()` from within the Styx project. I have still left the function there in case some external test code depends on in.

dvlato: Hi,  in this solution we still have to schedule new threads in several of the tests. Why don't we use the calling thread as the "Executor" when nothing has been passed?  I think it would make more sense and it's what we were doing previously, so tests should not be affected.

What I means is that we should be able to pass _Runnable::run_ as the third parameter to create() and in that case the Tests that were changed just to call the method with an executor could be rolled back (or at least changed to use the same 'calling thread executor')

mikkokar: >  Why don't we use the calling thread as the "Executor" when nothing has been passed? I think it would make more sense and it's what we were doing previously, so tests should not be affected.

Because `HttpInterceptorContext.create()` is fundamentally publicly visible test method in a production class. As such it has to go. That's why I have marked it as deprecated and removed all calls to it from within styx code. 

> What I means is that we should be able to pass Runnable::run as the third parameter to create() and in that case the Tests that were changed just to call the method with an executor could be rolled back (or at least changed to use the same 'calling thread executor')

👍 Yes, this makes sense. This would stop the thread leaks from any dependant code, for now.

support/api-testsupport/src/main/java/com/hotels/styx/support/Support.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.concurrent.Executors;
+
+import static java.net.InetSocketAddress.createUnresolved;
+
+public final class Support {
+
+    private Support() {
+    }
+
+    private static final Executor GLOBAL_EXECUTOR = Executors.newSingleThreadExecutor();
```
dvlato: Should we replace this with Runnable::run as discussed?

components/api/src/test/java/com/hotels/styx/api/extension/service/spi/MockContext.java
```diff
@@ -19,9 +19,12 @@
 
 import java.net.InetSocketAddress;
 import java.util.Optional;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 
 class MockContext implements HttpInterceptor.Context {
     static final HttpInterceptor.Context MOCK_CONTEXT = new MockContext();
+    final Executor executor = Executors.newSingleThreadExecutor();
```
dvlato: Should we replace this with Runnable::run as discussed?

components/proxy/src/test/kotlin/com/hotels/styx/Support.kt
```diff
@@ -149,3 +134,35 @@ fun failingMockObject() = mockk<RoutingObject> {
...
+fun Eventual<LiveHttpResponse>.wait(maxBytes: Int = 100*1024, debug: Boolean = false) = this.toMono()
+        .flatMap { it.aggregate(maxBytes).toMono() }
+        .doOnNext {
+            if (debug) {
+                LOGGER.info("${it.status()} - ${it.headers()} - ${it.bodyAs(UTF_8)}")
+            }
+        }
+        .block()
+
+val globalExecutor = Executors.newSingleThreadExecutor()
```
dvlato: Should we replace this with Runnable:run as discussed?

components/server/src/test/java/com/hotels/styx/server/routing/AntlrMatcherTest.java
```diff
@@ -18,14 +18,16 @@
...
 import static com.hotels.styx.api.LiveHttpRequest.get;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 
 public class AntlrMatcherTest {
 
-    private final HttpInterceptorContext contextHttps = new HttpInterceptorContext(true);
-    private final HttpInterceptorContext contextHttp = new HttpInterceptorContext();
+    private final HttpInterceptorContext contextHttps = new HttpInterceptorContext(true, null, Executors.newSingleThreadExecutor());
+    private final HttpInterceptorContext contextHttp = new HttpInterceptorContext(false, null, Executors.newSingleThreadExecutor());
```
dvlato: Should we replace this with Runnable:run as discussed? I don;t know if we can use requestContext()  here.

components/proxy/src/test/java/com/hotels/styx/routing/handlers/StandardHttpPipelineTest.java
```diff
@@ -209,7 +211,7 @@ private HttpInterceptor recordingInterceptor(String name, Consumer<String> onInt
     }
 
     private LiveHttpResponse sendRequestTo(StandardHttpPipeline pipeline) {
-        HttpInterceptor.Context context = new HttpInterceptorContext(InetSocketAddress.createUnresolved("127.0.0.1", 0));
+        HttpInterceptor.Context context = new HttpInterceptorContext(false, InetSocketAddress.createUnresolved("127.0.0.1", 0), Executors.newSingleThreadExecutor());
```
dvlato: Can't we use requestContext() here?

components/server/src/test/java/com/hotels/styx/server/routing/AntlrMatcherTest.java
mikkokar: 👍

components/proxy/src/test/java/com/hotels/styx/routing/handlers/StandardHttpPipelineTest.java
mikkokar: 👍

components/api/src/test/java/com/hotels/styx/api/extension/service/spi/MockContext.java
mikkokar: 👍

components/proxy/src/test/kotlin/com/hotels/styx/Support.kt
mikkokar: 👍

PR #597, opened 2020-01-24
Add fragment() to Url API

Noddy76:
Adds `fragment()` getter to `Url` class, and hides `Optional <UrlQuery> query()` method.

`Url.query()` returns `Optional<UrlQuery>` which is unusable at present because the class is package private. This change exposes the class and required (non-mutating) methods.

### DIFF
--- a/components/api/src/main/java/com/hotels/styx/api/Url.java
+++ b/components/api/src/main/java/com/hotels/styx/api/Url.java
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2019 Expedia Inc.
+  Copyright (C) 2013-2020 Expedia Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


mikkokar:
> Should this URLQuery class be private to "URL" ? We might expose the query as a string from "URL" if the Map of results is not useful for us.

Makes sense. I would rather remove the `Optional<UrlQuery> query()`. The `UrlQuery` is just a glorified map that adds very little value. Can't we just live with `public Map<String, List<String>> queryParams()` as @dvlato suggests?

Noddy76:
> Are you okay with deprecating `Optional<UrlQuery> query()` in favour of `public Map<String, List<String>> queryParams()`? Then we could some day perhaps even remove the `UrlQuery` class.

I'm fine with removing the `query()` method. It's the smallest change and, as noted, we still provide access to the data via other means/methods. I'll update the PR.

PR #610, opened 2020-02-10
Fix Http Message Logging Spec tests that depended on Styx API allowing passing invalid cookies.

dvlato:


system-tests/ft-suite/src/test/kotlin/com/hotels/styx/admin/OriginsFileCompatibilitySpec.kt
```diff
@@ -286,7 +286,7 @@ class OriginsFileCompatibilitySpec : FunSpec() {
                 }
             }
 
-            test("!TLS Settings modifications") {
+            test("TLS Settings modifications") {
```
mikkokar: Shall we back it off?

PR #609, opened 2020-02-07
Remove old application router

mikkokar:
A draft PR for now. Won't merge until we have some consumer feedback.

PR #616, opened 2020-02-12
Shareable executors

mikkokar:
## Summary

This PR allows end users to declare named executors (or threads) and share them between different Styx objects. 

The executors only work with the new `HostProxy` and `HttpServer` objects. The old `proxy` and `admin` servers, and the old Styx Client, which are being phased out, still use the old `clientWorkerThreadsCount`, `workerThreadsCount`, and `bossThreadsCount` settings.

## Configuration

The executors are declared in the executors block at the top-level of Styx configuration file:

```
executors:
  server:
    type: NettyExecutor
    config:
      threads: 4
      namePattern: "Styx-Server-Worker" 
```

The executor can be shared by other objects:

```
routing:
  originA:
    type: HostProxy
    config:
        ...
        executor: "server"

servers:
  https:
    type: HttpServer:
    config:
        ...
        executor: "server"
```

The top level yaml configuration follows the schema:

`optional("executors", map(routingObject())),`

Where `routingObject` is the familiar `type/config/tags` triple. The `NettyExecutor` object follows the schema:

        val SCHEMA = SchemaDsl.`object`(
                SchemaDsl.field("threads", SchemaDsl.integer()),
                SchemaDsl.field("namePattern", SchemaDsl.string())
        )

Styx is started with 3 "global" executors:

* `Styx-Client-Global-Worker`, `StyxHttpServer-Global-Worker`, `StyxHttpServer-Global-Boss`. The routing objects will use them unless some other are specified by `executor` field. Consequently these names are reserved. Styx will always make sure that default global executors are available.

* You can override the default global executor settings by specifying that reserved name in the `executor` configuration. For example, the following configuration will change the thread names for global server worker thread:

```
executors:
  StyxHttpServer-Global-Worker:
    type: NettyExecutor
    config:
      threads: 4
      namePattern: "styx-worker" 
```

Add a new `executor` field to YamlFileConfigurationService configuration. To specify the executor for the created `HostProxy` objects. When `executor` field is absent, it should use the default global executor.


components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -109,8 +111,10 @@ private static StyxServer createStyxServer(String[] args) {
         LOG.info("Styx configFileLocation={}", startupConfig.configFileLocation());
         LOG.info("Styx logConfigLocation={}", startupConfig.logConfigLocation());
 
+        StyxConfig styxConfig = parseConfiguration(startupConfig);
+
         StyxServerComponents components = new StyxServerComponents.Builder()
-                .styxConfig(parseConfiguration(startupConfig))
+                .styxConfig(styxConfig)
```
mikkokar: Unnecessary change.

components/common/src/main/java/com/hotels/styx/NettyExecutor.java
```diff
@@ -70,7 +72,11 @@ private NettyExecutor(EventLoopGroup eventLoopGroup,
     }
 
     public void shut() {
-        eventLoopGroup.shutdownGracefully();
+        try {
+            eventLoopGroup.shutdownGracefully(0, 0, TimeUnit.SECONDS).await(5000);
```
dvlato: Should we wait here?

components/common/src/main/java/com/hotels/styx/NettyExecutor.java
```diff
@@ -70,7 +72,11 @@ private NettyExecutor(EventLoopGroup eventLoopGroup,
...
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
```
dvlato: This is clearing the interrupted flag, we've discussed previously it might not be the best option.

components/proxy/src/main/java/com/hotels/styx/ExecutorFactory.java
```diff
@@ -0,0 +1,35 @@
...
+public interface ExecutorFactory {
+    /**
+     * Create an executor instance.
+     *
+     * @param name                 Executor name
+     * @param configuration        Styx executor configuration
+     *
+     * @return Styx service instance
+     */
+    NettyExecutor create(String name, JsonNode configuration);
```
dvlato: Should we try not to not pass a JsonNode here? It seems we keep propagating our Yaml based config.

components/proxy/src/test/kotlin/com/hotels/styx/services/YamlFileConfigurationServiceTest.kt
```diff
@@ -65,7 +66,7 @@ class YamlFileConfigurationServiceTest : FunSpec() {
         try {
             action(service)
         } finally {
-            service.stop().join()
+            service.stop().get(2, SECONDS)
```
dvlato: This could throw several exceptions here... makes me think we might want to implement Autocloseable.

components/common/src/main/java/com/hotels/styx/NettyExecutor.java
mikkokar: Hi. It does wait for `shutdownGracefully` to come back.

Or do you refer to the `quietPeriod` setting? But we never really supported a graceful shutdown properly. Yet. I think we need to implement that on a "styx level".

The reason for immediate quiet and grace period is to avoid excessive waiting in functional test suite.

However I do think, going forward, we need to add timeout controls and a shutdown future to this API. Perhaps that should be a next increment, providing that the work is completed before the Styx 1.1 release. WDYT?

components/common/src/main/java/com/hotels/styx/NettyExecutor.java
mikkokar: Thanks. A good catch :-)

components/proxy/src/main/java/com/hotels/styx/ExecutorFactory.java
mikkokar: This code follows the idiomatic fashion of creating styx objects from yaml/json nodes. Sadly the styx object model relies entirely on yaml and/or JsonNode based configuration.

I have created a a proof of concept that lifts this restriction. Check the details in PR: #609 (Remove old application router).

Till then, it needs to stay in this "idiomatic" form.

components/proxy/src/test/kotlin/com/hotels/styx/services/YamlFileConfigurationServiceTest.kt
mikkokar: You mean, for the `StyxService`?

components/common/src/main/java/com/hotels/styx/NettyExecutor.java
mikkokar: Fixed.

components/common/src/main/java/com/hotels/styx/NettyExecutor.java
mikkokar: The executors will only ever be closed when styx shuts. Therefore any shutdown future, etc doesn't really matter at the moment as the JVM is going to shut anyway. I would revisit this code if or once we'll allow consumers to shut injected executors.

components/proxy/src/main/java/com/hotels/styx/routing/config/Builtins.java
```diff
@@ -220,4 +231,24 @@ public static InetServer buildServer(
 
         return constructor.create(name, context, serverDef.config(), serverDb);
     }
+
+    /**
+     * Builds a Styx server.
```
chrisgresty: Should this javadoc refer to Netty Executors rather than Styx Servers?

components/proxy/src/main/java/com/hotels/styx/startup/StyxServerComponents.java
```diff
@@ -96,27 +100,61 @@ private StyxServerComponents(Builder builder) {
 
         this.executor = NettyExecutor.create("Styx-Client-Worker", environment.configuration().proxyServerConfig().clientWorkerThreadsCount());
 
+        // Overwrite any existing or user-supplied values:
```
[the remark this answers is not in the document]
mikkokar: Sure

components/proxy/src/main/java/com/hotels/styx/routing/config/Builtins.java
mikkokar: 👍

PR #629, opened 2020-02-18
Add support for null computations inside StyxObjectStore.compute()

dvlato:


components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
```diff
@@ -210,25 +216,16 @@ internal class ObjectDisappearedException : RuntimeException("Object disappeared
...
-                if (prevTags != newTags)
-                    it.copy(tags = newTags)
-                else
-                    previous
-            }
-        } catch (e: ObjectDisappearedException) {
-            // Object disappeared between the ifPresent check and the compute, but we don't really mind.
-            // We just want to exit the compute, to avoid re-creating it.
-            // (The ifPresent is not strictly required, but a pre-emptive check is preferred to an exception)
+    db.compute(name) { previous ->
```
chrisgresty: The same construction is used further up too, on about line 100 or so - this can probably also be reworked in the same way, and the `ObjectDisappearedException` removed.

components/api/src/main/java/com/hotels/styx/api/NettyCookie.java
```diff
@@ -69,7 +69,7 @@ public String toString() {
             buf.append(", HTTPOnly");
         }
         if (sameSite != null) {
-            buf.append(", SameSite=").append(sameSite.toString());
+            buf.append(", SameSite=").append(sameSite);
```
mikkokar: This is out of place.

components/proxy/src/main/kotlin/com/hotels/styx/services/HealthCheckMonitoringService.kt
```diff
@@ -206,29 +203,18 @@ internal fun objectHealthFrom(state: String?, health: Pair<String, Int>?) =
...
+        if (previous == null) {
+            null
```
mikkokar: Perhaps we could use `if (previous === null) return@compute null` construct there too? At least we would be idiomatic in its use?

components/api/src/main/java/com/hotels/styx/api/NettyCookie.java
dvlato: I missed pushing that small change, so I just let it go through this time.  I can rollback if you prefer it.

PR #631, opened 2020-02-20
Fix a race condition in ServerObjectSpec.

mikkokar:
There is a small race in ServerObjectSpec. The server objects have not yet been started before the test attempts to read its port number. Therefore wrap the API call in an eventually block.

### DIFF
--- a/system-tests/ft-suite/src/test/kotlin/com/hotels/styx/server/ServerObjectSpec.kt
+++ b/system-tests/ft-suite/src/test/kotlin/com/hotels/styx/server/ServerObjectSpec.kt
@@ -82,7 +82,7 @@ class ServerObjectSpec : FeatureSpec() {
                 styxServer.restart()
 
                 // 1. Query server addresses from the admin interface
-                val httpPort = eventually(2.seconds, AssertionError::class.java) {
+                val httpPort = eventually(2.seconds) {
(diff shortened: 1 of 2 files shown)


system-tests/ft-suite/src/test/kotlin/com/hotels/styx/server/ServerObjectSpec.kt
```diff
@@ -80,15 +82,20 @@ class ServerObjectSpec : FeatureSpec() {
...
-                val httpPort = testClient.send(get("/admin/servers/myHttp/port").header(HOST, styxServer().adminHostHeader()).build())
-                        .wait()
-                        .bodyAs(UTF_8)
-                        .toInt()
-
-                val httpsPort = testClient.send(get("/admin/servers/myHttps/port").header(HOST, styxServer().adminHostHeader()).build())
-                        .wait()
-                        .bodyAs(UTF_8)
-                        .toInt()
+                val httpPort = eventually(2.seconds, AssertionError::class.java) {
```
dvlato: This should not generate assertions, so we should be able to use  eventually(2.seconds) {} . Right now eventually is implemented less efficiently when passing Exceptions that should be ignored...

mikkokar: Looks like we can just drop the 2nd exception/error class argument.

Issue #635, opened 2020-02-26
Expose styx provider object database in YAML via admin interface


chrisgresty:
The Admin interface should display the Provider Object database in a YAML format, in response to a request to `/admin/providers/objects`, similar to the Routing Object database displayed on `/admin/routing/objects'.

The routing object database is not currently displayed as valid YAML - the individual objects are valid, but they are separated by `---` lines. They should be displayed as a valid YAML map or list instead. This should be fixed as part of this Issue, and the new provider object page should be displayed similarly.

bestokes:
Fixed in the above merge.

PR #637, opened 2020-02-28
Expose styx provider object database in YAML via admin interface

chrisgresty:
See #635

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProviderRoutingHandler.java
+++ b/components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProviderRoutingHandler.java
@@ -15,12 +15,12 @@
  */
 package com.hotels.styx.admin.handlers;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
-import com.google.common.collect.ImmutableList;
 import com.hotels.styx.StyxObjectRecord;
 import com.hotels.styx.api.Eventual;
 import com.hotels.styx.api.HttpInterceptor;
@@ -30,13 +30,13 @@ import com.hotels.styx.api.WebServiceHandler;
 import com.hotels.styx.api.configuration.ObjectStore;
 import com.hotels.styx.api.extension.service.spi.StyxService;
 import com.hotels.styx.common.http.handler.HttpStreamer;
-import com.hotels.styx.routing.config.StyxObjectDefinition;
 import com.hotels.styx.routing.db.StyxObjectStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 import static com.fasterxml.jackson.core.JsonParser.Feature.AUTO_CLOSE_SOURCE;
@@ -90,12 +90,10 @@ public class ProviderRoutingHandler implements WebServiceHandler {
     private UrlPatternRouter buildRouter(ObjectStore<? extends StyxObjectRecord<? extends StyxService>> db) {
         UrlPatternRouter.Builder routeBuilder = new UrlPatternRouter.Builder(pathPrefix)
             .get("", new ProviderListHandler(db))
-            .get("objects", (request, context) -> {
-                return handleRequestForAllObjects(request, context, db);
-            })
+            .get("objects", (request, context) -> handleRequestForAllObjects(db))
             .get("objects/:objectName", (request, context) -> {
                 String name = placeholders(context).get("objectName");
-                return handleRequestForOneObject(request, context, db, name);
+                return handleRequestForOneObject(db, name);
             });
 
         db.entrySet().forEach(entry -> {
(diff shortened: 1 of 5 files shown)


components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProviderRoutingHandler.java
```diff
@@ -67,16 +89,90 @@ private void refreshRoutes(ObjectStore<? extends StyxObjectRecord<? extends Styx
...
         return routeBuilder.build();
     }
+
+    private Eventual<HttpResponse> handleRequestForAllObjects(HttpRequest request, HttpInterceptor.Context context,
+                                                              ObjectStore<? extends StyxObjectRecord<? extends StyxService>> db) {
+        List<StyxObjectDefinition> objects = db.entrySet()
+                .stream()
+                .map(entry -> createObjectDef(entry.getKey(), entry.getValue()))
+                .collect(Collectors.toList());
+        String output = serialise(objects);
```
mikkokar: An opinionated comment, but I would inline the `objects` because it is only ever passed on to `serialise` method. ;-)

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProviderRoutingHandler.java
```diff
@@ -67,16 +89,90 @@ private void refreshRoutes(ObjectStore<? extends StyxObjectRecord<? extends Styx
...
+
+        return Eventual.of(response(OK)
+                .body(output, UTF_8)
+                .build());
+    }
+
+    private Eventual<HttpResponse> handleRequestForOneObject(HttpRequest request, HttpInterceptor.Context context,
```
mikkokar: The `request` and `context` arguments are never used.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/RoutingObjectHandler.java
```diff
@@ -68,10 +64,11 @@
...
         urlRouter = new UrlPatternRouter.Builder()
                 .get("/admin/routing/objects", (request, context) -> {
-                    String output = routeDatabase.entrySet()
+                    List<StyxObjectDefinition> objects = routeDatabase.entrySet()
                             .stream()
-                            .map(entry -> serialise(entry.getKey(), entry.getValue()))
-                            .collect(joining("\n"));
+                            .map(entry -> createObjectDef(entry.getKey(), entry.getValue()))
+                            .collect(Collectors.toList());
+                    String output = serialise(objects);
```
mikkokar: Opinionated: I would inline `objects` because it is only ever passed, unconditionally, as an argument to `objects`. Therefore it just adds noise. :-)

mikkokar: Possibly below, too.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/admin/ProviderAdminInterfaceSpec.kt
```diff
@@ -119,6 +119,45 @@ class ProviderAdminInterfaceSpec : FeatureSpec() {
...
+                    - name: "mySecondMonitor"
+                      type: "HealthCheckMonitor"
+                      tags: []
+                      config:
+                        objects: "bbb"
+                        path: "/healthCheck/y"
+                        timeoutMillis: 250
+                        intervalMillis: 500
+                        healthyThreshold: 3
+                        unhealthyThreshold: 2
```
mikkokar: We should add here another object. So that the list would render at least two or more objects.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/routing/RoutingRestApiSpec.kt
```diff
@@ -120,20 +120,18 @@ class RoutingRestApiSpec : StringSpec() {
...
+                                  tags: []
+                                  config:
+                                    status: 200
+                                    content: "Responder"
+                                - name: "root"
+                                  type: "StaticResponseHandler"
+                                  tags: []
+                                  config:
+                                    status: 200
+                                    content: "Root"
```
mikkokar: Hmm... 🤔 

So conceptually this is a map. Also in styx yaml configuration this is represented as a map. But now this implementation renders it as a list.

For consistency, I would prefer the admin interface also to render it as a map.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProviderRoutingHandler.java
chrisgresty: Hm. In this case I find it flows better as it is, because the stream construction is (in my mind) quite different to the method call, and inlining just results in my eye bouncing around a little more than I'd like.
Ideally I'd have liked to be able to follow the .collect() with .serialise() (or some construction with ::serialise) but that's not available. Visually, I think this is the next best thing. But I'll change it (and the others) if you feel moderately strongly about it.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ProviderRoutingHandler.java
chrisgresty: Not exactly sure why I left them in.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/routing/RoutingRestApiSpec.kt
chrisgresty: Yes, of course. I should have noticed that the config yaml uses maps, not lists here.

chrisgresty: For the case where we are requesting a single object, should it be as now (with `name` as an attribute) or in a map format (with `name` as the key)?

chrisgresty: Eh, forget that, it should be the single object payload with no `name:` attribute and no key.

system-tests/ft-suite/src/test/kotlin/com/hotels/styx/admin/ProviderAdminInterfaceSpec.kt
```diff
@@ -119,6 +119,64 @@ class ProviderAdminInterfaceSpec : FeatureSpec() {
...
+                      config:
+                        originsFile: "${originsFile.absolutePath}"
+                        ingressObject: "pathPrefixRouter"
+                        monitor: true
+                        pollInterval: "PT0.1S"
+                    """.trimIndent()
+            }
+
+            scenario("YAML configuration for a single provider is available") {
+                val body = styxServer.adminRequest("/admin/providers/objects/myMonitor")
```
mikkokar: A small nag: IMHO the url should be `/admin/provider/objects` (The `provider` component should be in singular). `/admin/providers/objects` sounds wrong because both `providers` and `objects` are plural.

Do you think we could change it as a part of this PR, or would this warrant a separate PR?

chrisgresty: I'll change it as part of this PR, it's such a small change.

PR #642, opened 2020-03-09
Remove guava - part 1

chrisgresty:
Replace usage of most classes in `com.google.common.base`
Note that `styx-api` doesn't have a dependency on `styx-common`, so some replacements are inlined here rather than using a common function as elsewhere.


components/common/src/main/java/com/hotels/styx/common/Preconditions.java
```diff
@@ -51,6 +53,54 @@ public static String checkNotEmpty(String value) {
...
+     * @throws NullPointerException if the check fails and either {@code errorMessageTemplate} or
+     *     {@code errorMessageArgs} is null (don't let this happen)
+     */
+    public static void checkArgument(boolean expression,
+                                     String errorMessageTemplate,
+                                     Object... errorMessageArgs) {
+        if (!expression) {
+            throw new IllegalArgumentException(format(errorMessageTemplate, errorMessageArgs));
+        }
+    }
```
mikkokar: Some `checkArgument` calls had been replaced with `if/else` statements earlier on in this diff. Perhaps those earlier changes can be backed off now?

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerConfig.java
```diff
@@ -62,8 +60,8 @@ private AdminServerConfig(Builder builder) {
         MetricsCache(
                 @JsonProperty("enabled") Boolean enabled,
                 @JsonProperty("expirationMillis") Integer expirationMillis) {
-            this.enabled = firstNonNull(enabled, false);
-            this.expirationMillis = firstNonNull(expirationMillis, 10_000);
+            this.enabled = enabled != null ? enabled : enabled;
```
mikkokar: `this.enabled = enabled != null ? enabled : enabled;`

Is this correct? Should it say `this.enabled = enabled != null ? enabled : false;`?

components/common/src/main/java/com/hotels/styx/common/Preconditions.java
chrisgresty: The `if/else` statements should only be in `styx/api`, which doesn't have a dependency on `styx/common`

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerConfig.java
chrisgresty: No, it's not - thanks.

components/api/src/main/java/com/hotels/styx/api/RequestCookie.java
```diff
@@ -51,8 +48,14 @@
      * @param value cookie value
      */
     private RequestCookie(String name, String value) {
-        checkArgument(!isNullOrEmpty(name), "name cannot be null or empty");
-        requireNonNull(value, "value cannot be null");
+        if (name == null || name.isEmpty()) {
+            throw new IllegalArgumentException("name cannot be null or empty");
+        }
```
mikkokar: Can we use our own `checkArgument` here?

components/api/src/main/java/com/hotels/styx/api/RequestCookie.java
```diff
@@ -51,8 +48,14 @@
...
+
+        if (value == null) {
+            throw new NullPointerException("value cannot be null");
+        }
```
mikkokar: We can still use `requireNonNull` here?

components/api/src/main/java/com/hotels/styx/api/RequestCookie.java
```diff
@@ -92,7 +95,9 @@ public static RequestCookie requestCookie(String name, String value) {
      * @return "Cookie" header value
      */
     public static String encode(Collection<RequestCookie> cookies) {
-        checkArgument(!cookies.isEmpty(), "Cannot create cookie header value from zero cookies");
+        if (cookies.isEmpty()) {
+            throw new IllegalArgumentException("Cannot create cookie header value from zero cookies");
+        }
```
mikkokar: Also `checkArgument` here?

components/api/src/main/java/com/hotels/styx/api/RequestCookie.java
```diff
@@ -51,8 +48,14 @@
      * @param value cookie value
      */
     private RequestCookie(String name, String value) {
-        checkArgument(!isNullOrEmpty(name), "name cannot be null or empty");
-        requireNonNull(value, "value cannot be null");
+        if (name == null || name.isEmpty()) {
+            throw new IllegalArgumentException("name cannot be null or empty");
+        }
+
+        if (value == null) {
```
Noddy76: Java 8 has an `Objects.requireNonNull()` method to do this.
https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-

components/api/src/main/java/com/hotels/styx/api/Url.java
```diff
@@ -119,7 +116,7 @@ public boolean isSecure() {
      */
     public boolean isFullyQualified() {
         Optional<String> host = host();
-        return host.isPresent() && !isNullOrEmpty(host.get());
+        return host.isPresent() && !host.get().isEmpty();
```
Noddy76: This is a change in behaviour since a `null` host would have returned `false` before. Is this expected?

components/api/src/main/java/com/hotels/styx/api/extension/Origin.java
```diff
@@ -88,11 +87,15 @@ public static Builder newOriginBuilder(Origin origin) {
...
+        if (ids.size() != origins.size()) {
+            throw new IllegalArgumentException("Duplicate ids in " + origins);
+        }
 
         Set<String> hosts = origins.stream().map(Origin::hostAndPortString).collect(toSet());
 
-        checkArgument(hosts.size() == origins.size(), "Duplicate host and port in " + origins);
+        if (hosts.size() != origins.size()) {
+            throw new IllegalArgumentException("Duplicate host and port in " + origins);
+        }
```
mikkokar: Can we use our own `checkArgument` in this file?

components/common/src/main/java/com/hotels/styx/common/Preconditions.java
mikkokar: Ok.

I have left some other comments about `checkArgument`. Please disregard them, unless the files are of course outside of API module.

mikkokar: I have now deleted my comments about them.

components/api/src/main/java/com/hotels/styx/api/UrlQuery.java
```diff
@@ -102,10 +101,13 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return toStringHelper(this)
-                .add("parameters", parameters)
-                .add("encodedQuery", encodedQuery)
-                .toString();
+        StringBuilder sb = new StringBuilder(64);
```
Noddy76: Purly subjective but you could chain the invocations together in a fluent style. IMO would look nicer.

components/api/src/main/java/com/hotels/styx/api/RequestCookie.java
mikkokar: We can still use `requireNonNull` here?

components/api/src/main/java/com/hotels/styx/api/extension/service/ConnectionPoolSettings.java
```diff
@@ -44,12 +41,12 @@
                            @Deprecated Integer socketTimeoutMillis,
                            Integer pendingConnectionTimeoutMillis,
                            Long connectionExpirationSeconds) {
-        this.maxConnectionsPerHost = firstNonNull(maxConnectionsPerHost, DEFAULT_MAX_CONNECTIONS_PER_HOST);
-        this.maxPendingConnectionsPerHost = firstNonNull(maxPendingConnectionsPerHost, DEFAULT_MAX_PENDING_CONNECTIONS_PER_HOST);
-        this.connectTimeoutMillis = firstNonNull(connectTimeoutMillis, DEFAULT_CONNECT_TIMEOUT_MILLIS);
-        this.socketTimeoutMillis = firstNonNull(socketTimeoutMillis, DEFAULT_SOCKET_TIMEOUT_MILLIS);
-        this.pendingConnectionTimeoutMillis = firstNonNull(pendingConnectionTimeoutMillis, DEFAULT_CONNECT_TIMEOUT_MILLIS);
-        this.connectionExpirationSeconds = firstNonNull(connectionExpirationSeconds, DEFAULT_CONNECTION_EXPIRATION_SECONDS);
+        this.maxConnectionsPerHost = maxConnectionsPerHost != null ? maxConnectionsPerHost : DEFAULT_MAX_CONNECTIONS_PER_HOST;
```
Noddy76: Just a suggestion.
```
import static java.util.Optional.ofNullable;

this.maxConnectionsPerHost = ofNullable(maxConnectionsPerHost).orElse(DEFAULT_MAX_CONNECTIONS_PER_HOST);
```

components/client/src/main/java/com/hotels/styx/client/OriginRestrictionLoadBalancingStrategy.java
```diff
@@ -88,7 +86,7 @@ public OriginRestrictionLoadBalancingStrategy(ActiveOrigins activeOrigins, LoadB
     }
 
     private Stream<String> regularExpressionStream(String cookieValue) {
-        return stream(COOKIE_SPLITTER.split(cookieValue).spliterator(), false);
+        return Arrays.stream(cookieValue.split(",")).map(String::trim);
```
Noddy76: Depending on where this is used, splitting using a regular expression might be too expensive. Maybe use `StringTokenizer`.

components/api/src/main/java/com/hotels/styx/api/extension/service/ConnectionPoolSettings.java
```diff
@@ -44,12 +41,12 @@
                            @Deprecated Integer socketTimeoutMillis,
                            Integer pendingConnectionTimeoutMillis,
```
mikkokar: I would just add a private method in this file:

```    
    private <T> T firstNonNull(T a, T b) {
        return a != null ? a : b;
    }
```

Just to avoid accidentally changing the initialisation logic.

components/api/src/main/java/com/hotels/styx/api/extension/service/ConnectionPoolSettings.java
```diff
@@ -74,12 +71,12 @@ public ConnectionPoolSettings(int maxConnectionsPerHost,
...
-        this.connectTimeoutMillis = firstNonNull(connectTimeoutMillis, DEFAULT_CONNECT_TIMEOUT_MILLIS);
-        this.socketTimeoutMillis = firstNonNull(socketTimeoutMillis, DEFAULT_SOCKET_TIMEOUT_MILLIS);
-        this.pendingConnectionTimeoutMillis = firstNonNull(pendingConnectionTimeoutMillis, DEFAULT_CONNECT_TIMEOUT_MILLIS);
-        this.connectionExpirationSeconds = firstNonNull(connectionExpirationSeconds, DEFAULT_CONNECTION_EXPIRATION_SECONDS);
+        this.maxConnectionsPerHost = maxConnectionsPerHost;
+        this.maxPendingConnectionsPerHost = maxPendingConnectionsPerHost;
+        this.connectTimeoutMillis = connectTimeoutMillis;
+        this.socketTimeoutMillis = socketTimeoutMillis;
+        this.pendingConnectionTimeoutMillis = pendingConnectionTimeoutMillis;
+        this.connectionExpirationSeconds = connectionExpirationSeconds;
```
mikkokar: This constructor is marked as deprecated. Perhaps this is an opportunity to remove it? AFAICS it is only called from within this file.

components/api/src/main/java/com/hotels/styx/api/extension/service/HealthCheckConfig.java
```diff
@@ -63,10 +61,18 @@ private HealthCheckConfig(Optional<String> uri,
...
 
-        checkArgument(this.intervalMillis >= 1, format("intervalMillis [%s] cannot be < 1 ms", intervalMillis));
-        checkArgument(this.timeoutMillis >= 1, format("timeoutMillis [%s] cannot be < 1 ms", intervalMillis));
-        checkArgument(this.healthyThreshold >= 1, format("healthyThreshold [%s] cannot be < 1", healthyThreshold));
-        checkArgument(this.unhealthyThreshold >= 1, format("unhealthyThreshold [%s] cannot be < 1", unhealthyThreshold));
+        if (this.intervalMillis < 1) {
+            throw new IllegalArgumentException(format("intervalMillis [%s] cannot be < 1 ms", intervalMillis));
+        }
+        if (this.timeoutMillis < 1) {
+            throw new IllegalArgumentException(format("timeoutMillis [%s] cannot be < 1 ms", intervalMillis));
```
mikkokar: This should be `timeoutMillis` above. But I can see this mistake is also in the original code.

components/api/src/main/java/com/hotels/styx/api/Url.java
chrisgresty: `isPresent()` is checking for null, so I took the opportunity to simplify other check.

components/common/src/main/java/com/hotels/styx/common/Preconditions.java
```diff
@@ -51,6 +53,54 @@ public static String checkNotEmpty(String value) {
         return reference;
     }
 
+    /**
```
mikkokar: If any of this code is sourced from Guava, it should be attributed accordingly.

components/common/src/main/java/com/hotels/styx/common/Strings.java
```diff
@@ -15,17 +15,17 @@
...
 
-/**
- * Common uses of Guava's Joiner class.
- */
-public final class Joiners {
-    public static final Joiner JOINER_ON_COMMA = Joiner.on(",").skipNulls();
-    public static final Joiner JOINER_ON_SEMI_COLON = Joiner.on(";").skipNulls();
-    public static final Joiner JOINER_ON_SPACE = Joiner.on(" ").skipNulls();
-    public static final Joiner JOINER_ON_NEW_LINE = Joiner.on("\n").skipNulls();
+    public static final boolean isNullOrEmpty(String s) {
```
mikkokar: If this is sourced from Guava, perhaps it should be attributed accordingly?

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/ConfigurationParser.java
```diff
@@ -53,7 +53,9 @@ public C parse(ConfigurationSource provider) {
 
         PlaceholderResolutionResult<C> resolved = configuration.resolvePlaceholders(overrides);
 
-        checkState(resolved.unresolvedPlaceholders().isEmpty(), "Unresolved placeholders: %s", resolved.unresolvedPlaceholders());
+        if (!resolved.unresolvedPlaceholders().isEmpty()) {
+            throw new IllegalStateException(format("Unresolved placeholders: %s", resolved.unresolvedPlaceholders()));
+        }
```
mikkokar: Use `checkState` here?

components/api/src/main/java/com/hotels/styx/api/extension/service/ConnectionPoolSettings.java
chrisgresty: ... although having seen @Noddy76 alternative below, I will use that instead.

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/ConfigurationParser.java
chrisgresty: `checkState` from where?

components/api/src/main/java/com/hotels/styx/api/extension/service/HealthCheckConfig.java
chrisgresty: Thanks, I missed that.

components/client/src/main/java/com/hotels/styx/client/OriginRestrictionLoadBalancingStrategy.java
chrisgresty: Maybe so. It's a bit more fiddly to shoehorn it into a stream, but doable.

components/common/src/main/java/com/hotels/styx/common/Strings.java
chrisgresty: In all honesty, I have been writing implementations of this method for many years. This one differs from guava, in that it uses `s.isEmpty()` rather than `s.length() == 0`.

components/api/src/main/java/com/hotels/styx/api/UrlQuery.java
chrisgresty: You are right, and much as I'd love to say "but there are so many to change" I think that's actually an argument for getting it done. So I'll get it done.

components/api/src/main/java/com/hotels/styx/api/Url.java
mikkokar: Yes I think Optional<> cannot store null values.

components/api/src/main/java/com/hotels/styx/api/UrlQuery.java
mikkokar: Hi guys. I believe there are many classes which do not frankly deserve a `toString` method. I would only convert for true data classes.

components/client/src/main/java/com/hotels/styx/client/OriginRestrictionLoadBalancingStrategy.java
mikkokar: This code is in the critical path. It would be called once for each HTTP request, subject to configuration.

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/ConfigurationParser.java
mikkokar: oh forget my comment.

components/server/src/main/java/com/hotels/styx/server/netty/NettyServerBuilder.java
```diff
@@ -51,7 +50,7 @@ public static NettyServerBuilder newBuilder() {
     }
 
     String host() {
-        return firstNonNull(host, "localhost");
+        return host != null ? host : "localhost";
```
mikkokar: You could apply @Noddy76 's `Optional.orElse()` technique here.

components/server/src/main/java/com/hotels/styx/server/netty/NettyServerConfig.java
```diff
@@ -215,49 +210,65 @@ public Builder httpPort(int port) {
 
         @JsonProperty("bossThreadsCount")
         public T setBossThreadsCount(Integer bossThreadsCount) {
-            this.bossThreadsCount = bossThreadsCount;
+            if (bossThreadsCount != null) {
+                this.bossThreadsCount = bossThreadsCount;
+            }
```
mikkokar: This looks really weird.
But let's keep as it is for now.

components/server/src/main/java/com/hotels/styx/server/netty/NettyServerBuilder.java
chrisgresty: True, and also in AdminServerConfig - I think these are the last ones.

Issue #613, opened 2020-02-11
HttpRequestOperation hardening


mikkokar:
`HttpRequestOperation` `execute` needs to have an `else` block to emit an `OriginUnreachableException` when the `nettyConnection` is not connected. Otherwise the returned Flux never emits anything.

        Flux<LiveHttpResponse> responseFlux = Flux.create(sink -> {
            if (nettyConnection.isConnected()) {
                RequestBodyChunkSubscriber bodyChunkSubscriber = new RequestBodyChunkSubscriber(request, nettyConnection);
                requestRequestBodyChunkSubscriber.set(bodyChunkSubscriber);
                addProxyBridgeHandlers(nettyConnection, sink);
                new WriteRequestToOrigin(sink, nettyConnection, request, bodyChunkSubscriber)
                        .write();
                if (requestLoggingEnabled) {
                    httpRequestMessageLogger.logRequest(request, nettyConnection.getOrigin());
                }
            }
        });

PR #647, opened 2020-04-01
Throw exception when connection is not connected

Noddy76:
Fixes: #613.

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/HttpRequestOperation.java
+++ b/components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/HttpRequestOperation.java
@@ -15,6 +15,22 @@
  */
 package com.hotels.styx.client.netty.connectionpool;
 
+import static com.hotels.styx.api.HttpHeaderNames.HOST;
+import static com.hotels.styx.api.extension.service.BackendService.DEFAULT_RESPONSE_TIMEOUT_MILLIS;
+import static io.netty.handler.codec.http.LastHttpContent.EMPTY_LAST_CONTENT;
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.slf4j.LoggerFactory.getLogger;
+
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/HttpRequestOperation.java
```diff
@@ -141,6 +142,8 @@ private static boolean requestIsOngoing(RequestBodyChunkSubscriber bodyChunkSubs
                 if (requestLoggingEnabled) {
                     httpRequestMessageLogger.logRequest(request, nettyConnection.getOrigin());
                 }
+            } else {
+                throw new OriginUnreachableException(nettyConnection.getOrigin(), null);
```
mikkokar, line 146: We use `OriginUnreachableException` specifically when a TCP connection establishment fails, and we use `TransportLostException` when an existing connection has unexpectedly drops or disconnects.

At `execute` method, we assume that `nettyConnection` is already connected. Therefore, we should emit a `TransportLostException` instead.

mikkokar: Instead of throwing an exception, I would emit it with `sink.error(new TransportLostException(...)`. This makes it immediately obvious that the exception will propagate through a `Flux` pipeline instead of unwinding the stack.

OTOH this doesn't affect program correctness because `Flux` subscriber will do this for us anyway.

components/client/src/main/java/com/hotels/styx/client/netty/connectionpool/HttpRequestOperation.java
```diff
@@ -37,25 +54,10 @@
...
-import java.util.concurrent.atomic.AtomicReference;
-
-import static com.hotels.styx.api.HttpHeaderNames.HOST;
-import static com.hotels.styx.api.extension.service.BackendService.DEFAULT_RESPONSE_TIMEOUT_MILLIS;
-import static io.netty.handler.codec.http.LastHttpContent.EMPTY_LAST_CONTENT;
-import static java.lang.String.format;
-import static java.util.Objects.requireNonNull;
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
-import static org.slf4j.LoggerFactory.getLogger;
-
```
mikkokar, line 58: Can you back off the unnecessary change please?

PR #653, opened 2020-04-14
Remove guava from styx-api

chrisgresty:
Remove the Guava dependency from `styx-api`
Reinstate Guava code in the other Styx components, that was removed in #642


chrisgresty:
This work is no longer required, as Guava removal is not currently a priority when the size of the required changes is considered. If this work is picked up again in future, then the most relevant commit is [1e1597b](https://github.com/HotelsDotCom/styx/pull/653/commits/1e1597b1bdfc0084de82cffe3e722f1a854b1319):
- Replaces simple Guava methods with inline native Java constructs where possible
- Introduces package-private `Collections` utility classes for handling the more complex collections and streams-related logic. The classes are package-private to avoid exposing them to API users, and are also duplicated to some extent because the logic is required in two packages.

There are a small number of Guava usages in `styx-api` that have not yet been removed:
```
import com.google.common.annotations.VisibleForTesting --- Url.java:18
import com.google.common.net.HostAndPort --- Origin.java:18
import com.google.common.collect.MapDifference --- AbstractRegistry.java:18
import static com.google.common.collect.Maps.difference --- AbstractRegistry.java:30
import static com.google.common.collect.Maps.filterKeys --- AbstractRegistry.java:31
```

PR #656, opened 2020-04-23
Clean up requests handling and string generation

Noddy76:


Noddy76:
`requested` is part of the reactive contract so we need to keep that. `receivedChunks` and `emittedChunks` are part of the metrics gathering and are not operational. I would like to remove them into a new class at some point that is responsible for recording metrics to leave the main class with a more clearly defined scope.

`AtomicLong`s are mostly a convenance wrapper around a `volatile` and that is exactly how we are using them. Requesting more via the `Subscription` API can happen asynchronously with the FSM transition.

Happy to discuss further on a call and summarise here.

mikkokar:
> `requested` is part of the reactive contract so we need to keep that. `receivedChunks` and `emittedChunks` are part of the metrics gathering and are not operational. I would like to remove them into a new class at some point that is responsible for recording metrics to leave the main class with a more clearly defined scope.
> 

Oh, sorry. I suffered a monetary loss of thought. Of course we need to keep `requested` count. 

What I really meant to say, is that instead of decrementing `requested` at every emission, we'd simply cumulatively incrementing that while maintaining the following relation:

`emittedChunks <= requested`

components/common/src/main/java/com/hotels/styx/common/content/FlowControllingHttpContentProducer.java
```diff
@@ -424,7 +410,9 @@ public void tearDownResources(Throwable cause) {
     }
 
     public void request(long n) {
-        stateMachine.handle(new RxBackpressureRequestEvent(n));
+        if (Operators.validate(n)) {
+            stateMachine.handle(new RxBackpressureRequestEvent(n));
+        }
```
mikkokar, line 415: This validation needs to go to the FSM transition, because:

* The input validation applies to non-terminal states only [Subscription 3.9](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#3.9). In this case, in all but COMPLETED and TERMINATED states.

* According to [Subscription 3.9](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#3.9) an error signal must be emitted to the `contentSubscriber`. But `Operators.validate` just appears to log it and ignore it.

* Performing the check at event injector is also problematic for debugging and visibility. It's because the illegal event is now invisible to the FSM. Therefore it won't be seen by any built in tracing logic etc built into the `StateMachine` class.

* The future maintenance concern: This will impact event ordering if/when `FlowControllingHttpContentProducer`  takes the ownership of serialising events for the `StateMachine`. Even if this never happened, the FSM model could not be "lifted and shifted" elsewhere because part of its behaviour is in the event injectors.

mikkokar: I originally implemented the event injector methods to reduce clutter in the calling code. They are:

* newChunk
* lastHttpContent
* channelException/Inactive/etc
* tearDownResources
* request
* onSubscribed

They are problematic because of their dual responsibility:
* They are factory methods to create event instances
* They also feed the event into the FSM

But it is not clear to new maintainers what functionality should go to the event injectors, and what should go to the state transition handlers.

Therefore I'll propose to separate these two concerns: To turn the event injectors to pure factory methods, and to add a separate `handle(Event)` method for feeding the event into the state machine. 

Therefore, the `request(long)` would become:

```
    public static Event request(long n) {
        return new RxBackpressureRequestEvent(n);
    }

    ...

    public void handle(Event e) {
        stateMachine.handle(e);
    }
```

Then the calling code becomes:

```
    producer.handle(request(initialCount));
```

PR #658, opened 2020-05-13
Clean up PathPrefixRouter and rewrite now broken tests.

Noddy76:
Move the routing object code into a separate class.
Change the ConcurrentSkipList into an array.

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
+++ b/components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
@@ -83,7 +83,7 @@ public class PathPrefixRouter implements RoutingObject {
         return CompletableFuture.allOf(stopFutures);
     }
 
-    private static class PrefixRoute implements Comparable<PrefixRoute> {
+    static class PrefixRoute implements Comparable<PrefixRoute> {
         private final String prefix;
         private final RoutingObject routingObject;
 
--- a/components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
+++ b/components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
@@ -24,14 +24,16 @@ import com.hotels.styx.api.LiveHttpResponse;
 import com.hotels.styx.config.schema.Schema;
 import com.hotels.styx.infrastructure.configuration.yaml.JsonNodeConfig;
 import com.hotels.styx.routing.RoutingObject;
+import com.hotels.styx.routing.RoutingObjectReference;
 import com.hotels.styx.routing.config.Builtins;
 import com.hotels.styx.routing.config.RoutingObjectFactory;
 import com.hotels.styx.routing.config.StyxObjectConfiguration;
 import com.hotels.styx.routing.config.StyxObjectDefinition;
+import com.hotels.styx.routing.config.StyxObjectReference;
 import com.hotels.styx.server.NoServiceConfiguredException;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
@@ -53,9 +55,9 @@ import static java.util.Objects.requireNonNull;
  * The destination can be a routing object reference or an inline definition.
  */
 public class PathPrefixRouter implements RoutingObject {
-    private final Iterable<PrefixRoute> routes;
+    private final PrefixRoute[] routes;
 
-    PathPrefixRouter(Iterable<PrefixRoute> routes) {
+    PathPrefixRouter(PrefixRoute[] routes) {
         this.routes = routes;
     }
 
@@ -74,9 +76,11 @@ public class PathPrefixRouter implements RoutingObject {
 
     @Override
     public CompletableFuture<Void> stop() {
-        List<CompletableFuture<Void>> stopFutures = new ArrayList<>();
-        routes.forEach(prefixRoute -> stopFutures.add(prefixRoute.routingObject.stop()));
-        return CompletableFuture.allOf(stopFutures.toArray(new CompletableFuture[stopFutures.size()]));
+        CompletableFuture[] stopFutures = new CompletableFuture[routes.length];
+        for (int i = 0; i < routes.length; i++) {
+            stopFutures[i] = routes[i].routingObject.stop();
+        }
+        return CompletableFuture.allOf(stopFutures);
     }
 
     private static class PrefixRoute implements Comparable<PrefixRoute> {
@@ -112,13 +116,14 @@ public class PathPrefixRouter implements RoutingObject {
                 throw missingAttributeError(configBlock, join(".", fullName), "routes");
             }
 
-            List<PrefixRoute> routes = new ArrayList<>();
+            PrefixRoute[] routes = new PrefixRoute[config.routes().size()];
+            int i = 0;
             for (PathPrefixConfig routeConfig : config.routes()) {
                 String prefix = routeConfig.prefix();
                 RoutingObject route = Builtins.build(singletonList(""), context, routeConfig.destination());
-                routes.add(new PrefixRoute(prefix, route));
+                routes[i++] = new PrefixRoute(prefix, route);
             }
-            routes.sort(null);
+            Arrays.sort(routes);
 
             return new PathPrefixRouter(routes);
         }
--- a/components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
+++ b/components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
@@ -31,14 +31,11 @@ import com.hotels.styx.routing.config.StyxObjectDefinition;
 import com.hotels.styx.server.NoServiceConfiguredException;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.*;
+import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
```diff
@@ -47,42 +43,61 @@
...
-                    field("destination", routingObject()))
-            ))
-    );
-
-    private final ConcurrentSkipListMap<String, RoutingObject> routes = new ConcurrentSkipListMap<>(
-                    comparingInt(String::length)
-                            .reversed()
-                            .thenComparing(naturalOrder())
-    );
+public class PathPrefixRouter implements RoutingObject {
```
mikkokar, line 55: The underlying routing object shouldn't be public. Because they are always created via the Factory class. You could make it a package private for the unit tests.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
```diff
@@ -47,42 +43,61 @@
...
+        @Override
+        public int compareTo(@NotNull PrefixRoute o) {
+            if (prefix.length() > o.prefix.length()) {
+                return -1;
+            } else if (prefix.length() < o.prefix.length()) {
+                return 1;
+            } else {
+                return prefix.compareTo(o.prefix);
+            }
+        }
```
mikkokar, line 100: Please look if you can use the `PathTrie` class (https://github.com/HotelsDotCom/styx/blob/master/components/server/src/main/java/com/hotels/styx/server/PathTrie.java) instead of this ordered list.

The ordered list was copy-pasted from the old monolithic application router code. It is beyond me why it was originally implemented like that. Could have been something to do with origin reloading. But this is just a guess.

I originally used the ordered list approach here to ensure identical behaviour with the old application router. I didn't have time to deal with potential incompatibilities.

Using the `PathTrie`:
* Might be more performant for larger configurations because it is based on hash lookups instead of linear search.
* Eliminates this ordered list as a yet another alternative path prefix lookup implementation.
* IIRC is bit more flexible in terms of it mapping capabilities.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
```diff
@@ -97,29 +112,25 @@ public RoutingObject build(List<String> fullName, Context context, StyxObjectDef
...
-                    pathPrefixRouter.routes.forEach((route, routingObject) -> routingObject.stop());
-                    return completedFuture(null);
-                }
-            };
+            List<PrefixRoute> routes = new ArrayList<>();
+            for (PathPrefixConfig routeConfig : config.routes()) {
+                String prefix = routeConfig.prefix();
+                RoutingObject route = Builtins.build(singletonList(""), context, routeConfig.destination());
+                routes.add(new PrefixRoute(prefix, route));
+            }
```
mikkokar, line 120: Please keep using the functional stream/map based approach to build the `routes` list. Because that is a standard idiom used elsewhere in the Styx codebase, and I believe it is good to keep it consistent.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
Noddy76: The constructor is package private so instances of this class cannot be created (outside of the package) except via the factory.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
Noddy76: `PathTrie` is an interesting class but not a suitable drop in here since it will match each path section as a whole whilst this class is designed to match each character. In this routing object you can match `/sa1` and `sa2` with the prefix `/sa` whilst you cannot with `PathTrie`.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
Noddy76: Since the `routes` structure is purely internal I was thinking of replacing it with an array which would be easier for the CPU to iterate over in the main function of this class. I'll update the PR and we can discuss that implementation.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
mikkokar: I suspected it might not be 100% compatible.

However I do believe this class is not matching each character by design. Instead it is an unintended consequence of the chosen implementation.

mikkokar: The [Styx manual](https://github.com/HotelsDotCom/styx/blob/master/docs/user-guide/configure-origins.md) for the old application router merely says that:

> path: the beginning of a (relative) URI used by Styx to determine which backend service to route a request to.

So what do you make of it? Would you think of "beginning of an URI" being a proper parent URI path, or is it just a string arbitrarily chopped off at some point?

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
mikkokar: In the idiomatic styx code you would do it something like this:

```
            ArrayList<PrefixRoute> x = config.routes().stream()
                    .map(route -> new PrefixRoute(
                            route.prefix(),
                            Builtins.build(singletonList(""), context, route.destination())))
                    .sorted()
                    .collect(toCollection(ArrayList::new));
```

components/proxy/src/test/java/com/hotels/styx/routing/handlers/PathPrefixRouterTest.java
```diff
@@ -0,0 +1,134 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class PathPrefixRouterTest {
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    public Map<StyxObjectReference, RoutingObject> routingObjects = new HashMap<>();
+    public RoutingObjectFactory.Context Routingcontext;
```
mikkokar, line 49: The variable name is capitalised: `RoutingContext`. Is this intentional?

Noddy76: No. It is not. WIll fix.

components/proxy/src/test/java/com/hotels/styx/routing/handlers/PathPrefixRouterTest.java
```diff
@@ -0,0 +1,134 @@
...
+        testRequestRoute(router, "/foo/x", fooPathHandler);
+        testRequestRoute(router, "/foo/", fooPathHandler);
+        testRequestRoute(router, "/foo/bar", fooBarFileHandler);
+        testRequestRoute(router, "/foo/bar/x", fooBarPathHandler);
+        testRequestRoute(router, "/foo/bar/", fooBarPathHandler);
+        testRequestRoute(router, "/foo/baz/", fooBazFileHandler);
+        testRequestRoute(router, "/foo/baz/y", fooBazFileHandler);
+    }
+
+    private PathPrefixRouter buildRouter(Map<String, String> prefixRoutes) {
```
mikkokar, line 115: Make `PrefixRoute` package private. Then you can simply invoke object's constructor, instead of having to go through the factory.

mikkokar: But this won't work because `Builtins.build` uses a lambda for an object reference lookup:

```
    public static RoutingObject build(...)
        ...
        } else if (configNode instanceof StyxObjectReference) {
            return (request, httpContext) -> context.refLookup()
                    .apply((StyxObjectReference) configNode)
                    .handle(request, httpContext);
        ...
```

At the moment this lambda could be promoted to a separate class. Perhaps like so:

```
public class ObjectRefLookup implements RoutingObject {
    private final RouteRefLookup routeRefLookup;
    private final StyxObjectReference configNode;

    public ObjectRefLookup(RouteRefLookup routeRefLookup, StyxObjectReference configNode) {
        this.routeRefLookup = routeRefLookup;
        this.configNode = configNode;
    }

    @Override
    public Eventual<LiveHttpResponse> handle(LiveHttpRequest request, HttpInterceptor.Context context) {
        return routeRefLookup.apply(configNode)
                .handle(request, context);
    }
}
```

But sadly this class would have to have a public visibility due to our packaging layout.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
Noddy76: The attempt here is to do a simple cleanup and refactor of the code without any functional change. I hope to make it easier to read and easier to refactor later. The behavior should stay exactly the same.

components/proxy/src/main/java/com/hotels/styx/routing/handlers/PathPrefixRouter.java
mikkokar: Ok. Lets explore some options as a future enhancement.

PR #663, opened 2020-05-15
Custom Via Header Support

wbautistaEXPEDIA:
Adds ability to configure a custom `via` header value for Styx server.

Sample configuration:

```
proxy:
  via: myStyxApp
```

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/BuiltInInterceptors.java
+++ b/components/proxy/src/main/java/com/hotels/styx/BuiltInInterceptors.java
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2019 Expedia Inc.
+  Copyright (C) 2013-2020 Expedia Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -50,10 +50,14 @@ final class BuiltInInterceptors {
             builder.add(new HttpMessageLoggingInterceptor(longFormatEnabled, httpMessageFormatter));
         }
 
+        final HttpInterceptor viaInterceptor = config.get("via")
+                .map(ViaHeaderAppendingInterceptor::new)
+                .orElseGet(ViaHeaderAppendingInterceptor::new);
+
         builder.add(new TcpTunnelRequestRejector())
                 .add(new ConfigurationContextResolverInterceptor(EMPTY_CONFIGURATION_CONTEXT_RESOLVER))
                 .add(new UnexpectedRequestContentLengthRemover())
-                .add(new ViaHeaderAppendingInterceptor())
+                .add(viaInterceptor)
                 .add(new HopByHopHeadersRemovingInterceptor())
                 .add(new RequestEnrichingInterceptor(config.styxHeaderConfig()));
 
--- a/components/proxy/src/main/java/com/hotels/styx/BuiltInInterceptors.java
+++ b/components/proxy/src/main/java/com/hotels/styx/BuiltInInterceptors.java
@@ -50,14 +50,12 @@ final class BuiltInInterceptors {
             builder.add(new HttpMessageLoggingInterceptor(longFormatEnabled, httpMessageFormatter));
         }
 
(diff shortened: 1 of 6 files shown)


components/proxy/src/main/java/com/hotels/styx/BuiltInInterceptors.java
```diff
@@ -50,10 +50,14 @@ private BuiltInInterceptors() {
             builder.add(new HttpMessageLoggingInterceptor(longFormatEnabled, httpMessageFormatter));
         }
 
+        final HttpInterceptor viaInterceptor = config.get("via")
+                .map(ViaHeaderAppendingInterceptor::new)
+                .orElseGet(ViaHeaderAppendingInterceptor::new);
```
mikkokar, line 55: AFAICS this adds a new `via` config option in the styx server configuration. Right?

In that case:

* It needs to be added to the `ServerConfigSchema` file that validates the configuration. Otherwise styx wouldn't recognise the new `via` option, and would refuse to start:

```
styx_1       | 11:07:53.040 [main] ERROR com.hotels.styx.StyxServer - Styx server failed to start due to configuration error in file [/styx/config/styxconf.yml]: Unexpected field: 'via'
```

 If you don't see this error, you may have disabled the config validation with `validateServerConfig` system property.

* Rather than adding the new `via` option in the configuration root, it should go under the `proxy` server settings. This way, when we'll start properly supporting multiple server ports, the via header could be configured separately for each connector. I would add it here:

```
        STYX_SERVER_CONFIGURATION_SCHEMA_BUILDER = newDocument()
                    .rootSchema(object(
                            optional("proxy", object(
                                    optional("compressResponses", bool()),

                                    ...

                                    optional("via", string())    // <-- here
                            )),
```

* In fact, styx already supports multiple HTTP servers. Check the `StyxHttpServer.kt` file. When this feature becomes de-facto way of starting styx servers, the StyxHttpServer configuration block would be the only way to configure the new `via` header.

components/proxy/src/test/java/com/hotels/styx/proxy/interceptors/ViaHeaderAppendingInterceptorTest.java
```diff
@@ -97,4 +97,17 @@ public void appendsViaHeaderValueAtEndOfListInResponse() throws Exception {
...
+        final String customVia = "MyAwesomeProxy";
+        final ViaHeaderAppendingInterceptor localInterceptor = new ViaHeaderAppendingInterceptor(customVia);
+        final LiveHttpRequest request = post("/foo")
+                .header(HOST, "www.example.com:8000")
+                .build();
+
+        final LiveHttpRequest interceptedRequest = interceptRequest(localInterceptor, request);
+
+        assertThat(interceptedRequest.headers().get(VIA), isValue("1.1 " + customVia));
+    }
```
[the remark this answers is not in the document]
wbautistaEXPEDIA: 👍 I will update the test.

components/proxy/src/main/java/com/hotels/styx/BuiltInInterceptors.java
wbautistaEXPEDIA: > AFAICS this adds a new `via` config option in the styx server configuration. Right?
> 

Yes, this is correct.

> * It needs to be added to the `ServerConfigSchema` file that validates the configuration. Otherwise styx wouldn't recognise the new `via` option, and would refuse to start:
> 
> ```
> styx_1       | 11:07:53.040 [main] ERROR com.hotels.styx.StyxServer - Styx server failed to start due to configuration error in file [/styx/config/styxconf.yml]: Unexpected field: 'via'
> ```
> 
> If you don't see this error, you may have disabled the config validation with `validateServerConfig` system property.

Will do.

> * Rather than adding the new `via` option in the configuration root, it should go under the `proxy` server settings. This way, when we'll start properly supporting multiple server ports, the via header could be configured separately for each connector.

This makes sense, and I almost put it there to begin with, but opted to defer to review. I will move it.

wbautistaEXPEDIA:
Updated to incorporate feedback.

PR #668, opened 2020-06-02
Updated examples of Styx Core Plugins to micrometer meter registry

JulioDeLeon:
This change takes any examples of Core Styx plugins to use Micrometer meters instead of the Codahale implementation.

### DIFF
--- a/components/api/pom.xml
+++ b/components/api/pom.xml
@@ -84,13 +84,6 @@
       <artifactId>reactor-test</artifactId>
       <scope>test</scope>
     </dependency>
-
-
-    <dependency>
-      <groupId>io.micrometer</groupId>
-      <artifactId>micrometer-registry-prometheus</artifactId>
-        <version>1.5.1</version>
-    </dependency>
   </dependencies>
 
   <build>
--- a/components/api/pom.xml
+++ b/components/api/pom.xml
@@ -86,6 +86,11 @@
     </dependency>
 
 
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-registry-prometheus</artifactId>
+        <version>1.5.1</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/components/proxy/src/main/java/com/hotels/styx/proxy/plugin/InstrumentedPlugin.java
+++ b/components/proxy/src/main/java/com/hotels/styx/proxy/plugin/InstrumentedPlugin.java
@@ -15,7 +15,6 @@
  */
 package com.hotels.styx.proxy.plugin;
 
-import com.codahale.metrics.Meter;
 import com.hotels.styx.api.Environment;
 import com.hotels.styx.api.Eventual;
 import com.hotels.styx.api.HttpHandler;
@@ -25,6 +24,8 @@ import com.hotels.styx.api.LiveHttpResponse;
 import com.hotels.styx.api.plugins.spi.Plugin;
 import com.hotels.styx.api.plugins.spi.PluginException;
 import com.hotels.styx.common.SimpleCache;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Metrics;
 import org.slf4j.Logger;
 import reactor.core.publisher.Flux;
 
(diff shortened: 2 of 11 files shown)


components/proxy/src/main/java/com/hotels/styx/proxy/plugin/InstrumentedPlugin.java
```diff
@@ -42,21 +43,23 @@
...
 
     public InstrumentedPlugin(NamedPlugin plugin, Environment environment) {
         this.plugin = requireNonNull(plugin);
         requireNonNull(environment);
 
+        Metrics.addRegistry(environment.meterRegistry());
+
         this.errorStatusMetrics = new SimpleCache<>(statusCode ->
-                environment.metricRegistry().meter("plugins." + plugin.name() + ".response.status." + statusCode.code()));
+                Metrics.counter("plugins." + plugin.name() + ".response.status." + statusCode.code()));
```
chrisgresty, line 57: I'd suggest using a tag for the plugin name, rather than prefixing the name with "plugins.<name>". 
Perhaps also the status code should be a tag, but I'm less sure about this.
Then you end up with the meter just being a counter called "plugin.response" with tags for plugin name and status code. WDYT?

components/proxy/src/main/java/com/hotels/styx/proxy/plugin/InstrumentedPlugin.java
```diff
@@ -42,21 +43,23 @@
...
 
         this.exceptionMetrics = new SimpleCache<>(type ->
-                environment.metricRegistry().meter("plugins." + plugin.name() + ".exception." + formattedExceptionName(type)));
+                Metrics.counter("plugins." + plugin.name() + ".exception." + formattedExceptionName(type)));
```
chrisgresty, line 60: Similarly here, plugin name and exception class could be tags, leaving the meter being called "plugin.exception"

components/proxy/src/main/java/com/hotels/styx/proxy/plugin/InstrumentedPlugin.java
JulioDeLeon: To be idiomatic with the usage of tags, I think this change would be reasonable.

components/proxy/src/main/java/com/hotels/styx/proxy/plugin/InstrumentedPlugin.java
```diff
@@ -42,21 +43,23 @@
...
-    private final Meter errors;
+    private final SimpleCache<HttpResponseStatus, Counter> errorStatusMetrics;
+    private final SimpleCache<Class<? extends Throwable>, Counter> exceptionMetrics;
+    private final Counter errors;
 
     public InstrumentedPlugin(NamedPlugin plugin, Environment environment) {
         this.plugin = requireNonNull(plugin);
         requireNonNull(environment);
 
+        Metrics.addRegistry(environment.meterRegistry());
```
chrisgresty, line 54: You shouldn't need to add the environment registry to the global registry. This is not so much a problem for test code, but for application code you should just work with the registry that you are given. Then to create metrics you can use `Counter.builder(...)....register(registry)`

components/proxy/src/test/java/com/hotels/styx/proxy/plugin/InstrumentedPluginTest.java
```diff
@@ -76,128 +72,132 @@ public void setUp() {
...
                 aChain.proceed(request)
                         .map(response -> responseWithNewStatusCode(response, INTERNAL_SERVER_ERROR)));
 
         LiveHttpResponse response = Mono.from(plugin.intercept(someRequest, chain)).block();
 
         assertThat(response.status(), is(INTERNAL_SERVER_ERROR));
-        assertThat(metricRegistry.meter("plugins.replaceStatusCode.response.status.500").getCount(), is(1L));
-        assertThat(metricRegistry.meter("plugins.replaceStatusCode.errors").getCount(), is(1L));
+
+        assertThat(Metrics.counter("plugins.response.status", "plugin", pluginName, "status", "500").count(), is(1.0));
```
chrisgresty, line 84: You can retrieve the counter from your metricRegistry with `metricRegistry.find(name).tags(tags).counter()`. It's a bit more long-winded, but then you don't need to link your registry with the global registry in `InstrumentedPlugin`

components/api/pom.xml
```diff
@@ -86,6 +86,11 @@
     </dependency>
 
 
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-registry-prometheus</artifactId>
+        <version>1.5.1</version>
```
chrisgresty, line 92: Is this dependency required?

PR #671, opened 2020-06-08
Dropwizard meter facades

Noddy76:


components/common/src/main/java/com/hotels/styx/api/metrics/codahale/CodaHaleMetricRegistry.java
```diff
@@ -65,50 +128,76 @@ public MetricRegistry scope(String name) {
 
     @Override
     public <T extends Metric> T register(String name, T metric) throws IllegalArgumentException {
-        return metricRegistry.register(name, metric);
+        Tags tags = Tags.of("metricSource", "dropwizard");
```
chrisgresty, line 131: Unused here, and I think you handle this in addAttributes anyway.

Noddy76: Done

PR #675, opened 2020-06-17
Add the prometheus scraping endpoint.

ThaerAlAjlouni:


components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -223,6 +226,7 @@ private HttpHandler adminEndpoints(StyxConfig styxConfig, StartupConfig startupC
         builder.add(link("JVM", "/admin/jvm?pretty"));
         builder.add(link("Plugins", "/admin/plugins"));
         builder.add(link("Providers", "/admin/providers"));
+        builder.add(link("Prometheus", "/prometheus"));
```
Noddy76, line 229: Very minor point. I think this should either be `/metrics` the default path for Prom scrape metrics path or `/admin/prometheus` which would also happen to be the same as Spring Boot's actuator endpoint for prom metrics IIRC.

ThaerAlAjlouni: I totally agree. But prefer to use Prometheus' defaults, so will change it to `/metric`.

PR #676, opened 2020-06-18
Add a micrometer-backed Graphite reporter

ThaerAlAjlouni:


components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteConfig.java
```diff
@@ -30,16 +30,19 @@
...
     private final String prefix;
+    private final boolean enabled;
 
     @JsonCreator
     GraphiteConfig(@JsonProperty("host") String host,
                    @JsonProperty("port") Integer port,
                    @JsonProperty("intervalMillis") Long intervalMillis,
-                   @JsonProperty("prefix") String prefix) {
+                   @JsonProperty("prefix") String prefix,
+                   @JsonProperty("enabled") String enabled) {
```
Noddy76, line 40: This should probably be a `boolean` or a `Boolean` to be null safe. Either way it should be the same type as the getter.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteConfig.java
```diff
@@ -62,16 +65,23 @@ public long intervalMillis() {
...
 
+    @JsonProperty("enabled")
+    public boolean enabled() {
+        return enabled;
+    }
+
     @Override
     public String toString() {
-        return new StringBuilder(96)
+        return new StringBuilder()
```
Noddy76, line 75: NBD: I realise that this is just how you found it but there is no advantage over string concatenation to using `StringBuilder`.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteConfig.java
ThaerAlAjlouni: What led me to use a `String` is to default the value to `true` when the `enabled` property is missing (to maintain backward compatibility) while using a `boolean` will always default to `false` with `@JsonCreator` if the property is missing.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/graphite/GraphiteConfig.java
ThaerAlAjlouni: I agree, will simplify this implementation.

PR #677, opened 2020-06-22
Convert metrics to use Micrometer

chrisgresty:
Once the work on the `micrometer-migration` branch is complete, this PR will cover merging it back into `master`. Until then, comments are welcome.


components/client/src/test/unit/java/com/hotels/styx/client/StyxBackendServiceClientTest.java
```diff
@@ -354,21 +354,14 @@ public void updatesCountersWhenTransactionIsCancelled() {
...
 
-        assertThat(metricRegistry.getNames(), hasItems(
-                "origins.App-X.requests.cancelled",
-                "origins.App-X.Origin-Y.requests.cancelled"));
-
-        // metricRegistry is already scoped at "origins". Therefore the following metric
-        // names don't need to be prefixed with it:
-        assertThat(metricRegistry.counter("App-X.requests.cancelled").getCount(), is(1L));
-        assertThat(metricRegistry.counter("App-X.Origin-Y.requests.cancelled").getCount(), is(1L));
+        assertThat(meterRegistry.get(OriginMetrics.CANCELLATION_COUNTER_NAME).counter().count(), is(1.0));
```
mikkokar, line 364: Static import

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -200,6 +201,8 @@ private HttpHandler adminEndpoints(StyxConfig styxConfig, StartupConfig startupC
         httpRouter.aggregate("/admin/servers", serverHandler);
         httpRouter.aggregate("/admin/servers/", serverHandler);
 
+        httpRouter.aggregate("/metrics", new PrometheusHandler(environment.meterRegistry()));
```
mikkokar, line 204: We still have `/admin/metrics` interface. What does it show now with the Prometheus metrics?

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
```diff
@@ -200,6 +201,8 @@ private HttpHandler adminEndpoints(StyxConfig styxConfig, StartupConfig startupC
...
+
```
mikkokar, line 205: All other admin interface endpoints have an `/admin` prefix, but the new `/metrics` does not.  The new metrics endpoint should be consistent with the rest and therefore should be `/admin/metrics`.

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
Noddy76: `/metrics` was chosen because it's the default path for prometheus metrics. `the old endpoint still works but will only show the older Dropwizard meters.

components/proxy/src/main/java/com/hotels/styx/admin/AdminServerBuilder.java
Noddy76: `/metrics` was chosen because it's the default path for prometheus metrics.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/ThreadsHandler.java
```diff
@@ -50,9 +45,18 @@ public HttpResponse doHandle(HttpRequest request, HttpInterceptor.Context contex
...
+        PrintWriter writer = new PrintWriter(new OutputStreamWriter(contents));
+
+        writer.println();
+        ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
+        for (ThreadInfo threadInfo : threadMXBean.dumpAllThreads(true, true)) {
+            writer.println(threadInfo.toString());
+        }
+
+        writer.println();
+        writer.flush();
```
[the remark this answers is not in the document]
ThaerAlAjlouni: I thought the same first, but this code is replacing a [codahale implementation](https://github.com/HotelsDotCom/styx/pull/677/files#diff-a9330ef151bdee9ede78b52b85748b7dL18) of `ThreadDump`

components/proxy/src/main/java/com/hotels/styx/admin/handlers/UptimeHandler.java
```diff
@@ -31,20 +34,31 @@
...
+
+    private String formatTime(long timeInMilliseconds) {
+        Duration duration = Duration.ofMillis(timeInMilliseconds);
+
+        long days = duration.toDays();
+        long hours = duration.minusDays(days).toHours();
+        long minutes = duration.minusHours(duration.toHours()).toMinutes();
+
+        return format("%dd %dh %dm", days, hours, minutes);
+    }
```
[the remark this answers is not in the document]
ThaerAlAjlouni: Yes, we have a metric for this, but the added value of keeping this handler IMHO is having a user-friendly view of the uptime directly in each instance's dashboard.

components/proxy/src/main/java/com/hotels/styx/admin/handlers/PrometheusHandler.java
```diff
@@ -0,0 +1,51 @@
...
+    public PrometheusHandler(MeterRegistry meterRegistry) {
+        prometheusRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
+        ((CompositeMeterRegistry) meterRegistry).add(prometheusRegistry);
+    }
+
+    @Override
+    public Eventual<HttpResponse> handle(final HttpRequest request, final HttpInterceptor.Context context) {
+        return Eventual.of(HttpResponse
+                .response(HttpResponseStatus.OK)
+                .disableCaching()
```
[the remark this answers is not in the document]
ThaerAlAjlouni: Done.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/jmx/JmxReporterServiceFactory.java
```diff
@@ -20,16 +20,27 @@
...
 
-        return new JmxReporterService(domain, codaHaleMetricRegistry(environment));
+            @Override
+            public CompletableFuture<Void> stop() {
+                return CompletableFuture.completedFuture(null);
+            }
+        };
+        // String domain = serviceConfiguration.get("domain").orElse("com.hotels.styx");
+
+        // return new JmxReporterService(domain, codaHaleMetricRegistry(environment));
```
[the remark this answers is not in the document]
ThaerAlAjlouni: The `JmxReporterService` was refactored to use Micrometer instead.

components/client/src/test/unit/java/com/hotels/styx/client/StyxBackendServiceClientTest.java
ThaerAlAjlouni: Done

components/proxy/src/main/java/com/hotels/styx/routing/handlers/HostProxy.java
```diff
@@ -287,24 +286,24 @@ private static HostAndPort addDefaultPort(HostAndPort hostAndPort, TlsSettings t
...
+                MeterRegistry meterRegistry,
                 HostAndPort hostAndPort,
                 ConnectionPoolSettings poolSettings,
                 TlsSettings tlsSettings,
                 int responseTimeoutMillis,
                 int maxHeaderSize,
-                String metricPrefix,
-                String objectName) {
+                String appId,
+                String originId) {
```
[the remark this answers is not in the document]
Noddy76: Have a look at the proposal in PR#691 to discuss this further. It's hard to remove completely because Styx has the concept of Origin spread throughout the codebase. But that PR is an attempt to begin removing that.

components/client/src/main/java/com/hotels/styx/client/applications/metrics/OriginMetrics.java
```diff
@@ -40,114 +39,97 @@
...
      *
-     * @param applicationMetrics application metrics
-     * @param originId           an origin
+     * @param registry       a meter registry
+     * @param originId       an origin
+     * @param appId          application ID
      */
-    public OriginMetrics(ApplicationMetrics applicationMetrics, String originId) {
-        this.applicationMetrics = requireNonNull(applicationMetrics);
+    public OriginMetrics(MeterRegistry registry, String originId, String appId) {
```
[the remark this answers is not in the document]
Noddy76: Have a look at the proposal in PR#691 to discuss this further. `OriginMetrics` is by definition about "origins". That PR attempts to move to a more generic nomenclature using "host" and "port".

components/client/src/main/java/com/hotels/styx/client/applications/metrics/OriginMetrics.java
```diff
@@ -40,114 +41,100 @@
...
+    public static final String SUCCESS_COUNTER_NAME = "request.success";
+    public static final String FAILURE_COUNTER_NAME = "request.error";
+    public static final String STATUS_COUNTER_NAME = "response.status";
+    public static final String CANCELLATION_COUNTER_NAME = "request.cancellation";
+    public static final String LATENCY_TIMER_NAME = "request.latency";
+    public static final String TTFB_TIMER_NAME = "request.timetofirstbyte";
 
-    private final String requestMetricPrefix;
+    private static final Duration MIN_HISTOGRAM_BUCKET = of(10, MILLIS);
+    private static final Duration MAX_HISTOGRAM_BUCKET = of(5, MINUTES);
```
mikkokar, line 57: The histogram bucket sizing must be configurable. Seriously.

For example MIN_HISTOGRAM_BUCKET way too high. In AWS load test environment 99.9% of the requests would complete within 10ms. The 75th percentile is around 2ms. These figures are with a test redirect rule engine plugin. Removing this plugin would only speed things up. Sadly I don't have any data with bare styx without plugins.

Depending on the computing environment, instance sizing, origins, and plugin configuration, we could even see sub-millisecond latencies.

Also the max value can be influenced by many factors:
* Database queries in the backend service feeding more HTTP content, over the chunked transfer encoding, as the query progresses
* External payment processors, etc.
* Streaming file downloads/uploads.
* End user "last mile" link speeds. For example a slow mobile connection could trigger Styx flow control on the origin side and so inflate latencies.

Noddy76: Choosing the boundaries for the upper and lower buckets is a trade-off between capturing all possible values and controlling the number of time series created.

Should the percentile being calculated fall outside of the range then it will be capped to the range. We should note that in the documentation.

Good point about the lower bound. The value should be lowered and I'll amend it to 1ms. At that point if the percentile you're looking for has a value outside the range then it will be clipped to the 1ms - 5min range. You would still have the mean and max values.

If, once we try out these values, there is a need to change the value per application, origin or host-proxy then we should raise a ticket to extend the configuration. It might also make sense in the near future to allow the user to specify SLOs so that they can measure their compliance. But I think that belongs in a follow up PR.

kvosper:
Closing old draft PR (no action for 8 months)

PR #681, opened 2020-06-23
Convert remaining codahale metrics

chrisgresty:


components/api/src/main/java/com/hotels/styx/api/MetricRegistry.java
```diff
@@ -32,6 +33,8 @@
  */
 public interface MetricRegistry {
 
+    MeterRegistry micrometerRegistry();
```
chrisgresty, line 36: This is just to enable smaller migration steps. I'll remove it before finalizing this PR.

PR #682, opened 2020-06-24
Add the PluginMeterRegistry to inject the plugin name tag by default

ThaerAlAjlouni:


components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @Incubating(since = "1.1.0")
+    @Nullable
+    public Meter remove(Meter.Id mappedId) {
+        return registry.remove(mappedId);
+    }
+
+    @Incubating(since = "1.2.0")
+    public void clear() {
+        registry.clear();
```
chrisgresty, line 149: This will clear out every plugin's meters, and styx core's. Either limit it to this plugin's meters, or remove this method.

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    public void close() {
+        registry.close();
```
chrisgresty, line 153: This will close the underlying registry. I'd suggest removing this method.

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.ToDoubleFunction;
+import java.util.stream.Stream;
+
+public class PluginMeterRegistry {
```
chrisgresty, line 38: I'd suggest providing methods to return both the underlying registry, and the set of plugin tags. That way, the plugin developer has the convenience of this class, but also has all the tools needed to do whatever he/she may need to do.

components/proxy/src/main/java/com/hotels/styx/startup/extensions/PluginLoadingForStartup.java
```diff
@@ -115,10 +116,19 @@ public MetricRegistry metricRegistry() {
                 return environment.metricRegistry().scope(DEFAULT_PLUGINS_METRICS_SCOPE + "." + factory.name());
             }
 
+            /**
+             * @deprecated deprecated in favor of {@link PluginFactory.Environment#pluginMeterRegistry()}
+             */
+            @Deprecated
```
chrisgresty, line 122: I understand why you have done this. But the fact that you need to indicates that the inheritance hierarchy is not right. If a plugin should not have access to `meterRegistry()` (except through the new class, as I've suggested above) then the plugin environment should not inherit this method.

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+    private final MeterRegistry registry;
+    private final Tag pluginNameTag;
```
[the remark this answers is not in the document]
ThaerAlAjlouni: Updated 👍

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+
+    public PluginMeterRegistry(MeterRegistry registry, String pluginName) {
+        this.registry = registry;
+        this.pluginNameTag = Tag.of("plugin", pluginName);
+    }
+
+    public List<Meter> getMeters() {
+        return registry.getMeters();
```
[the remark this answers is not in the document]
ThaerAlAjlouni: Updated 👍

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+        return registry.timer(name, newTags);
+    }
+
+    public Timer timer(String name, String... tags) {
+        String[] newTags = injectPluginNameTag(tags);
+        return registry.timer(name, newTags);
+    }
+
+    public MeterRegistry.More more() {
+        return registry.more();
```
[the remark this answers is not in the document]
ThaerAlAjlouni: Updated 👍

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
ThaerAlAjlouni: Updated 👍

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
ThaerAlAjlouni: Updated 👍

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
ThaerAlAjlouni: Updated 👍

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Collection;
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.ToDoubleFunction;
+
+public class PluginMeterRegistry {
+    public static final String DEFAULT_TAG_KEY = "plugin";
+    private final MeterRegistry meterRegistry;
+    private final Tag defaultTag;
+    private Tags commonTags;
```
chrisgresty, line 39: Is there a need for this additional set of tags?

ThaerAlAjlouni: They were added to give the user the option of adding custom tags that are common to all meters of a plugin.

chrisgresty: Yes, I understand that, but I couldn't think of an example use case for it. I suppose you could use it to attach the plugin version, maybe. Or perhaps some aspect of the environment that is particular to this instance, that is not captured by Styx's own common tags. Although, Styx doesn't have any common tags, I think - we should probably add some.

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
```diff
@@ -0,0 +1,157 @@
...
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.ToDoubleFunction;
+
+public class PluginMeterRegistry {
+    public static final String DEFAULT_TAG_KEY = "plugin";
+    private final MeterRegistry meterRegistry;
+    private final Tag defaultTag;
```
chrisgresty, line 38: I'd suggest calling this (and related constants/methods) `pluginTag`, it's more meaningful.

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
```diff
@@ -0,0 +1,157 @@
...
+    public Search find(String name) {
+        return Search.in(meterRegistry).tags(Tags.of(defaultTag)).name(name);
+    }
+
+    public RequiredSearch get(String name) {
+        return RequiredSearch.in(meterRegistry).tags(Tags.of(defaultTag)).name(name);
+    }
+
+    public Counter counter(String name, Iterable<Tag> tags) {
+        return meterRegistry.counter(name, commonTags.and(defaultTag).and(tags));
```
chrisgresty, line 88: `commonTags.and(defaultTag).and(tags)` - you use this construction a lot, when `commonTags.and(defaultTag)` is always the same value. I'd suggest that you have a field `allTags` (maybe not the best name) that you update when `commonTags` changes.
Alternatively, you could store the pluginTag as a `Tags` object rather than `Tag`. The `and(..)` method has an optimisation where, if the parameter is empty, it just returns itself, so the common case of `commonTags` being empty, and `tags` being empty, results in no actual object creation/array copying if ordered as `pluginTags.and(commonTags).and(tags)`.  This also has the nice benefit, if you carry it through to the getDefaultTag() method (which would become `Tags getPluginTags()`), that styx core could decide to provide additional plugin tags by default, without changing the interface.

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
ThaerAlAjlouni: I originally started with the name `pluginTag`, but then changed it to `defaultTag` to give the notion of how this tag is **mandatory** and **injected** by default without any intervention needed. What do you think of it?

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
ThaerAlAjlouni: Very nice catch, I will update the code to benefit from that optimization.

components/api/src/main/java/com/hotels/styx/api/plugins/spi/PluginMeterRegistry.java
ThaerAlAjlouni: On second thought, I will rename it to `defaultPluginTag`, what do you say?

ThaerAlAjlouni: On second thought, I will rename it to defaultPluginTag, how does that sound?

PR #685, opened 2020-06-29
Update UrlRequestHealthCheck to use micrometer

JulioDeLeon:


components/client/src/main/java/com/hotels/styx/client/healthcheck/UrlRequestHealthCheck.java
```diff
@@ -15,38 +15,37 @@
...
-     * @param metricRegistry metric registry
+     * @param meterRegistry meter registry
      */
-    public UrlRequestHealthCheck(String healthCheckUri, MetricRegistry metricRegistry) {
+    public UrlRequestHealthCheck(String healthCheckUri, MeterRegistry meterRegistry) {
         this.healthCheckUri = uriWithInitialSlash(healthCheckUri);
-        this.meterCache = new SimpleCache<>(origin -> new FailureMeter(origin, metricRegistry));
+        this.meterCache = new SimpleCache<>(origin -> Counter.builder("origins.healthcheck.failures")
+        .tag("host", origin.host())
+        .tag("port", Integer.toString(origin.port()))
```
chrisgresty, line 47: Most other origin related metrics use the origin ID and application ID (both available on the origin) as tags, rather than host and port.

PR #702, opened 2020-08-05
Provide control of Graphite metrics tags support

chrisgresty:
By default, the Graphite reporter assumes that Graphite does not support tags on metrics. The reporter configuration can be changed to `tagsEnabled: true` to send tags to Graphite.

Also removed the `prefix` configuration item from the Graphite config. This prefix feature does not exist in the micrometer graphite integration.

### DIFF
--- a/components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterServiceTest.java
+++ b/components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterServiceTest.java
@@ -17,10 +17,7 @@ package com.hotels.styx.metrics.reporting.graphite;
 
 import com.hotels.styx.common.StyxFutures;
 import com.hotels.styx.support.matchers.LoggingTestSupport;
-import io.micrometer.core.instrument.Counter;
-import io.micrometer.core.instrument.MeterRegistry;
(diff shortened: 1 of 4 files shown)


components/proxy/src/test/java/com/hotels/styx/metrics/reporting/graphite/GraphiteReporterServiceTest.java
```diff
@@ -17,48 +17,98 @@
...
+        GraphiteReporterService service = serviceBuilder().build();
+
+        try {
+            StyxFutures.await(service.start());
+            GraphiteMeterRegistry graphiteRegistry = (GraphiteMeterRegistry) meterRegistry.getRegistries().iterator().next();
+            assertThat(graphiteRegistry.config().namingConvention(), instanceOf(GraphiteHierarchicalNamingConvention.class));
+        } finally {
+            StyxFutures.await(service.stop());
+        }
+
```
Noddy76, line 109: Could we lose the extra whitespace here and the multiple blank lines at the end of the class, please?

chrisgresty:
By default, the Graphite reporter assumes that Graphite does not support tags on metrics. The reporter configuration can be changed to `tagsEnabled: true` to send tags to Graphite.

Also removed the `prefix` configuration item from the Graphite config. This prefix feature does not exist in the Micrometer Graphite integration.

PR #723, opened 2020-12-17
Prevent NPE when throwable is null

Noddy76:


components/proxy/src/main/java/com/hotels/styx/infrastructure/logging/ExceptionConverter.java
```diff
@@ -63,12 +63,8 @@ public String convert(final ILoggingEvent loggingEvent) {
...
 
     private String getExceptionData(final ILoggingEvent loggingEvent) {
-        IThrowableProxy throwableProxy = loggingEvent.getThrowableProxy();
-        StackTraceElementProxy loggableStackTraceElement =
-                findFirstStackTraceElementForClasses(throwableProxy, readStyxClassesFromProperty())
-                        .orElse(getTopStackTraceElement(throwableProxy));
-
-        return Optional.ofNullable(loggableStackTraceElement)
+        return Optional.ofNullable(loggingEvent.getThrowableProxy())
+                .map(throwableProxy -> findFirstStackTraceElementForClasses(throwableProxy, readStyxClassesFromProperty()).orElse(getTopStackTraceElement(throwableProxy)))
```
kvosper, line 67: This line is quite long.
```
findFirstStackTraceElementForClasses(throwableProxy, readStyxClassesFromProperty()).orElse(getTopStackTraceElement(throwable
Proxy))
```
To make it more readable, I'd suggest *either* putting it on a separate line (but using a block lambda { } to avoid confusion with the outside map/orElse chain) *or* extract the statement into a separate method with an informative name.

kvosper (approved):

PR #729, opened 2021-04-08
Update build version to java11

kvosper:
Currently Styx is built with Java 8, the previous long-term-support version from 2014.
This PR will update it to build with Java 11, which is the latest long-term-support version (from 2018).

See also https://en.wikipedia.org/wiki/Java_version_history

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 dist: trusty
 language: java
 jdk:
-  - openjdk8
   - openjdk11
 script: mvn verify -Pquality,linux
 cache:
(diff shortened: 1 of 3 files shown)


Noddy76:
I looks like you will need to update the `.travis.yaml` script to remove the openjdk8 build before the CI check will pass.

kvosper:
Yes, we need to update `.travis.yml`. I am looking to see if the JDK version can be set for a specific branch.

Noddy76 (approved):

OwenLindsell (approved):

PR #742, opened 2021-07-14
Merge styx-te into Styx Core

olindsell:


Noddy76 (approved):

components/api/src/main/java/com/hotels/styx/api/HttpResponse.java
```diff
@@ -192,7 +192,7 @@ public LiveHttpResponse stream() {
      *
      * @return a set of cookies
      */
-    public Set<ResponseCookie> cookies() {
+    public List<ResponseCookie> cookies() {
```
NersesAM, line 195: This is a breaking change in the api. Not sure what the implications would be.

olindsell: Well spotted!
This could break StyxPres or Vrbo plugins if they upgrade to use the latest version of Styx.
I will check all plugins and update them

components/api/src/main/java/com/hotels/styx/api/LiveHttpResponse.java
```diff
@@ -82,12 +82,14 @@
     private final HttpResponseStatus status;
     private final HttpHeaders headers;
     private final ByteStream body;
+    private final LiveHttpRequest request;
```
kvosper, line 85: This field does not seem to be used anywhere (including via it's accessor method). We should leave it out.

components/api/src/main/java/com/hotels/styx/api/extension/ActiveOrigins.java
```diff
@@ -27,4 +29,8 @@
      * @return a list of connection pools for each active origin
      */
     Iterable<RemoteHost> snapshot();
+
+    String getApplicationId();
```
kvosper, line 33: Looks like these two methods are never called. We should leave them out.

kvosper (changes requested):

components/api/src/main/java/com/hotels/styx/api/LiveHttpResponse.java
olindsell: it can be accessed by the plugins

components/api/src/main/java/com/hotels/styx/api/extension/ActiveOrigins.java
olindsell: It looks like they're being used by LoadBalancingGroup.kt

kvosper: That file only has references to those methods because you were forced to implement them when an anonymous subclass of `ActiveOrigins` is created.

olindsell: Yash Vyas: these are being used in the AdaptiveFailoverStrategy implemented in CGP, we might be able to work around it, i’ll have to take another look tomorrow and i can add some details around these then,

components/api/src/main/java/com/hotels/styx/api/LiveHttpResponse.java
olindsell: Yash Vyas: this is actually used by the logging plugin like you said, and i can provide some context into why this is needed.

kvosper (approved):

PR #745, opened 2021-08-05
Metrics centralisation

kvosper:
A PR to clean up the metrics by registering them from a central file.

Changes:

* Metrics declared in a file called `CentralisedMetrics`
* Some of the names of metrics have been changed to make their purpose more obvious.
* Some unnecessary metrics have been removed

Reviewers please check that no required metrics have been accidentally removed.


components/common/src/main/kotlin/com/hotels/styx/metrics/CentralisedMetrics.kt
```diff
@@ -0,0 +1,326 @@
...
+ * which provides the ability to register the gauge, but not to change the name or tags after they have been defined.
+ *
+ * ## Counters
+ * As counters are quite straightforward, we define them using the micrometer Counter class directly.
+ *
+ * ## Timers
+ * For timers we use our own interface "TimerMetric" instead of the Timer interface from micrometer. This is to make the API more
+ * intuitive to work with, and improve encapsulation.
+ */
+class CentralisedMetrics(val registry: MeterRegistry) {
```
tshe-expedia, line 49: Just to check. I don't see `proxy.connections.openssl` in this file. Do we stop tracking TCP connections?

components/common/src/main/kotlin/com/hotels/styx/metrics/CentralisedMetrics.kt
```diff
@@ -0,0 +1,326 @@
...
+    /* ------ GAUGES ------ */
+
+    /**
+     * Measures the number of requests that have been received, but not yet fully responded to.
+     */
+    @get:JvmName("requestsInProgress")
+    val requestsInProgress: GaugeId = InnerGaugeId("proxy.request.outstanding", Tags.empty())
```
tshe-expedia, line 56: Can we not use `proxy` prefix? It confused me at the beginning when I was migrating HCOM dashboards.

tshe-expedia (approved):

PR #747, opened 2021-09-08
Scope metrics

kvosper:
Create ScopedMeterRegistry for passing to plugins, so that metrics declared within plugins will not clash with metrics from other plugins or core Styx.


components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
```diff
@@ -0,0 +1,216 @@
...
+import io.micrometer.core.instrument.*
+import io.micrometer.core.instrument.MeterRegistry.Config
+import io.micrometer.core.instrument.search.RequiredSearch
+import io.micrometer.core.instrument.search.Search
+import java.util.concurrent.TimeUnit
+import java.util.function.Consumer
+import java.util.function.ToDoubleFunction
+import java.util.function.ToLongFunction
+
+interface MeterRegistry {
```
olindsell, line 27: Possibly this interface could be named to include information about what extra functionality it adds to the Micrometer MeterRegistry?
So maybe XxxMeterRegistry.  Just an idea :)

components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    // Functions below here are not for general metric handling, but needed by Styx internally.
+
+    fun scope(scope: String): ScopedMeterRegistry = ScopedMeterRegistry(this, scope)
+    fun startTimer(): Timer.Sample
+    fun micrometerRegistry(): io.micrometer.core.instrument.MeterRegistry?
+}
+
+class ScopedMeterRegistry(private val registry: MeterRegistry, private val scope: String) : MeterRegistry {
```
olindsell, line 77: Could this take a Micrometer MeterRegistry?

components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
```diff
@@ -0,0 +1,216 @@
...
+    // Functions below here are not for general metric handling, but needed by Styx internally.
+
+    override fun startTimer(): Timer.Sample = registry.startTimer()
+    override fun micrometerRegistry(): io.micrometer.core.instrument.MeterRegistry? = registry.micrometerRegistry()
+
+    private val String.scoped: String get() = "$scope.$this"
+    private val Meter.Id.scoped: Meter.Id get() = withName(name.scoped)
+}
+
+class MicrometerRegistry(private val registry: io.micrometer.core.instrument.MeterRegistry) : MeterRegistry {
```
olindsell, line 150: If above class took a Micrometer MeterRegistry, this class would be obsolete

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
```diff
@@ -117,7 +117,7 @@ private static StyxServer createStyxServer(String[] args) {
         LOG.info("Styx logConfigLocation={}", startupConfig.logConfigLocation());
 
         StyxServerComponents components = new StyxServerComponents.Builder()
-                .registry(Metrics.globalRegistry)
+                .registry(new MicrometerRegistry(new CompositeMeterRegistry()))
```
olindsell, line 120: could we just have a new PrometheusMeterRegistry here?

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpPipelineHandler.java
```diff
@@ -667,7 +668,7 @@ private void cancelSubscription() {
         private HttpErrorStatusListener httpErrorStatusListener = IGNORE_ERROR_STATUS;
         private RequestProgressListener progressListener = IGNORE_REQUEST_PROGRESS;
         private HttpResponseWriterFactory responseWriterFactory = HttpResponseWriter::new;
-        private Supplier<CentralisedMetrics> metricsSupplier = () -> new CentralisedMetrics(new CompositeMeterRegistry());
+        private Supplier<CentralisedMetrics> metricsSupplier = () -> new CentralisedMetrics(new MicrometerRegistry(new CompositeMeterRegistry()));
```
olindsell, line 671: I know you didn't add this functionality,  but I'm not sure why we need a metrics supplier and can't just set metrics directly?

components/server/src/main/java/com/hotels/styx/server/netty/handlers/ChannelStatisticsHandler.java
```diff
@@ -69,12 +68,10 @@ public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
...
-        Thread thread = currentThread();
-        Counter channelCount = metrics.proxy().server().registeredChannelCount(thread);
-        channelCount.increment(amount);
+        ThreadSpecificMetrics metrics = metricsLocal.get();
 
-        DistributionSummary channels = metrics.proxy().server().channelCount(thread);
-        channels.record(channelCount.count());
+        metrics.channelCount.increment(amount);
+        metrics.channels.record(metrics.channelCount.count());
     }
```
olindsell, line 75: Nice! 👍

olindsell (approved):

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/jmx/JmxReporterService.java
```diff
@@ -46,15 +44,15 @@ public JmxReporterService(String domain, MeterRegistry registry) {
     public CompletableFuture<Void> start() {
         return CompletableFuture.runAsync(() -> {
             jmxMeterRegistry = new JmxMeterRegistry(key -> null, Clock.SYSTEM, nameMapper);
-            ((CompositeMeterRegistry) meterRegistry).add(jmxMeterRegistry);
+            meterRegistry.add(jmxMeterRegistry);
```
olindsell, line 47: Any idea why JMX needs its own registry? These metrics won't get reported to DD

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
olindsell: i.e. Do we need a CompositeMeterRegistry if there is only in Registry in it?

components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
kvosper: It doesn't really add new functionality, besides the ability to create a `ScopedMeterRegistry` from it. The main improvement is that it is an interface rather than a class with internal behaviour.
From Styx's PoV, it is a general purpose meter registry, so adding things to the name would be misleading.

components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
kvosper: Possible, but unnecessarily restrictive. Although we could add a secondary "convenience" constructor that does the wrapping to save on typing elsewhere.

components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
kvosper: This is incorrect. If we did that we would only be able to have a single level of scoping. The current implementation allows `ScopedMeterRegistry` to wrap other instances of itself, allowing for multi-level scoping.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
kvosper: I am not certain about this part. I can look into how the `PrometheusMeterRegistry` is added. It probably needs config injection to construct.

components/server/src/main/java/com/hotels/styx/server/netty/connectors/HttpPipelineHandler.java
kvosper: I can't see any reason. I will see if I can simplify this.

components/proxy/src/main/java/com/hotels/styx/metrics/reporting/jmx/JmxReporterService.java
kvosper: I don't know. This has been in Styx for a very long time. It may have been added before we had a clear idea of what to implement.

components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
olindsell: 👍

components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
olindsell: ah ok, yes.  You explained this to me already.  I forgot 🤦

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
olindsell: I just checked - it's created like this: `new PrometheusMeterRegistry(PrometheusConfig.DEFAULT)`

components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
kvosper: Currently, none of the code calls the `ScopedMeterRegistry` constructor besides `MeterRegistry.scope(scope : String)`, so I don't think a secondary constructor is needed.

components/proxy/src/main/java/com/hotels/styx/StyxServer.java
kvosper: It seems to be added to the parent registry in the constructor of one of the admin interface handlers. This is definitely not the right place. I will move it.

kvosper: There is currently a `JmxMeterRegistry` added as well. To avoid changing too much at once, I will keep the `CompositeMeterRegistry`.

components/api/src/main/kotlin/com/hotels/styx/api/MeterRegistry.kt
```diff
@@ -0,0 +1,267 @@
...
+    // Functions below here are not for general metric handling, but needed by Styx internally.
+
+    /**
+     * Creates a new scoped registry based on this registry. If this registry is already scoped, the new scope will be prefixed to the existing scope.
+     *
+     * So, if `fooRegistry` has scope "foo", and we call `val barRegistry = fooRegistry.scope("bar")`, then `barRegistry` will have the scope "bar.foo".
+     */
+    fun scope(scope: String): ScopedMeterRegistry = ScopedMeterRegistry(this, scope)
+    fun startTimer(): Timer.Sample
+    fun micrometerRegistry(): io.micrometer.core.instrument.MeterRegistry?
```
olindsell, line 86: I think maybe we should remove the '?' from the line above as the micrometer registry should never be null

kvosper: OK, I've taken a quick look at the rest of the code, which all assumes that the return value is non-null anyway, so I will remove the `?`.

olindsell (approved):

PR #754, opened 2021-10-11
Add unencoded method

olindsell:


components/api/src/main/java/com/hotels/styx/api/Url.java
```diff
@@ -220,26 +220,24 @@ public URI toURI() {
...
     public Iterable<String> queryParamNames() {
         return query.map(UrlQuery::parameterNames).orElse(emptySet());
     }
 
-    /**
-     * Return the encoded url string.
-     *
-     * @return the encoded url string
-     */
-    public String encodedUri() {
```
helospark, line 237: The Url class is exposed to the plugins via LiveHttpRequest.url(), therefore I think we should make sure to not do any backward-incompatible changes on this API.

By removing this method from the public API, this code change will not be backward compatible, any Styx plugin that used this method will break with `java.lang.NoSuchMethodError: 'java.lang.String com.hotels.styx.api.Url.encodedUri()` Exception during runtime.

Also if a plugin wants to be compatible with a Styx version before and after this change, it would only be possible by using reflection to inspect what methods are available in the API of this class, which is not an ideal design.

I think this method should be put back to the code, delegating to the new method and marked with a `@Deprecated` annotation. Then it could be removed in a major breaking Styx release hopefully in the future when most users have already upgraded.

What do you think?

PR #760, opened 2021-11-17
Remove leftover guava collections

kvosper:
This is a continuation of the Guava removal https://github.com/ExpediaGroup/styx/pull/755

> To reduce the need to continually update third-party dependencies for security vulnerabilities, we can remove dependencies we don't need. Reduced dependencies are also good for Styx, as it is a plugin framework and dependencies can clash with plugins.
> 
> Removing Guava entirely would be a very big PR, so this PR merely reduces it, targeting the use of Guava collections in particular. Future PRs will remove more.
> 
> Many of the Guava methods used are now replicated (or sufficiently approximated) by the Java standard library. Others are easy to implement ourselves, and a few were never needed to begin with.
> 
> These method calls have been replaced or removed.


yashvyas (dismissed):

components/server/src/main/kotlin/com/hotels/styx/server/netty/connectors/ExceptionStatusMapper.kt
```diff
@@ -0,0 +1,70 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package com.hotels.styx.server.netty.connectors
+
+import com.hotels.styx.api.HttpResponseStatus
+import org.slf4j.LoggerFactory.getLogger
+import java.util.*
```
JulioDeLeon, line 20: Can  we break this import down?

components/common/src/main/kotlin/com/hotels/styx/javaconvenience/Util.kt
```diff
@@ -15,14 +15,17 @@
  */
 package com.hotels.styx.javaconvenience
 
-import java.nio.file.Files
-import java.nio.file.Path
-import java.util.concurrent.Callable
-import java.util.function.Supplier
 import java.io.InputStream
 import java.io.OutputStream
+import java.util.*
```
JulioDeLeon, line 20: Can we break this down as well.

JulioDeLeon (approved):

PR #766, opened 2021-11-30
Remove Guava Caches and fix small mistake

kvosper:
This is a continuation of the Guava removal https://github.com/ExpediaGroup/styx/pull/755

Use of Guava's `LoadingCache` can easily be replaced with our own `SimpleCache` since no eviction functionality is required.

This PR also fixes a mistake wherein a reference to Guava `Predicate` was accidentally left in (it extends the Java SL `Predicate` so it still compiled).


ryanellison (approved):

ryanellison (approved):
👍

PR #770, opened 2022-01-10
Allow backend service configuration to specify when incoming host hea…

yashvyas:
…der should be replaced with the value of origin's remote hostname

### DIFF
--- a/components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.java
+++ b/components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.java
@@ -15,13 +15,7 @@
  */
 package com.hotels.styx.client;
 
-import com.hotels.styx.api.HttpInterceptor;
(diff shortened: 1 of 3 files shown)


components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.java
```diff
@@ -140,22 +136,33 @@ private static boolean isHeadRequest(LiveHttpRequest request) {
         Optional<RemoteHost> remoteHost = selectOrigin(request);
         if (remoteHost.isPresent()) {
             RemoteHost host = remoteHost.get();
+
+            LiveHttpRequest withHostHeaderUpdated = request;
```
kvosper, line 140: This logic could be extracted as a separate method.
If we do convert this class to Kotlin, the method would be most intuitive as a (private) extension method.

PR #773, opened 2022-01-12
#771: allow host header override

### LINKED ISSUE #771: BackendService override host header option
In some use cases, it would be desirable to be able to route a request with its Host header set to the value of the hostname property of the selected origin. This could be configured by an additional property on the BackendService (boolean overrideHostHeader). Current behavior maintains the host header of the incoming request, and this should still be the default behavior after this option is added.

ahsansyed97:
fixes #771


components/api/src/main/java/com/hotels/styx/api/extension/service/BackendService.kt
```diff
@@ -0,0 +1,411 @@
...
+package com.hotels.styx.api.extension.service
+
+import com.hotels.styx.api.Id
+import com.hotels.styx.api.Identifiable
+import com.hotels.styx.api.extension.Origin
+import com.hotels.styx.api.extension.Origin.checkThatOriginsAreDistinct
+import java.lang.IllegalArgumentException
+import java.lang.StringBuilder
+import java.net.URI
+import java.util.*
```
ahsansyed97, line 22: No importing *!

components/api/src/main/java/com/hotels/styx/api/extension/service/BackendService.kt
```diff
@@ -0,0 +1,280 @@
...
+    private val path: String,
+    private val connectionPoolSettings: ConnectionPoolSettings,
+    private val origins: Set<Origin>,
+    private val healthCheckConfig: HealthCheckConfig,
+    private val stickySessionConfig: StickySessionConfig,
+    private val rewrites: List<RewriteConfig>,
+    private val overrideHostHeader: Boolean,
+    private val responseTimeoutMillis: Int,
+    private val maxHeaderSize: Int,
+    private val tlsSettings: TlsSettings
```
ryanellison, line 39: I believe we need to allow SSL settings to be null for non-secure services.

components/api/src/main/java/com/hotels/styx/api/extension/service/BackendService.kt
```diff
@@ -0,0 +1,280 @@
...
+        fun maxHeaderSize(maxHeaderSize: Int) = apply {
+            this.maxHeaderSize = maxHeaderSize
+        }
+
+        /**
+         * Sets the https settings.
+         * For Jackson JSON serialiser that de-serialises from Option&lt;TlsSettings&gt;.
+         */
+        fun https(tlsSettings: Optional<TlsSettings?>) = apply {
+            this.tlsSettings = tlsSettings.orElse(null)
```
ryanellison, line 152: Yes, related to my other comment, this can be null.

components/api/src/main/java/com/hotels/styx/api/extension/service/BackendService.kt
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * A builder for [BackendService].
+     */
+    class Builder {
+        var id: Id? = Id.GENERIC_APP
+            private set
+        var path: String? = "/"
+            private set
+        var origins: Set<Origin>? = emptySet()
```
ryanellison, line 77: Can this ever be null? If not, adjust type (similar comment for others in the builder, here)

components/api/src/main/java/com/hotels/styx/api/extension/service/BackendService.kt
```diff
@@ -0,0 +1,280 @@
...
+            private set
+        var connectionPoolSettings: ConnectionPoolSettings? = ConnectionPoolSettings.defaultConnectionPoolSettings()
+            private set
+        var stickySessionConfig: StickySessionConfig? = StickySessionConfig.stickySessionDisabled()
+            private set
+        var healthCheckConfig: HealthCheckConfig? = null
```
ryanellison, line 83: If HealthCheckConfig isn't nullable on BackendService, this is sort of a problem if they don't set this, right? The solution might be to make this nullable on BackendService.

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
```diff
@@ -0,0 +1,358 @@
...
+
+
+    private fun isBodilessResponse(response: LiveHttpResponse): Boolean {
+        val status = response.status().code()
+        return status == 204 || status == 304 || status / 100 == 1
+    }
+
+    private fun isHeadRequest(request: LiveHttpRequest): Boolean = request.method() == HttpMethod.HEAD
+
+    private fun shouldOverrideHostHeader(host: RemoteHost, request: LiveHttpRequest): LiveHttpRequest {
```
ryanellison, line 169: probably use
```
LiveHttpRequest =
...
```
instead of 
```
LiveHttpRequest {
...
```
where possible

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
```diff
@@ -0,0 +1,358 @@
...
+        cause: Throwable,
+        context: HttpInterceptor.Context
+    ): Flux<LiveHttpResponse> {
+        val lbContext: LoadBalancer.Preferences = object : LoadBalancer.Preferences {
+            override fun preferredOrigins(): Optional<String> {
+                return Optional.empty()
+            }
+
+            override fun avoidOrigins(): List<Origin> {
+                return previousOrigins.stream()
```
ryanellison, line 231: un-Java-tize this.

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
```diff
@@ -0,0 +1,358 @@
...
+                return if (nonNull(originsRestrictionCookieName)) {
+                    rewrittenRequest.cookie(originsRestrictionCookieName)
+                        .map { obj: RequestCookie -> obj.value() }
+                        .or { rewrittenRequest.cookie("styx_origin_$id").map { obj: RequestCookie -> obj.value() } }
+                } else {
+                    rewrittenRequest.cookie("styx_origin_$id").map { obj: RequestCookie -> obj.value() }
+                }
+            }
+
+            override fun avoidOrigins(): List<Origin> {
```
ryanellison, line 322:
```
override fun avoidOrigins(): List<Origin> = emptyList()
```

components/api/src/main/java/com/hotels/styx/api/extension/service/BackendService.kt
```diff
@@ -0,0 +1,194 @@
...
+import com.hotels.styx.api.extension.Origin.checkThatOriginsAreDistinct
+import java.lang.IllegalArgumentException
+import java.lang.StringBuilder
+import java.util.Objects
+import java.util.Optional
+
+/**
+ * Represents the configuration of an application (i.e. a backend service) that Styx can proxy to.
+ */
+class BackendService(
```
yashvyas, line 27: consider making this a data class. seems to me like it just holds data.

components/api/src/main/java/com/hotels/styx/api/extension/service/BackendService.kt
```diff
@@ -0,0 +1,194 @@
...
+
+    override fun equals(obj: Any?): Boolean {
+        if (this === obj) {
+            return true
+        }
+        if (obj == null || javaClass != obj.javaClass) {
+            return false
+        }
+        val other = obj as BackendService
+        return (id == other.id
```
yashvyas, line 160: i think this might be the same as the generated equals method on the data class. Pretty sure it compares each property, but maybe reviewing the generated implementation will tell us whether we need to override this (and others like hashCode) at all.

yashvyas: For toString it seems like the choice of curl braces vs parenthesis is the difference.

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
```diff
@@ -0,0 +1,308 @@
...
+        var stickySessionConfig: StickySessionConfig = StickySessionConfig.stickySessionDisabled()
+        var originIdHeader: CharSequence = StyxHeaderConfig.ORIGIN_ID_DEFAULT
+        var retryPolicy: RetryPolicy = RetryNTimes(3)
+        var overrideHostHeader: Boolean = false
+
+        fun build() = StyxBackendServiceClient(this)
+    }
+
+
+    private val rewriteRuleset: RewriteRuleset = RewriteRuleset(rewriteRules)
```
yashvyas, line 90: This should probably move to the top.

yashvyas: do review: https://kotlinlang.org/docs/coding-conventions.html#class-layout, seems like other changes are required. Just makes it easier to read code if everything is organized in the same way. For example, i was wondering why the class doesn't have the 
```
private val rewriteRuleset ...
```
until i reached this line.

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
```diff
@@ -0,0 +1,307 @@
...
+        context: HttpInterceptor.Context
+    ): Publisher<LiveHttpResponse> {
+        if (attempt >= MAX_RETRY_ATTEMPTS) {
+            return Flux.error(NoAvailableHostsException(id))
+        }
+        val remoteHost = selectOrigin(request)
+        return if (remoteHost.isPresent) {
+            val host = remoteHost.get()
+            val updatedRequest = shouldOverrideHostHeader(host, request)
+            val newPreviousOrigins: MutableList<RemoteHost> = ArrayList(previousOrigins)
```
yashvyas, line 135: too verbose, newPreviousOrigins = mutableListOf() works yeah?

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
```diff
@@ -0,0 +1,307 @@
...
+            newPreviousOrigins.add(host)
+            ResponseEventListener.from(
+                host.hostClient().handle(updatedRequest, context)
+                    .map { response: LiveHttpResponse -> addStickySessionIdentifier(response, host.origin()) }
```
yashvyas, line 139: too verbose, don't need to specify types here. Using named argument or the implicit "it" seems like a pretty subjective thing so i would follow project conventions there.

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
yashvyas: maybe this doesn't need to be mutable either.. do verify and consider.

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
```diff
@@ -0,0 +1,307 @@
...
+        fun hosts(): String = hosts(previouslyUsedOrigins)
+
+        companion object {
+            private fun hosts(origins: Iterable<RemoteHost>): String {
+                return origins.asSequence().map { it.origin().hostAndPortString() }.joinToString(", ")
+            }
+        }
+    }
+
+    private fun logError(request: LiveHttpRequest, throwable: Throwable) {
```
yashvyas, line 227: some of these are still just single expressions that don't require a method body. do review.

components/proxy/src/main/java/com/hotels/styx/proxy/StyxBackendServiceClientFactory.kt
```diff
@@ -0,0 +1,117 @@
...
+        val stickySessionEnabled = backendService.stickySessionConfig().stickySessionEnabled()
+        val retryPolicy = ServiceProvision.loadRetryPolicy(
+            styxConfig,
+            environment,
+            "retrypolicy.policy.factory",
+            RetryPolicy::class.java
+        )
+            .orElseGet {
+                defaultRetryPolicy(
+                    environment
```
yashvyas, line 54: seems odd with having just 1 argument, maybe it can be one line?

yashvyas:
1. Use the public github issue in the PR title and commit message, this will link the change to the issue appropriately. Internal ticket is not going to be reference-able. 
2. Doc update should be included in this commit.

components/proxy/src/main/kotlin/com/hotels/styx/services/OriginsConfigConverter.kt
```diff
@@ -98,17 +98,22 @@ internal class OriginsConfigConverter(
...
-        assert(healthCheckConfig.uri().isPresent)
-
-        val str = MAPPER.writeValueAsString(HealthCheckConfiguration(
-                appId,
-                healthCheckConfig.uri().get(),
-                healthCheckConfig.timeoutMillis(),
-                healthCheckConfig.intervalMillis(),
-                healthCheckConfig.healthyThreshold(),
-                healthCheckConfig.unhealthyThreshold()))
+    internal fun healthCheckService(serviceName: String, appId: String, healthCheckConfig: HealthCheckConfig?): ProviderObjectRecord {
```
yashvyas, line 101: is there a good reason to change this signature? Why is it okay to take in a nullable healthCheckConfig here now?

yashvyas: better to use this method instead of the currently defined one? 
```
    internal fun healthCheckServices(apps: List<BackendService>) = apps
            .filter(::isHealthCheckConfigured)
            .mapNotNull {
                it.healthCheckConfig()?.let { hc ->
                    it.id() to hc
                }
            }
            .map { (id, healthCheckConfig) ->
                val appId = id.toString()
                val serviceName = "$appId-monitor"
                
                Pair(serviceName, healthCheckService(serviceName, appId, healthCheckConfig))
            }
```

kvosper (changes requested):

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
```diff
@@ -0,0 +1,332 @@
...
+    private val retryPolicy: RetryPolicy,
+    private val metrics: CentralisedMetrics,
+    private val overrideHostHeader: Boolean
+) : BackendServiceClient {
+    private val rewriteRuleset: RewriteRuleset = RewriteRuleset(rewriteRules)
+
+    private constructor(builder: Builder): this(
+        id = builder.id,
+        originStatsFactory = builder.originStatsFactory!!,
+        loadBalancer = builder.loadBalancer!!,
```
ryanellison, line 64: Do you think we should add a check in .build() for loadBalancer not being null, as you did for metrics? Not a huge fan of using !! in production code, in any case...
You could potentially do like
```
originStatsFactory = builder?.originStatsFactory ?: somethingElseHere
```
even though you know that should never happen.

ahsansyed97: @ryanellison yeah I could definitely do that; @kvosper I'll tag you to get your thoughts too since you suggested the !!

I already check metrics and originStatsFactory being null in .build(), do you think I should also add a check for loadBalancer there **and** do something like your suggestion for **all three** of metrics, originStatsFactory, and loadBalancer here in the constructor to remove the need of !!?

ryanellison: I guess if @kvosper is alright with !! I'd just add an additional check into .build() to protect against this being null, since we're saying it's definitely not with !!.
The tricky think about providing a default for originStatsFactory is the the factory for one of those requires non-null "metrics" which starts to get a little ugly putting it all here...
For reference, in your builder:
```
OriginStatsFactory.CachingOriginStatsFactory(metrics)
```

ryanellison:
Looks good to me, other than the comment added around the builder. @kvosper do we have the ability to release a snapshot of this for use in a CGP canary before merging?

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
ahsansyed97: Cool, I'll add a null check in .build() for loadBalancer and leave the !! as is for now then

ahsansyed97: ^ pushed changes, squashed into the previous commit

ryanellison (approved):
Approving, but let's wait for an approval from Kyle before merging.

components/client/src/main/java/com/hotels/styx/client/StyxBackendServiceClient.kt
kvosper: After thinking about it, `!!` is not the best choice, as this simply throws a `NullPointerException` if something is null. It would be better to use Kotlin's `requireNotNull` like so `requireNotNull(builder.loadBalancer) { "loadBalancer is required" } `.

I may have confused this with java's `Objects.requireNonNull` earlier (which should not be used in Kotlin), but using Kotlin's functions is OK.

kvosper (approved):

yashvyas (approved):

PR #782, opened 2022-03-14
ISSUE-781 - Update Jackson to 2.13.x, Kotlin to 1.6.x

ryanellison:
Resolves: https://github.com/ExpediaGroup/styx/issues/781

Note: Upgrading to Jackson 2.13.x changes the behavior of how config properties left undefined are loaded (null instead of NullNode object). A few changes were added to handle null correctly.

Also...
Refactored JsonNodeConfig and SpiExtension Java --> Kotlin.
Updated patch version of Scala-compiler as required by Jackson upgrade.


battaglia-v (dismissed):
Looks good to me!

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/yaml/JsonNodeConfig.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Construct an instance from a JSON node.
+     *
+     * @param rootNode a JSON node
+     * @param mapper   mapper to convert JSON into objects
+     */
+    init {
+        this.rootNode = rootNode ?: NullNode.getInstance()
+        this.mapper = ObjectMappers.addStyxMixins(Objects.requireNonNull(mapper))
```
kvosper, line 57: There is a Kotlin function that can be used here

```suggestion
        this.mapper = ObjectMappers.addStyxMixins(requireNotNull(mapper))
```

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/yaml/JsonNodeConfig.kt
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    override fun get(key: String): Optional<String> = get(key, String::class.java)
+
+    override fun <T> get(property: String, tClass: Class<T>): Optional<T> =
+        nodeAt(property)
+            .map { node: JsonNode ->
```
kvosper, line 64: This can be simplified to something much easier to understand:
```
.map { node ->
  if (tClass == Path::class.java) {
    Paths.get(node.textValue()) as T
  } else {
    parseNodeToClass(node, tClass)
  }
}
```

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/yaml/JsonNodeConfig.kt
```diff
@@ -0,0 +1,93 @@
...
+                parseNodeToClass(node, tClass)
+            }
+
+    private fun nodeAt(property: String): Optional<JsonNode> = NodePath(property).findMatchingDescendant(rootNode)
+
+    private fun <T> parseNodeToClass(node: JsonNode, tClass: Class<T>): T {
+        val parser = node.traverse()
+        return try {
+            mapper.readValue(parser, tClass)
+        } catch (e: IOException) {
```
kvosper, line 77: There are no checked exceptions in Kotlin, so it is not necessary to wrap in `RuntimeException`.

components/proxy/src/main/java/com/hotels/styx/spi/config/SpiExtension.kt
```diff
@@ -0,0 +1,86 @@
...
+    fun <T> config(configClass: Class<T>?): T {
+        val parser = config.traverse()
+        return try {
+            MAPPER.readValue(parser, configClass)
+        } catch (e: IOException) {
+            throw RuntimeException(e)
+        }
+    }
+
+    override fun hashCode(): Int = Objects.hashCode(factory)
```
kvosper, line 61: Using a Kotlin data class would eliminate the need to manually add `hashCode`, `equals` and `toString` functions.

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/yaml/JsonNodeConfig.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    @Throws(ConversionException::class)
+    override fun <X> `as`(type: Class<X>): X = parseNodeToClass(rootNode, type)
+
+    override fun toString(): String = rootNode.toString()
+
+    companion object {
+        private val YAML_MAPPER = ObjectMappers.addStyxMixins(
+            ObjectMapper(YAMLFactory()).registerKotlinModule())
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
kvosper, line 90: Please import constants (where there is no conflict)

```suggestion
            .configure(FAIL_ON_UNKNOWN_PROPERTIES, false)
```

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/yaml/JsonNodeConfig.kt
```diff
@@ -0,0 +1,93 @@
...
+            .configure(JsonParser.Feature.AUTO_CLOSE_SOURCE, true)
```
kvosper, line 91: Please import constants (where there is no conflict)

```suggestion
            .configure(AUTO_CLOSE_SOURCE, true)
```

kvosper (changes requested):
When performing automatic conversion of Java code to Kotlin, please manually clean up the resulting Kotlin code, as it is often verbose and uses direct-translation of Java paradigms instead of replacing with equivalent Kotlin paradigms. 

Using Kotlin paradigms/style simplifies the code and makes it easier to work with.

components/proxy/src/main/java/com/hotels/styx/spi/config/SpiExtension.kt
ryanellison: If I do change this to a data class, it complains:

"Data class primary constructor must have only property (val / var) parameters"

But, I believe I can't just make these vals in the constructor because the init block needs to set factory/config/enabled to modified values from what is passed into the constructor. Current init with instance vars defined outside of constructor:
```
private val factory: SpiExtensionFactory
private val config: JsonNode
private val enabled: Boolean

init {
        this.factory = requireNonNull(factory, "Factory attribute missing")
        this.config = config ?: getInstance()
        this.enabled = enabled == null
    }
```
Let me know if there's a better way to do this. For now, I'll keep hashCode, equals, and toString but check for potential improvements within those.

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/yaml/JsonNodeConfig.kt
ryanellison: fixed

components/proxy/src/main/java/com/hotels/styx/infrastructure/configuration/yaml/JsonNodeConfig.kt
ryanellison: fixed

olindsell (approved):

components/proxy/src/main/java/com/hotels/styx/spi/config/SpiExtension.kt
kvosper: On further examination, it looks like a `data class` doesn't fit well after all, as only one of the properties is used to make equals/hashcode. Let's leave it as a normal class then.

kvosper (changes requested):

kvosper (approved):

Issue #719, opened 2020-10-17
after unzipped, it cannot run with default.yml,even after several error is fixed.


autoappcreator:
11:27:48.649 [main] ERROR com.hotels.styx.StyxServer - Styx server failed to start due to configuration error in file [D:\workspace\styx\distribution\.\conf\default.yml]: Unexpected field type. Field 'routingObjects[h2Handler].config' should be OBJECT(status, content, headers), but it is NULL

guoxiaodonggit:
same error

PR #783, opened 2022-03-17
ISSUE-719 fix: cannot run with default.yml

kenluluuuluuuuu:
Resolves: #719 

- fix: incorrect yml config in default.yml
- fix: incompatible version of netty-tcnative
- chore: adding build files to .gitignore

Updated the version of dependency - netty-tcnative-boringssl-static because of the incompatibility between netty-tcnative-boringssl-static and netty.
Screenshot 2022-03-17 at 11 01 36

Also, the outdated config properties in default.yml has been removed or renamed to match the current schema requirements, e.g. settings in https://github.com/ExpediaGroup/styx/blob/master/components/proxy/src/main/java/com/hotels/styx/ServerConfigSchema.java and https://github.com/ExpediaGroup/styx/blob/master/components/proxy/src/main/kotlin/com/hotels/styx/servers/StyxHttpServer.kt.

Some of the ports in the default configs are also changed to prevent conflicts against some other ports which have been listened to.

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -35,7 +35,7 @@ If `styx-test/` already exists, remove it first to avoid conflicts on dependenci
 
 Let's unzip this file:
 
-    $ unzip distribution/target/styx.zip -d styx-test
+    $ unzip distribution/target/styx-<VERSION>-<OS>-<PLATFORM>.zip -d styx-test
 
 
 ## 1.4 Running Styx
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -8,8 +8,7 @@ In this guide we will walk through how to build applications on top of Styx.
 
 ## 1.2 System Requirements
 
-Building Styx requires Java 1.8. It can be built with 1.8.0_45. Earlier maintenance
-releases may, but are not guaranteed to work.
+Building Styx requires Java 11.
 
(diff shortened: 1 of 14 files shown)


docs/developer-guide.md
```diff
@@ -30,6 +29,10 @@ To build for Linux, you need to override a *PLATFORM* argument, like so:
 
     $ make release-no-tests PLATFORM=linux
 
+If `styx-test/` already exists, remove it first to avoid conflicts on dependencies:
+
+    $ rm -rf styx-test
+
 Let's unzip this file:
 
     $ unzip distribution/target/styx.zip -d styx-test
```
kvosper, line 38: This appears to be inaccurate, the file appears at `styx-<VERSION>-<OS>-<PLATFORM>.zip` instead (as described in `docs/user-guide/basic-usage.md`

kvosper (changes requested):

docs/developer-guide.md
```diff
@@ -8,8 +8,7 @@ In this guide we will walk through how to build applications on top of Styx.
```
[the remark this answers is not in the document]
kenluluuuluuuuu: done

docs/developer-guide.md
kenluluuuluuuuu: done

kvosper (approved):

PR #801, opened 2022-10-24
fix: overrideHostHeader not being consumed by HostProxy if default backend service is used for httpPipelines

kenluluuuluuuuu:
### :pencil: Description
- To make it recognise overrideHostHeader in orgins.yaml if YamlFileConfigurationService is used

### :link: Related Issues
-


zckoh (approved):
LGTM

PR #803, opened 2023-01-16
fix: modify tcp keep alive idle time to avoid connection dropped by requested servers

kenluluuuluuuuu:
### :pencil: Description
The connection was dropped by servers in the upstream. We need to tweak the TCP keepalive idle time on Styx to make it send heartbeat to upstream to reset the TCP connection idle timeout before the upstream drops the connection.

Linux has TCP_KEEPALIVE_IDLE_TIME set to 7200 seconds by default. We need to set this to smaller than the idle timeout of server/upstream to avoid TransportLostException.

---
I found it a bit hard to write tests for this scenario as we have to simulate a connection idle timeout on the server side/upstream and our current test utility for mocked http server (WireMockServer) doesn’t allow us to modify the idle timeout.

### DIFF
--- a/components/api/pom.xml
+++ b/components/api/pom.xml
@@ -95,6 +95,11 @@
       <artifactId>reactor-test</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>io.kotlintest</groupId>
+      <artifactId>kotlintest-runner-junit5</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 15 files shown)


kvosper (approved):

components/api/pom.xml
```diff
@@ -95,6 +95,11 @@
       <artifactId>reactor-test</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>io.kotlintest</groupId>
```
NersesAM, line 100: 😬 this is really old, needs to be updated to io.kotest

PR #819, opened 2023-07-24
feat: increment responsesByStatus metric when Styx returns 408 Request Timeout

zckoh:
### :pencil: Description
Currently when Styx itself returns a "408 Request Timeout" response, currently it does not increment the `responsesByStatus` metric, as the `HttpPipelineHandler`'s `handleChannelException` method directly calls the respondAndClose method.

This change updates the `HttpErrorStatusMetrics` which is a `HttpErrorStatusListener` to increment the metric if the response status is a 408 Request Timeout

I've also made a few code cleanup changes

### DIFF
--- a/components/proxy/src/main/java/com/hotels/styx/proxy/HttpErrorStatusMetrics.java
+++ b/components/proxy/src/main/java/com/hotels/styx/proxy/HttpErrorStatusMetrics.java
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2013-2021 Expedia Inc.
+  Copyright (C) 2013-2023 Expedia Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


components/proxy/src/main/java/com/hotels/styx/proxy/HttpErrorStatusMetrics.java
```diff
@@ -79,7 +80,9 @@ private void incrementExceptionCounter(Throwable cause, HttpResponseStatus statu
         if (!(cause instanceof PluginException)) {
             if (INTERNAL_SERVER_ERROR.equals(status)) {
                 metrics.proxy().styxErrors().increment();
-            } else if (status != null && status.code() > 500) {
+            } else if (REQUEST_TIMEOUT.equals(status)) {
+                metrics.proxy().server().responsesByStatus(status.code()).increment();
+            } else if (status.code() > 500) {
```
ryanellison, line 85: Just double checking, status can't be null? I see we're removing a null check.

ryanellison (dismissed):
Approving assuming my comment can be disregarded.

ryanellison (approved):

PR #823, opened 2023-08-09
Refactoring

kenluluuuluuuuu:
### :pencil: Description
- Renaming `BusyConnectionsStrategy` to `BusyActivitiesStrategy` for more general usage, e.g. load balancing based on number of ongoing requests in http2
- `OriginsInventory` is now an abstract class. `StyxOriginsInventory` has become the default implementation of `OriginsInventory`.
- Allow passing in `SslContextBuilder` for getting/creating `SslContext` in `SslContextFactory`
- Fixed E2E tests broken by TLSv1 and TLSv1.1 being switched off by default in OpenJDK


components/client/src/main/kotlin/com/hotels/styx/client/StyxOriginsInventory.kt
```diff
@@ -0,0 +1,134 @@
...
+import com.hotels.styx.common.QueueDrainingEventProcessor
+import com.hotels.styx.common.StyxFutures.await
+import com.hotels.styx.metrics.CentralisedMetrics
+import javax.annotation.concurrent.ThreadSafe
+
+/**
+ * An inventory of the origins configured for a single application.
+ */
+@ThreadSafe
+class StyxOriginsInventory(
```
kvosper, line 35: Why do we need to distinguish a "Styx" OriginsInventory when all the code in the project is Styx? We need to be clear about why we use polymorphism.

kvosper (changes requested):

components/client/src/main/kotlin/com/hotels/styx/client/StyxOriginsInventory.kt
kenluluuuluuuuu: Maybe the name is not good, but the intention is to allow people to extends the implementation of OriginsInventory. For example, if people want to implement their own host http client.

components/client/src/main/java/com/hotels/styx/client/ssl/SslContextFactory.java
```diff
@@ -55,17 +55,28 @@ public static SslContext get(TlsSettings tlsSettings) {
...
+    public static SslContext get(TlsSettings tlsSettings, SslContextBuilder sslContextBuilder) {
+        return SSL_CONTEXT_CACHE.computeIfAbsent(tlsSettings, settings -> create(settings, sslContextBuilder));
+    }
+
     private static SslContext create(TlsSettings tlsSettings) {
         try {
-            return createSslContext(tlsSettings);
+            return createSslContext(tlsSettings, SslContextBuilder.forClient());
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
kenluluuuluuuuu: done

kvosper (approved):

PR #827, opened 2023-09-08
chore: always include amd linux and arm linux dependencies for netty dependencies and eliminate os based profiles

kenluluuuluuuuu:
This is mainly to fix epoll not available issue on ARM machines
macos won't be a problem because it should fallback to nio.


NersesAM (approved):
LGTM

PR #831, opened 2023-10-19
chore: Setup PR check builds

NersesAM:


kenluluuuluuuuu (dismissed):

zckoh:
Do we have runners to run this?

.github/workflows/pr-check.yml
```diff
@@ -0,0 +1,26 @@
+name: PR Check
+on:
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+    inputs:
+      REF:
+        required: false
+        default: master
```
zckoh, line 10: Just curious, does this do anything?

zckoh (approved):

.github/workflows/pr-check.yml
NersesAM: it is for triggering on a branch without PR, isn't it?

NersesAM: it is to be able to run it agains any commit/branch/tag isn't it without a PR

zckoh: Ok I don't see it being used explicitly in the checkout action. Is this being used implicitly?

kenluluuuluuuuu (approved):

PR #834, opened 2023-10-20
chore: Upgrade most of the libraries to latest versions apart from ScalaTest ones.

NersesAM:


ryanellison (approved):

components/proxy/src/main/java/com/hotels/styx/infrastructure/logging/LOGBackConfigurer.java
```diff
@@ -121,26 +120,6 @@ public static void initLogging(URL url, boolean installJULBridge) {
...
-        LoggerContext loggerContext = selector.getLoggerContext();
-        String loggerContextName = loggerContext.getName();
-        LoggerContext context = selector.detachLoggerContext(loggerContextName);
-        if (uninstallJULBridge) {
-            SLF4JBridgeHandler.uninstall();
-        }
-        context.stop();
-    }
-
-
```
NersesAM, line 143: Due to logback internal changes this needed to be rewritten as it wasn't compiling anymore. I've opted in removing this as we are not deploying styx in J2EE and it didn't worth investing effort in it.

PR #851, opened 2024-03-04
Fix: Introducing header list size configuration

JulioDeLeon:
### :pencil: Description
- Introducing a configuration in HTTPConfig to allow users to control header list size on the Netty Client.


JulioDeLeon:
Based upon further discussion, we have determined a change like this is unnecessary.

PR #852, opened 2024-03-13
support brotli compression

sullis:
### :pencil: Description
- add support for brotli compression

This PR adds the [brotli4j](https://github.com/hyperxpro/Brotli4j) library.

Netty 4.x will automatically detect the presence of brotli4j.

When brotli4j is on the runtime classpath, Netty enables brotli compression.

If brotli4j is missing from the runtime classpath, Netty does not provide brotli compression.


sullis:
cc: @hyperxpro

components/proxy/src/test/java/com/hotels/styx/proxy/BrotliTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.jupiter.api.condition.OS;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+
+public class BrotliTest {
+    @Test
+    @EnabledOnOs(value = { OS.LINUX, OS.MAC })
+    public void brotliCompressionIsAvailable() throws Throwable {
+        Brotli.ensureAvailability();
```
hyperxpro, line 30: You can remove this because the below `assert` will fail if the native library is not available.

hyperxpro (changes requested):

components/proxy/src/test/java/com/hotels/styx/proxy/BrotliTest.java
sullis: 👍   I just pushed an update.  I removed BrotliTest.java and moved 1 assertion into HttpCompressorTest

sullis:
@kenluluuuluuuuu  this PR is a proof of concept. Please let me know what you think.

sullis:
@kenluluuuluuuuu is this an active project?

kenluluuuluuuuu:
Hi @sullis,
Thanks for attempting to contribute to our project.
However, we're gradually decommissioning the project, and we will only put minimal effort to maintain it.
We won't be adding new features unless there's an absolute need to do so.

Appreciate for the effort and your understanding.

Thanks,
Ken

PR #858, opened 2024-05-20
fix: incorrect implementation of request and response timing

kenluluuuluuuuu:


components/client/src/main/kotlin/com/hotels/styx/client/connectionpool/LatencyTiming.kt
```diff
@@ -46,7 +46,27 @@ object LatencyTiming  {
...
+     * @param metrics metrics - required to access timers to start them
+     * @param context context that can have timers set in it
+     */
+    @JvmStatic
+    fun stopResponseTiming(
+        metrics: CentralisedMetrics?,
+        context: HttpInterceptor.Context?,
+    ) {
         if (metrics != null && context is TimeMeasurable) {
             context.timers?.startTiming(metrics, RESPONSE_PROCESSING)
```
ThaerAlAjlouni, line 71: The `stopResponseTiming` method seems to be calling the `startTiming` one, is that intended?

kenluluuuluuuuu: oh, shoot, let me fix it

kenluluuuluuuuu: done

ThaerAlAjlouni (approved):

JulioDeLeon (approved):

